PR #27, opened 2023-01-06
Run the CLA bot only on comments to PR

mnvr:
Right now, it unnecessarily runs and sends an email when customers open issues.

Not tested, but can test it by opening an issue after this is merged.

Reference:
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment

### DIFF
--- a/.github/workflows/cla.yaml
+++ b/.github/workflows/cla.yaml
@@ -7,6 +7,8 @@ on:
 
 jobs:
     CLAAssistant:
+        # This job only runs for pull request comments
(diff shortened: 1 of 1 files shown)


mnvr:
Okay, so it looks like it is skipping the action (tested with https://github.com/ente-io/auth/issues/28), but now it is sending the email that it is skipping the action 🤦

I am hoping customers opening the issue won't get this skip email, only people watching the repo will get it. So I guess it is an improvement.

PR #387, opened 2023-12-05
Added: When the app is in the background, hided te contents of the app

### LINKED ISSUE #47: iOS - Hide the contents in recent apps when lockscreen is set
Feature request via email

> After I set Lockscreen, and I swipe up the auth app (iPhone), the contents of the app can be seen like many other apps except password manager (set lockscreen). Like password manager, can the contents be hidden when Lockscreen is set?

i-aiymen:
## Description

Added: Hide the contents of the app, when is it in the background - Both android and ios

In android, it shows a blank white/black background, when in background

In iOS, it shows the launch icon, when in background

photo_2023-12-06_12-53-02

Android Physical Device

photo_2023-12-06_12-52-55

Android stimulator

Also, compared the bitwarden apps in android, the flow is same

iOS Device: Need to verify 
@ua741 @vishnukvmd Could you run on iOS devices and verify it?

Closes #47 

## Type of Change

- [ ] 🖼️ New icon
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -41,7 +41,9 @@ void main() async {
     androidOptions: const PrivacyAndroidOptions(
       enableSecure: true,
     ),
-    backgroundColor: Colors.white.withOpacity(0),
+    backgroundColor: savedThemeMode == AdaptiveThemeMode.dark
+        ? Colors.black.withOpacity(0)
+        : Colors.white.withOpacity(0),
     blurEffect: savedThemeMode == AdaptiveThemeMode.dark
         ? PrivacyBlurEffect.dark
         : PrivacyBlurEffect.extraLight,
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -36,12 +36,10 @@ void main() async {
     iosOptions: const PrivacyIosOptions(
       enablePrivacy: true,
       privacyImageName: "LaunchImage",
-      autoLockAfterSeconds: 5,
       lockTrigger: IosLockTrigger.didEnterBackground,
     ),
     androidOptions: const PrivacyAndroidOptions(
       enableSecure: true,
-      autoLockAfterSeconds: 5,
     ),
     backgroundColor: Colors.white.withOpacity(0),
     blurEffect: PrivacyBlurEffect.extraLight,
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -42,8 +42,8 @@ void main() async {
       enableSecure: true,
     ),
     backgroundColor: savedThemeMode == AdaptiveThemeMode.dark
-        ? Colors.black.withOpacity(0)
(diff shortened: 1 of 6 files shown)


lib/main.dart
```diff
@@ -24,13 +24,28 @@ import "package:flutter/material.dart";
...
+      privacyImageName: "LaunchImage",
+      autoLockAfterSeconds: 5,
+      lockTrigger: IosLockTrigger.didEnterBackground,
+    ),
+    androidOptions: const PrivacyAndroidOptions(
+      enableSecure: true,
+      autoLockAfterSeconds: 5,
+    ),
+    backgroundColor: Colors.white.withOpacity(0),
+    blurEffect: PrivacyBlurEffect.extraLight,
```
vishnukvmd: Let's set this depending on the app's theme?

lib/main.dart
```diff
@@ -24,13 +24,28 @@ import "package:flutter/material.dart";
...
 final _logger = Logger("main");
 
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
   final savedThemeMode = await AdaptiveTheme.getThemeMode();
   await _runInForeground(savedThemeMode);
+  await PrivacyScreen.instance.enable(
+    iosOptions: const PrivacyIosOptions(
+      enablePrivacy: true,
+      privacyImageName: "LaunchImage",
```
vishnukvmd, line 38: Does this asset exist?

lib/main.dart
```diff
@@ -24,13 +24,28 @@ import "package:flutter/material.dart";
...
+      autoLockAfterSeconds: 5,
```
vishnukvmd: Do we need `autoLock` since we already have a separate lockscreen?

vishnukvmd: Also, how does this affect the behavior of the app when the existing lock screen is turned on?

lib/main.dart
i-aiymen: it's inside the ios folder

i-aiymen: LaunchImage asset was already created in the ios folder

i-aiymen: if its not working, then we have to create it

lib/main.dart
i-aiymen: will remove the autolock. Its irrelevant

vishnukvmd (approved):

PR #406, opened 2024-01-02
fix: bitwarden imports

ImJustToNy:
## Description

There are two issues with Bitwarden imports ATM:

1. TOTP secrets in Bitwarden are sometimes saved in a format of `otpauth://[...]` and sometimes in a format of secret only. Ente auth assumes that only the second one is used, which is not true and causes an exception. To make both methods work as excepted, in the code, I'm checking which format is used and create `Code` object accordingly.
2. Sometimes items in Bitwarden exports are not accounts, but rather notes. In that case, the `item` variable in the `_processBitwardenExportFile` method does not contain the `login` attribute, which causes an exception. I've added a check to make sure that the item we're parsing is indeed an account.

It is my first time making something in Dart, so please don't be too harsh about my code :)

## Type of Change

- [ ] 🖼️ New icon
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/ui/settings/data/import/bitwarden_import.dart
+++ b/lib/ui/settings/data/import/bitwarden_import.dart
@@ -60,7 +60,7 @@ Future<void> _pickBitwardenJsonFile(BuildContext context) async {
     if (count != null) {
       await importSuccessDialog(context, count);
     }
-  } catch (e, stacktrace) {
+  } catch (e) {
     await progressDialog.hide();
     await showErrorDialog(
       context,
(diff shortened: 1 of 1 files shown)


vishnukvmd (approved):

vishnukvmd:
Thanks for this @ImJustToNy! 🙏

PR #407, opened 2024-01-02
Fix: dark icon hard to see

### LINKED ISSUE #403: Black icons are hard to see in dark mode
**Describe the bug**
If a service uses the icon from Simple Icons and the brand color is black (`#000000`), ente Auth displays the icon in black, even if dark mode is enabled. This makes the icon hard to see (black on dark grey).

**To Reproduce**
1. Switch the app to dark mode
2. Create a TOTP entry for a service with a black or very dark icon (e.g. CodePen)

**Expected behavior**
Black or very dark icons should be flipped to white when the app is in dark mode. This would also avoid having to duplicate icons from Simple Icons to grey custom icons (like with GitHub at the moment). All those black icons would automatically be displayed correctly.

**Possible implementations**
- Easiest, but does not catch all cases: If the brand color is `#000000` and dark mode is active, use `#ffffff` as brand color
- More elaborate: If the brand color is neutral grey (R, G and B values are the same), calculate the lightness of the color. In dark mode, check if the lightness is below e.g. 10%. In light mode, check if the lightness is above e.g. 90%. In these cases, flip the lightness (e.g convert 7% to 93%).

**Screenshots**
Example with CodePen

**Smartphone (please complete the following information):**
iPhone 13 mini, iOS 16.7.2, ente Auth 2.0.25

aadarsh-patel:
## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

## Description

Fall-back to theme's icon-color when the icon-color is too light or dark. Fixes #403

Logic
1. If RGB values are almost equal (`#000000`, `#0F0F11`,  `#212121`, `#27272A`, `#464949`, `#FFFFFF`)
2. Compute its luminance/brightness (cache this value as it is an expensive task)
3. If its too bright or dark, return theme's icon-color

I've manually set the threshold values for brightness in light-theme to be `0.7` and in dark-theme to be `0.05`

https://github.com/ente-io/auth/assets/53324291/aa1e8413-631d-4039-8c08-f8c4d1856fdb

### DIFF
--- a/lib/ui/utils/icon_utils.dart
+++ b/lib/ui/utils/icon_utils.dart
@@ -1,5 +1,6 @@
 import 'dart:convert';
 
+import 'package:ente_auth/ente_theme_data.dart';
 import 'package:ente_auth/theme/ente_theme.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
@@ -14,6 +15,8 @@ class IconUtils {
   // Map of icon-title to the color code in HEX
   final Map<String, String> _simpleIcons = {};
   final Map<String, CustomIconData> _customIcons = {};
+  // Map of icon-color to its luminance
+  final Map<Color, double> _colorLuminance = {};
 
   Future<void> init() async {
     await _loadJson();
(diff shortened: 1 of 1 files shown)


vishnukvmd (approved):

vishnukvmd:
Thanks for this @aadarsh-patel!

aadarsh-patel:
Happy to help😊

PR #516, opened 2024-03-02
Allow developers to customize their VSCode configuration

mnvr:
For folks who're using the git integration from within VSCode, the top level directory needs to be open in VSCode for using git GUI from VSCode.

However, for using the Run & debug configurations they need to open ente/mobile is opened in VSCode.

As a solution (at least to start with):

- gitignore .vscode at the top level
- provide sample VSCode configuration files is mobile/docs/vscode and auth/docs/vscode
- mention it in the README

(will update the auth README later, have to do a few more changes to it first)


mnvr:
Please feel free to edit the sample configuration files (e.g. to rename the launch configs etc as you see fit). I haven't actually run this is VSCode, so I don't know how they look, please adjust as per your ergonomics.

mobile/.gitignore
```diff
@@ -42,4 +38,4 @@ android/app/.settings/*
 
 
 fastlane/report.xml
-TensorFlowLiteC.framework
\ No newline at end of file
+TensorFlowLiteC.framework
```
ashilkn, line 41: @ua741 This is no more necessary right?

PR #671, opened 2024-03-04
Add GitHub Action for releasing the auth app (mobile and desktop)

mnvr:
## Description

This action will take the latest auth code, build the iOS/Android apps, and create a draft release using it.

It will also then create a desktop build of auth. Since the desktop build is currently in beta, this might be neutered when we merge this into main for now.

## Tests

This PR _is_ the test. I'll be adding commits as I fill this in. These commits will create a test tag (that I'll also delete later).


mnvr:
@prateekmedia @ua741 You can now pull this workflow file into your beta branch. It builds both the mobile and desktop apps for auth if a tag like `auth-v1.2.3-xxx` is pushed. The `-xxx` is optional, and can be used for testing the workflow (but remember to delete these test tags later).

Instructions are in docs/release.md (part of the same PR), but please ask if something's not apparent. Note that for now I've disabled the steps that actually build the desktop binaries, so maybe there is some typos in them (I took these bits from the corresponding workflow file in the beta branch, but maybe I accidentally introduced some typo).

An example run: https://github.com/ente-io/ente/actions/runs/8152165219




*

PR #726, opened 2024-03-06
[Auth] Allow for configuring a custom server

vishnukvmd:
## Description
Users can now tap on the onboarding screen **7 times** to bring up a page where they can configure the endpoint the app should be connecting to.

self-host

## Tests
- [x] Verified that production flows are working as expected
- [x] Verified that configuring the endpoint to a local instance lets you
  - [x] Connect to that instance 
  - [x] Create an account
  - [x] Add a key
  - [x] Modify a key
  - [x] Logout and log back in

### DIFF
--- a/auth/lib/core/configuration.dart
+++ b/auth/lib/core/configuration.dart
@@ -6,6 +6,7 @@ import 'dart:typed_data';
 import 'package:bip39/bip39.dart' as bip39;
 import 'package:ente_auth/core/constants.dart';
 import 'package:ente_auth/core/event_bus.dart';
+import 'package:ente_auth/events/endpoint_updated_event.dart';
 import 'package:ente_auth/events/signed_in_event.dart';
 import 'package:ente_auth/events/signed_out_event.dart';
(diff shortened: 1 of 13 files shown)


ua741:
Should we show the server url on UI when it's not the production endpoint (https://api.ente.io)

As this url will be cleared on logout, it will be probably helpful when it's shown during login/sign up and may be the in hamburger menu below version.

ua741 (approved):

vishnukvmd:
> Should we show the server url on UI when it's not the production endpoint (https://api.ente.io)
> 
> As this url will be cleared on logout, it will be probably helpful when it's shown during login/sign up and may be the in hamburger menu below version.

Done. Have also added a check to `/ping` to verify the endpoint.

jo-chemla:
See comment [here](https://github.com/ente-io/ente/pull/1302#issuecomment-2456902981), if you want to switch back to using the official ente api endpoint rather than a custom endpoint, please replace that string with `https://api.ente.io`

PR #733, opened 2024-03-07
Fix warnings around missing awaits

vishnukvmd:


auth/lib/ui/settings/data/import_page.dart
```diff
@@ -105,7 +105,7 @@ class ImportCodePage extends StatelessWidget {
                               index != importOptions.length - 1,
                           isTopBorderRadiusRemoved: index != 0,
                           onTap: () async {
-                            ImportService().initiateImport(context, type);
+                            await ImportService().initiateImport(context, type);
```
vishnukvmd, line 108: @ua741 please double check

auth/lib/ui/settings/data/import/import_service.dart
```diff
@@ -19,29 +19,29 @@ class ImportService {
   Future<void> initiateImport(BuildContext context, ImportType type) async {
```
vishnukvmd, line 19: @ua741 please double check

ua741 (approved):

PR #749, opened 2024-03-08
[Photos] Fix auto scaling on loading final image

ashilkn:
## Description

When an image is zoomed in, there were cases where when then final image is loaded and rendered on screen, the image gets zoomed even more. 

#### Case 1
Double tapping twice to zoom and then the final image is loaded.

##### Before

https://github.com/ente-io/ente/assets/77285023/7d9acb0f-2849-4ffb-863d-cab61ef4bd16

##### After

https://github.com/ente-io/ente/assets/77285023/89fb1238-1444-4681-a94f-7a58679f9350

#### Case 2
When screen is pressed or dragged (any contact with screen).

##### Before

https://github.com/ente-io/ente/assets/77285023/3aefdec5-3bc3-42ef-9442-592442273569

##### After

https://github.com/ente-io/ente/assets/77285023/7b458184-5abe-4983-ba21-f38080cc8bee

## Tests

Tested for regressions. Found one, which it not quite a regression as the issue was already present but reproducible in a different flow. It happens too fast for screen recording to capture it.

Not a blocker as it doesn't affect usability in anyway and since this fix (auto scaling on final image loading) is an important fix.


mobile/lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -54,6 +53,7 @@ class _ZoomableImageState extends State<ZoomableImage>
   ValueChanged<PhotoViewScaleState>? _scaleStateChangedCallback;
   bool _isZooming = false;
   PhotoViewController _photoViewController = PhotoViewController();
+  late PhotoViewScaleStateController _scaleStateController;
```
vishnukvmd: Why not initialize this here?

vishnukvmd (approved):

PR #775, opened 2024-03-09
[workflow] Retain the PlayStore AAB as a workflow artifact

mnvr:
Reasoning in the workflow comment.

While this wouldn't have helped in cases like the recent workflow failure (https://github.com/ente-io/ente/actions/runs/8211812037/job/22461255230), it is maybe a good idea to attach the AAB to the workflow artifacts so that we can get at the binary if needed.

I'm not fully sure if we need this though. So we can close this PR too - I just wanted to figure how to do this if we need to make this amendment later.

## Tests

Untested. I validated the YAML though.

    yq eval .github/workflows/auth-release.yml

### DIFF
--- a/.github/workflows/auth-release.yml
+++ b/.github/workflows/auth-release.yml
@@ -80,6 +80,15 @@ jobs:
                   SIGNING_KEY_PASSWORD: ${{ secrets.SIGNING_KEY_PASSWORD }}
                   SIGNING_STORE_PASSWORD: ${{ secrets.SIGNING_STORE_PASSWORD }}
 
+            # Retain the PlayStore AAB as an workflow artifact so that it can be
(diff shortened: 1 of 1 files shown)


vishnukvmd:
It'd help us to have the `AAB` (so we don't have to bother building it) but that artifact getting published with the release might confuse some users, who might opt for it instead of the `APK`, and that could break the auto-update flow. I'll close this for now. That said, I don't have a solution to the original issue. Ideally it shouldn't happen as long as we resort to the action as the only path to upload to PlayStore.

mnvr:
It won't get published with the release. The release artifacts uploader and the workflow artifact uploader are different. This'll just appear as an artifact if someone goes to the workflow that built the release (most likely us).

But anyways, yes this is not needed for now.

PR #986, opened 2024-03-09
Correct the order for creating s3 presigned objects

blmhemu:
*Note: I am still finding my way through the codebase. If this PR seems illogical LMK.*
## Description
In the current scenario, we first create a s3 object, presign it and then insert into the DB.
If this fails just before inserting into the db, we have a dangling object we don't cleanup.

But if we first insert it in DB, then it should work fine as background cleaner will be able to find it always.

Maybe this is needed for multipart uploads as well, but it may not be trivial to fix that.

### DIFF
--- a/server/pkg/controller/file.go
+++ b/server/pkg/controller/file.go
@@ -803,15 +803,15 @@ func (c *FileController) verifyFileAccess(actorUserID int64, fileID int64) error
 }
 
 func (c *FileController) getObjectURL(s3Client *s3.S3, dc string, bucket *string, objectKey string) (ente.UploadURL, error) {
+	err := c.ObjectCleanupCtrl.AddTempObjectKey(objectKey, dc)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


ua741:
Thank you for the PR and detailed description.

>  We first create a s3 object, presign it and then insert into the DB.

Slight correction. 
`PutObjectRequest` creates a request to upload an object to S3 but does not created it. We are preparing the request for signing.

ua741:
Closing this pull request. Please let us know if you see opposite behaviour.. i.e. objects are being automatically created after calling `PutObjectRequest`.

blmhemu:
Hey ! Thanks for the reply - looks like it just stores the [metadata](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/PutObjectRequest.html) - `Uploads a new object to the specified Amazon S3 bucket. The PutObjectRequest optionally uploads object metadata and applies a canned access control policy to the new object.`

PR #992, opened 2024-03-10
[mobile][photos] Home Widget fixes

prateekmedia:
## Description

- Upscale default placeholder
- Keep image ready to be rendered in home widget before adding the home widget.
- Change the photo in home widget every time it's clicked.
- Open favourites when home widget is clicked.
- Fix multiple issues

## Tests

Did a good amount of testing.


mobile/lib/main.dart
```diff
@@ -275,9 +275,15 @@ Future<void> _scheduleHeartBeat(
 }
 
 Future<void> _scheduleFGHomeWidgetSync() async {
```
ashilkn, line 283: What's the reason for scheduling it after 15 minutes?

mobile/lib/services/favorites_service.dart
```diff
@@ -51,6 +52,7 @@ class FavoritesService {
       }
     });
     await _warmUpCache();
+    await _checkHomeWidget();
```
ashilkn: What's the context of adding this? Is it for updating the widget when app in opened after it is killed?

prateekmedia: So basically after app login we want to populate the favorites in home as soon as favorite service is initialized. this may be cleaned up a bit to don't conflict with initialization of favorites in background fetch.

mobile/lib/main.dart
prateekmedia: On login two services will try to populate home, one is this and other is favorite service initFav. So I had to resolve this conflict by delaying this.

ashilkn (approved):

PR #1000, opened 2024-03-11
[server] Release previous locks taken by host on statup

ua741:

### DIFF
--- a/server/cmd/museum/main.go
+++ b/server/cmd/museum/main.go
@@ -682,7 +682,7 @@ func main() {
 	publicAPI.GET("/offers/black-friday", offerHandler.GetBlackFridayOffers)
 
 	setKnownAPIs(server.Routes())
-
+	lockController.ReleaseHostLock()
 	setupAndStartBackgroundJobs(objectCleanupController, replicationController3)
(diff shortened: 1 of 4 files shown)


server/cmd/museum/main.go
```diff
@@ -682,7 +682,7 @@ func main() {
 	publicAPI.GET("/offers/black-friday", offerHandler.GetBlackFridayOffers)
 
 	setKnownAPIs(server.Routes())
-
+	lockController.ReleaseHostLock()
```
mnvr, line 685: We'll need to guarantee that none of the `NewFooController` methods above take any locks before we run this line. I don't think they do, and you've likely already checked, but just mentioning in case.

mnvr (approved):

server/cmd/museum/main.go
ua741: Yes, I did check this part. May be it make sense to move it at the top [right after db migration], to reduce the chance of regressions. Does that make sense?

mnvr: Can do that too, yes, maybe that's safer.

PR #1005, opened 2024-03-11
[auth][mob] Format longer codes with length > 6

### LINKED ISSUE #475: [Feature Request] Spaces in 7 & 8 digit codes
6 digit TOPT codes contain spaces to make it easier to read/recall when manually entering codes, 7 & 8 digit codes do not.

As an example, having come from Authy - their layouts are:
12 **345** 67
123 **45** 678
(n.b. central blocks appear bold in authy desktop app as well which also helps the readability/UX)

Between the phone app and the web interface it is possible to copy-paste in most situations avoiding having to read/remember codes, so is principally a cosmetic/UX issue & does not affect functionality.

EDIT: also just noticed that the web interface does not have spaces in 6 digit codes either.  It would be nice for both the Web and the App UIs to have a consistent look & feel/display layout.

AbhishekSharma55:
## Description
This modification adds a space after every third character in the PinPut widget, making the 8-digit code visually separated into blocks of 3 characters. Additionally, it adjusts the verification logic to remove spaces from the code before verifying it.

## Tests

fixes #475

### DIFF
--- a/auth/lib/ui/two_factor_authentication_page.dart
+++ b/auth/lib/ui/two_factor_authentication_page.dart
@@ -69,6 +69,7 @@ class _TwoFactorAuthenticationPageState
       ),
       borderRadius: BorderRadius.circular(15.0),
     );
+
     return Column(
       crossAxisAlignment: CrossAxisAlignment.stretch,
       mainAxisAlignment: MainAxisAlignment.center,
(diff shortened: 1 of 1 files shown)


ua741:
Hello,

The change needs to be made on the screen where we display code for various screen.
This is method that we should modify.

https://github.com/ente-io/ente/blob/main/auth/lib/ui/code_widget.dart#L448
If the code is between [7,9] digits, let's keep the middle group small.
123 4 567
123 45 678
123 456 789

If it's larger than 9, let's just divide it in group of 3, with smallest group in the end.

123 456 789 1 
123 456 789 12
123 456 789 123


Let's add a method in `totp_util` that will return the formatted code. It will be great if you can add a test + screenshot for the screen to see how it looks visually.
```dart
String getGroupedCode(code) {

}
```

Thanks!

ua741:
Hey @AbhishekSharma55 ,

As this PR anyways contain redundant change, we will close it. Please feel free to create a fresh PR with the suggested changes as mentioned in https://github.com/ente-io/ente/pull/1005#issuecomment-1987923730

Thanks

PR #1065, opened 2024-03-12
Add Dockerfile for the web app

Bramas:
## Description

A dockerfile for the web application to build and prepare a ready to deploy container with customizable endpoints.

API endpoints and Albums endpoints can be set a the execution of the container with two environment variables:
* `ENTPOINT` the museum endpoint
* `ALBUMS_ENDPOINT` another host that points to the same front-end.

By default the nginx base image expose port 80 so an image build with the given docker file can be used like this:

```bash
docker build -t ente-web web  --build-arg git_sha=AN_OPTIONNAL_GIT_SHA
docker run -p 8081:80 -p 8081:80 -e ENDPOINT=<your_museum_endpoint> -e ALBUMS_ENDPOINT=http://localhost:8082 ente-web
```

I had to modify `web/packages/next/next.config.base.js` in order to be able to set the GIT_SHA without calling git.

### DIFF
--- a/web/.dockerignore
+++ b/web/.dockerignore
@@ -1,0 +1,4 @@
+.git
+.dockerignore
+node_modules
+Dockerfile
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -1,0 +1,59 @@
+# syntax=docker/dockerfile:1
+
+FROM node:21-bookworm-slim as ente-builder
+
+WORKDIR /app
+
+RUN apt update && apt install -y ca-certificates && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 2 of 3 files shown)


mnvr:
I've cherry-picked the change to next.config.base.js (well, effectively something that get the same functional result by making the git sha an optional property) in https://github.com/ente-io/ente/pull/1073

As for the dockerfile, we'll let that be part of the docs only for now, but I'll definitely continue thinking more how to integrate this better in the repo itself (https://github.com/ente-io/ente/pull/1066#issuecomment-1993453853).

Thank you, again 🙏

PR #1066, opened 2024-03-12
add guide for self-hosting server + webapp with external S3 storage

Bramas:
## Description

I wrote a small guide to run the server and the web app using docker compose (locally or on a server), with an external S3 storage (which I assume will be a common use-case).

It requires #1065 , or at least this line change: https://github.com/ente-io/ente/pull/1065/files#diff-5a9426639947f5afb92612a4583a5b7d496f9cb6a791db9c29f28ff298282aebR16

### DIFF
--- a/docs/docs/self-hosting/guides/index.md
+++ b/docs/docs/self-hosting/guides/index.md
@@ -7,3 +7,235 @@ description: Guides for self hosting Ente Photos and/or Ente Auth
 
 If you've figured out how to do something, help others out by adding
 walkthroughs, tutorials and other FAQ pages in this directory.
+
+This guide is for self hosting the server and the web application of Ente Photos using docker compose and an external S3 bucket. So we assume that you already have the keys and secrets for the S3 bucket.
+The plan is as follows:
+
+1. create a `compose.yaml` file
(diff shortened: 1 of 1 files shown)


mnvr:
Thank you so much for this ☺️ This is the first guide someone outside of us has written for running the self hosted instance in an non-trivial setup. Thank you for the new perspective.

I'll merge this (after merging I'll move it to a separate file, the index is sort of meant as a TOC only).

Also, thanks for including the web dockerfile verbatim in the guide. I'm not sure yet about adding the web Dockerfile to the repo itself (I understand that it is useful in some cases, but I also want to make sure I fully understand how all the moving parts will tie together before I commit to adding it in the workflow and maintaining it. Perhaps there is some sort of a way of a _separate_ folder that has both the web + server dockerfiles and docker-compose yml, as you were hinting at in Discord. All this is to say - I think it need a bit more thought before committing to it).

The other change you called out - about overriding the GIT_SHA, that I'll cherry pick and add.

Thanks again!

mnvr (approved):

coracoo:
Looking forward to deploying ENTE with docker

PR #1067, opened 2024-03-12
Specify which Google Takeout format is supported

tgf9:
Currently, the upload menu says it accepts Google Takeout. Unfortunately, it's not clear to the user which format, ZIP or TGZ, is actually supported. This can lead a user to accidentally request Takeout in the wrong format, which is especially painful if the user has a large library that takes a long time to export.

Perhaps both ZIP and TGZ should be supported. At the very least, this change explicitly tells the user which format is supported, so they don't waste a day exporting the wrong format.

### DIFF
--- a/web/apps/accounts/public/locales/de-DE/translation.json
+++ b/web/apps/accounts/public/locales/de-DE/translation.json
@@ -418,7 +418,7 @@
     "DOWNLOAD_UPLOAD_LOGS": "",
     "UPLOAD_FILES": "Datei",
     "UPLOAD_DIRS": "Ordner",
-    "UPLOAD_GOOGLE_TAKEOUT": "Google Takeout",
+    "UPLOAD_GOOGLE_TAKEOUT": "Google Takeout (ZIP)",
     "DEDUPLICATE_FILES": "",
     "AUTHENTICATOR_SECTION": "Authenticator",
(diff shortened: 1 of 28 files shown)


mnvr:
While I understand that this tripped you up, I am not sure about this change. I might be an outlier, but I feel every extra button and word in the UI has a cognitive cost, and I'm not sure this one is needed:

1. The upload will immediately fail if someone tries to upload a tar.gz
2. If someone did export their data as a .tar.gz, converting it into a .zip on their local machine is generally quick
3. The [existing FAQ](https://ente.io/faq/migration/from-google-photos/) already mentions that the format needs to be ZIP (which is also the default that Google selects AFAIK)

That said, I do feel some improvement can be done. The best is that we can start supporting .tar.gz (it won't be too much work, but it'll increase the size of our binary for everyone). Meanwhile I'll try to modify the FAQ to highlight the ZIP file requirement a bit more (it already mentions this, but that can perhaps be highlighted better).

mnvr:
I've called out the need to select the ZIP file as the export format better in the [FAQ](https://ente.io/faq/migration/from-google-photos/). And I'll keep thinking if there is a better way in which we can surface this in the app's UI (without needing to put a ZIP everywhere).

Thanks for pointing out the issue!

PR #1088, opened 2024-03-13
[auth] Authy guide rewrite

greeeen-dev:
## Description
Authy migration guide rewrite to make things much clearer and easier to follow.
## Tests
N/A


greeeen-dev:
I might try to see if I can get the guide up on the new help page... 🤔

mnvr (approved):
You're right - having it duplicated isn't best. I'll merge this though. Then what I'll do is 

* Also move the other migration guide to the docs
* Remove the content of these files, and add a link to the new wiki instead

But all that can be done after merging this. Thank you again for your help!

PR #1096, opened 2024-03-14
[meta] Document the tag format, and call out the need to keep tag == title

mnvr:
Related: https://github.com/ente-io/ente/discussions/1087#discussioncomment-8775807

### DIFF
--- a/mobile/docs/release.md
+++ b/mobile/docs/release.md
@@ -4,23 +4,21 @@ Create a PR to bump up the version in `pubspec.yaml`.
 
 > [!NOTE]
 >
-> Use [semver](https://semver.org/) for the tags, with `auth-` as a prefix.
+> Use [semver](https://semver.org/) for the tags, with `photos-` as a prefix.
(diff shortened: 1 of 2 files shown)


mobile/docs/release.md
```diff
@@ -0,0 +1,32 @@
...
+Once that is merged, tag main, and push the tag.
+
+```sh
+git tag photos-v1.2.3
+git push origin photos-v1.2.3
+```
+
+This'll trigger a GitHub workflow that:
+
+* Creates a new draft GitHub release and attaches the build artifacts to it
```
mnvr, line 20: @vishnukvmd I've kept this document separate instead of linking to the auth's docs, even though it is mostly a copy paste. This is because the release processes are divergent currently. We can also document the f-droid stuff here (one liners are enough, just how the f-droid build gets triggered) and mention how people can verify the authenticity of the build by looking for the sha in the GitHub workflow logs. Then we can just give out a link to this doc for people looking for build authenticity.

PR #1130, opened 2024-03-18
[museum] Support running behind Nginx

mnvr:
- Move deployment related stuff into a separate folder
- Document the overall approach better
- Add an Nginx specific environment and service definition
- Remove the restart limiter


server/cmd/museum/main.go
```diff
@@ -712,7 +712,7 @@ func main() {
 }
 
 func runServer(environment string, server *gin.Engine) {
-	if environment == "local" {
+	if environment == "local" || environment == "proxy" {
```
mnvr: @ua741 This is the only source change (apart from packaging a proxy.yaml, duplicate of production.yaml). This felt like a more robust solution to start with instead of using config params, but can tinker too. Untested yet, will do a test deployment after building from this branch.

ua741 (approved):

server/cmd/museum/main.go
```diff
@@ -712,9 +712,8 @@ func main() {
 }
 
 func runServer(environment string, server *gin.Engine) {
-	if environment == "local" {
-		server.Run(":8080")
-	} else {
+	useTLS := viper.GetBool("http.use-tls")
```
mnvr, line 715: Updated the PR to read TLS setting from config instead of creating a new environment.

**Tests**

* Default settings are unchanged

```sh
$ docker compose up --build
$ curl http://localhost:8080/ping
```

* Configuration value is used

```sh
# Edit compose.yaml to add
    environment:
      ENTE_HTTP_USE-TLS: 1

# Then
$ docker compose up

# Fails as expected (since it cannot find the TLS creds)
...[GIN-debug] Listening and serving HTTPS on :443
...FATA[0000]main.go:727 runServer open : no such file or directory
```

PR #1152, opened 2024-03-19
[server] Fix Stripe related issues

abhinavkgrd:
### Stripe
- Settings have been modified to "Cancel" the subscription if all attempts to charge an invoice fails. The previous configuration was to set the subscription to "Past Due".

### Museum
- Updated our Stripe workflow to use `PaymentBehaviorAllowIncomplete` in case of SEPA payments
- Fixed the issue caused by us spawning unnecessary subscriptions when a customer tried to renew their expired subscription
- Removed the email that was being sent out when a Stripe subscription expired. We will have to run our own cron to send a notification and perform necessary clean up. When payments fail, we will still send a notification to intimate the customer that their account is on hold.

### Mobile
- Fixed the dialogs and error messages on the payment page

## Tests
The following cases were tested for SEPA, Credit Cards, and Credit Cards with 3DS.
- [x] Purchase
  - [x] Success
  - [x] Failure
- [x] Renewal
  - [x] Success
  - [x] Failure
- [x] Upgrades
  - [x] Success
  - [x] Failure
- [x] Cancellation
- [x] Resumption post recovery

## Note

These changes are likely to break in the event of some corner cases we haven't thought of. But this will land us in a state that is better than where we are right now in terms of collecting payments for expired subscriptions and those via SEPA.


server/pkg/controller/stripe.go
```diff
@@ -341,8 +345,18 @@ func (c *StripeController) handleCustomerSubscriptionUpdated(event stripe.Event)
...
+	if err != nil {
+		return ente.StripeEventLog{}, stacktrace.Propagate(err, "")
+	}
+	if currentSubscription.ProductID == newSubscription.ProductID {
+		// Webhook is reporting an outdated update that was already verified
+		// no-op
+		log.Warn("Webhook is reporting an outdated purchase that was already verified stripeSubscriptionID:", stripeSubscription.ID)
+		return ente.StripeEventLog{UserID: userID, StripeSubscription: stripeSubscription, Event: event}, nil
+	}
+	c.BillingRepo.ReplaceSubscription(currentSubscription.ID, newSubscription)
```
abhinavkgrd: this was required to handle the of failed sepa update, where the user's subscription has directly changed to past_due, 

In which our earlier code would not updated the local museum subscription details

server/pkg/controller/stripe.go
```diff
@@ -458,14 +522,15 @@ func (c *StripeController) UpdateSubscription(stripeID string, userID int64) (en
...
-	if newStripeSubscription.PendingUpdate != nil {
-		switch newStripeSubscription.LatestInvoice.PaymentIntent.Status {
-		case stripe.PaymentIntentStatusRequiresAction:
+	if newStripeSubscription.Status == stripe.SubscriptionStatusPastDue {
+		if newStripeSubscription.LatestInvoice.PaymentIntent.Status == stripe.PaymentIntentStatusRequiresAction {
 			return ente.SubscriptionUpdateResponse{Status: "requires_action", ClientSecret: newStripeSubscription.LatestInvoice.PaymentIntent.ClientSecret}, nil
-		case stripe.PaymentIntentStatusRequiresPaymentMethod:
+		} else if newStripeSubscription.LatestInvoice.PaymentIntent.Status == stripe.PaymentIntentStatusRequiresPaymentMethod {
 			inv := newStripeSubscription.LatestInvoice
 			invoice.VoidInvoice(inv.ID, nil)
```
abhinavkgrd, line 530: here we should call particular client specific invoice api

server/pkg/controller/stripe.go
```diff
@@ -409,6 +408,70 @@ func (c *StripeController) handleInvoicePaid(event stripe.Event) (ente.StripeEve
...
+	var paymentIntent stripe.PaymentIntent
+	json.Unmarshal(event.Data.Raw, &paymentIntent)
+	// Figure out the user
+	invoiceID := paymentIntent.Invoice.ID
+	client := c.StripeClients[country]
+	invoice, err := client.Invoices.Get(invoiceID, nil)
+	if err != nil {
+		return ente.StripeEventLog{}, stacktrace.Propagate(err, "")
+	}
+	// void the invoice, in case the payment intent failed
```
abhinavkgrd: can't void the invoice as it leads to subscription going into a active state, instead of expired / payment_due

server/pkg/controller/stripe.go
```diff
@@ -458,14 +521,15 @@ func (c *StripeController) UpdateSubscription(stripeID string, userID int64) (en
...
 			return ente.SubscriptionUpdateResponse{Status: "requires_action", ClientSecret: newStripeSubscription.LatestInvoice.PaymentIntent.ClientSecret}, nil
-		case stripe.PaymentIntentStatusRequiresPaymentMethod:
+		} else if newStripeSubscription.LatestInvoice.PaymentIntent.Status == stripe.PaymentIntentStatusRequiresPaymentMethod {
 			inv := newStripeSubscription.LatestInvoice
-			invoice.VoidInvoice(inv.ID, nil)
+			client.Invoices.VoidInvoice(inv.ID, nil)
 			return ente.SubscriptionUpdateResponse{Status: "requires_payment_method"}, nil
+		} else if newStripeSubscription.LatestInvoice.PaymentIntent.Status == stripe.PaymentIntentStatusProcessing {
+			return ente.SubscriptionUpdateResponse{Status: "success"}, nil
 		}
```
abhinavkgrd, line 495: same reason as above, can't void the invoice, will lead to bad state

vishnukvmd (approved):

PR #1171, opened 2024-03-21
[photos][mobile] Add support for viewing HEIC images with proXDR

ashilkn:
## Description

New Oneplus devices have a "ProXDR" feature when viewing images. These images (when in HEIC format) decode fine on devices that supports ProXDR, but fails to decode on other devices. 

So if decoding fails, we convert it to a JPEG and use that image for viewing it. 

## Tests

- [x] Tries converting to jpeg only if decoding fails
- [x]  If converting also fails, the behaviour remains the same as before when decoding fails.


mobile/lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -239,7 +242,10 @@ class _ZoomableImageState extends State<ZoomableImage> {
             _isGIF(), // since on iOS GIFs playback only when origin-files are loaded
       ).then((file) {
         if (file != null && file.existsSync()) {
-          _onFinalImageLoaded(Image.file(file).image);
+          _onFinalImageLoaded(
```
vishnukvmd: `_onFinalImageLoaded` can create the `ImageProvider` with the `file`, so we could reduce one parameter? It would then make sense to rename the method to `_onFileLoaded`.

mobile/lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -348,4 +366,36 @@ class _ZoomableImageState extends State<ZoomableImage> {
...
+    _logger.info("Compressing ${_photo.displayName} to viewable format");
+    _convertToSupportedFormat = true;
+
+    final compressedFile =
+        await FlutterImageCompress.compressWithFile(file.path);
+
+    if (compressedFile != null) {
+      final imageProvider = MemoryImage(compressedFile);
+
+      unawaited(
```
vishnukvmd, line 384: We could reuse `_onFileLoaded` here?

mobile/lib/ui/viewer/file/zoomable_image.dart
ashilkn: Have made these changes.

mobile/lib/ui/viewer/file/zoomable_image.dart
ashilkn: I feel it's better to keep them separate, there are multiple differences here and in `_onFileLoaded`.
- `onError` closure of `precacheImage` in `_onFileLoaded`.
- `compressedFile` is a `Uint8List?` and `file` in `_onFileLoaded` is a `File` object.
- Here, we are converting `compressedFile` to `ImageProvider` in one step. Will have to write multiple steps to covert it into a `File` object so that it could be passed to `_onFileLoaded`.

vishnukvmd: Got it, we could maybe extract the following bit into a single function?

```dart
await _updatePhotoViewController(
  previewImageProvider: _imageProvider,
  finalImageProvider: imageProvider,
);
setState(() {
  _imageProvider = imageProvider;
  _loadedFinalImage = true;
  _logger.info("Final image loaded");
});
```

ashilkn: Done

vishnukvmd (approved):

PR #1175, opened 2024-03-22
Fix an error in the order of the arguments for imagemagick

MatMaul:
## Description

When importing a folder, I am getting this error on each image:
```
  cmd: "'/tmp/.mount_ente-xHvuBzN/resources/image-magick' -auto-orient -define 'jpeg:size=1440x1440' 'IMG_9694.heic' -thumbnail '720x720>' -unsharp '0x.5' -quality 70 '/tmp/ente/E6YH91g3T3-thumb.jpeg'",
  stdout: '',
  stderr: "magick: no images found for operation `-auto-orient' at CLI arg 1 @ error/operation.c/CLIOption/5432.\n"
```

## Tests

This is UNTESTED because I don't have a build env ready, however it seems like this error is coming out because we pass the input file after specifying some modifier flag that applies to it, and we should put the input before.


mnvr (approved):
Thanks @MatMaul!

PR #1185, opened 2024-03-23
Fix cases

vishnukvmd:
## Description
The intent was to lower the case for "Files" and "Each" (that were in caps), since the file was open, also ended up flipping the case for "ente".

## Tests
Have verified the places "files" and "each" were being printed. As for "Ente", the change should not cause any side effects.


mnvr:
We'll need to manually update the translations. I'll do that after merging this.

mnvr (approved):

mnvr:
> We'll need to manually update the translations.

In Crowdin I mean.

vishnukvmd:
I thought if the value corresponding to a key changed, Crowdin would expire the older string? 🤔

mnvr:
> I thought if the value corresponding to a key changed, Crowdin would expire the older string?

You're right. But in these cases the rest of the string remains unchanged, and I can save the translators a bunch of work by updating it manually in these cases.

Issue #1191, opened 2024-03-23
[web/desktop] Some commands may not work in the Windows environment.


LI-NA:
### Description

Hi there.

I'm currently trying to develop and build ente in a Windows environment.

I've been able to successfully build mobile apps without any issues, but on the web or desktop I'm having problems due to the different environments.

First, look at the next config file on the web (`web/packages/next/next.config.base.js`), using the following command to get the git head.

```
git rev-parse --short HEAD 2>/dev/null || true
```

These commands will not work in the Windows environment, so you will need to modify them with the following commands.

```
git rev-parse --short HEAD 2> NUL || cd .
```

I came up with the following code, which I'll turn into a PR if you point me in the right direction.

```
const result = cp
        .execSync(
            "git rev-parse --short HEAD 2> " + (os.platform() === "win32" ? "NUL || cd ." : "/dev/null || true"),
            {
                cwd: __dirname,
                encoding: "utf8",
            },
        )
        .trimEnd();

// or?

try {
    const result = cp
        .execSync("git rev-parse --short HEAD", {
            cwd: __dirname,
            encoding: "utf8",
        })
        .trimEnd();

    return result;
}catch (_error) {
    return undefined;
}
```

-----------

The next problem is in the desktop. In `package.json` there are commands to delete file and create symbolic link.

```
cd ../web && yarn install && yarn build:photos && cd ../desktop && rm -f out && ln -sf ../web/apps/photos/out
```

Similarly, the rm and ln commands do not work on Windows. So I'm trying to use the [shx](https://www.npmjs.com/package/shx) package to support cross-platform like this:

```
cd ../web && yarn install && yarn build:photos && cd ../desktop && shx rm -f out && shx ln -sf ../web/apps/photos/out out
```

Since this is a new package installation, I've raised it as a issue rather than a PR. Please let me know what you think and I'll fix it.

Thanks for checking it out! I'm looking forward to the day when ente can support multiplatform in the development environment :)

### Version

main

### What product are you using?

Ente Photos

### What platform are you using?

Desktop - Windows

mnvr:
Thank you!

For both web/desktop cases, I think the approach you're mentioning is the correct one, and since you've done the hard work figuring this out, it'll be great if you can create a PR for this too.

For the web case, I think the first option is the better one, since with the second option there will be spurious errors printed to the console. However, for readability's sake, I think we should split out the ternary into a separate variable, something like this (maybe you already had this in mind, I understand what you showed were just prototype examples):

[...]

PR #1198, opened 2024-03-25
[web/desktop] Support windows environment on command

LI-NA:
## Description

This is my first time in ente! I modified two commands to allow the build scripts for web and desktop to work in a Windows environment.

- Before running the command to get the git SHA in `next.config.js`, I checked the platform of the current OS and changed it so that we could run a command that would work on Windows.
  - `NUL` is same as `/dev/null` in Windows, and `cd .` is always return nothing with 0 status code.
- Added a new package called [`shx`](https://[github.com/shelljs/shx](https://github.com/shelljs/shx)) to the development dependencies to support `rm` and `ln`, which are commands not supported by Windows when running the Desktop build command.

#1191 

## Tests

I've tested the web build and desktop build in a Windows environment. I don't have a Mac, so I'd love for someone to test it on a Mac and let me know the results.

### DIFF
--- a/desktop/docs/dependencies.md
+++ b/desktop/docs/dependencies.md
@@ -52,3 +52,6 @@ Some extra ones specific to the code here are:
 
 * [concurrently](https://github.com/open-cli-tools/concurrently) for spawning
   parallel tasks when we do `yarn dev`.
+* [shx](https://github.com/shelljs/shx) for providing a portable way to use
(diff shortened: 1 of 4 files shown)


mnvr (approved):
Thank you!

**Tests on macOS**

After doing a `yarn install` (to pick up the new dependency):

- [x] Commit SHA is printed in web version
- [x] Desktop `yarn build-renderer` creates a symlink

PR #1228, opened 2024-03-27
Add SMTP for sending emails

vs704vs:
## Tests
Tested for templated mails in user.go, userauth.go, stripe.go, public_collection.go, playstore.go, admin.go, email_notification.go

Screenshots for all the templates - [https://drive.google.com/drive/folders/1f7BjM5MHicRUbr5jlCe2-4MGz7s1moLw?usp=sharing](url)

### DIFF
--- a/server/configurations/local.yaml
+++ b/server/configurations/local.yaml
@@ -157,21 +157,28 @@ key:
 jwt:
     secret: i2DecQmfGreG6q1vBj5tCokhlN41gcfS2cjOs9Po-u8=
 
+# SMTP configuration (optional)
+#
+# Configure credentials here for sending mails from museum (e.g. OTP emails).
+#
+# The smtp credentials will be used if the host is specified. Otherwise it will
+# try to use the transmail credentials. Ideally, one of smtp or transmail should
+# be configured for a production instance.
+smtp:
+    host:
+    port:
+    username:
+    password:
+
 # Zoho Zeptomail config (optional)
-# Use case: Sending emails
+#
+# This is an alternative to the `smtp` configuration for sending emails. If this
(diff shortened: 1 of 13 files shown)


server/configurations/local.yaml
```diff
@@ -164,6 +164,15 @@ transmail:
     # Mail agent: dev
     key:
 
+# AWS SES
+# Use case: Sending mails
+ses-smtp:
```
vishnukvmd: Since we are using plain simple SMTP, we could remove the SES prefix

server/pkg/utils/email/email.go
```diff
@@ -18,6 +19,69 @@ import (
 	"github.com/spf13/viper"
 )
 
+func SendSES(toEmails []string, fromName string, fromEmail string, subject string, htmlBody string, inlineImages []map[string]interface{}) error {
```
vishnukvmd: Since this is a private method, the convention is to start with a smaller case. Also, we could call it `sendViaSMTP` (again discarding the reference to SES)

server/pkg/utils/email/email.go
```diff
@@ -18,6 +19,69 @@ import (
 	"github.com/spf13/viper"
 )
 
+func sendViaSMTP(toEmails []string, fromName string, fromEmail string, subject string, htmlBody string, inlineImages []map[string]interface{}) error {
```
vishnukvmd, line 32: nit: It's nice to push private methods to the bottom of the file, so that if you're using an IDE that shows an outline of the methods within the file, you can easily find out the public APIs it exposes

mnvr (approved):

PR #1278, opened 2024-04-01
[mobile][photos] Prepare for release v0.8.73

ashilkn:


mobile/lib/services/update_service.dart
```diff
@@ -16,7 +16,7 @@ class UpdateService {
   static final UpdateService instance = UpdateService._privateConstructor();
   static const kUpdateAvailableShownTimeKey = "update_available_shown_time_key";
   static const changeLogVersionKey = "update_change_log_key";
-  static const currentChangeLogVersion = 16;
+  static const currentChangeLogVersion = 17;
```
vishnukvmd, line 19: Does this need to change?

ashilkn: This needs to change for the user to see the change log in the bottom sheet. Just realised that there was no full release with logVersion 16, so yes, the change was not actually needed. But let's merge this change now as I've used logVersion 17 for the internal build that's out, and if everything is good we'll be promoting the release with logVersion 17.

vishnukvmd (approved):

PR #1289, opened 2024-04-02
Listmonk mailing list control 

vs704vs:
## Description
Add listmonk mailing list subscribe and unsubscribe

## Tests
Tested getSubscriberID, unsubscribe, and subscribe using a locally running listmonk server with different parameters.
- [x] A new subscriber is created and added to the listmonk campaign mailing list on listmonkSubscribe() for the given list IDs
- [x] Subscriber is removed from listmonk campaign mailing list on listmonkUnsubscribe() for the given list IDs
- [x] The old email address is unsubscribed, and the new email is subscribed when a user updates the email

### DIFF
--- a/server/pkg/controller/mailing_lists.go
+++ b/server/pkg/controller/mailing_lists.go
@@ -13,23 +13,23 @@ import (
 	"github.com/spf13/viper"
 )
 
-// ZohoMailingListsController is used to keeping the external mailing lists in sync
+// MailingListsController is used to keeping the external mailing lists in sync
 // with customer email changes.
 //
-// ZohoMailingListsController contains methods for keeping external mailing lists in
+// MailingListsController contains methods for keeping external mailing lists in
 // sync when new users sign up, or update their email, or delete their account.
 // Currently, these mailing lists are hosted on Zoho Campaigns.
 //
 // See also: Syncing emails with Zoho Campaigns
-type ZohoMailingListsController struct {
+type MailingListsController struct {
 	zohoAccessToken string
 	zohoListKey     string
 	zohoTopicIds    string
 	zohoCredentials zoho.Credentials
 }
 
-// Return a new instance of ZohoMailingListsController
-func NewZohoMailingListsController() *ZohoMailingListsController {
+// Return a new instance of MailingListsController
+func NewMailingListsController() *MailingListsController {
 	zohoCredentials := zoho.Credentials{
 		ClientID:     viper.GetString("zoho.client-id"),
 		ClientSecret: viper.GetString("zoho.client-secret"),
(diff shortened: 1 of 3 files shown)


server/pkg/controller/mailing_lists.go
```diff
@@ -158,3 +186,56 @@ func (c *MailingListsController) doListAction(action string, email string) error
...
+	// Listmonk dosen't provide an endpoint for unsubscribing users from a particular list
+	// directly via their email
+	//
+	// Thus, fetching subscriberID through email address,
+	// and then calling endpoint to modify subscription in a list
+	id, err := listmonk.GetSubscriberID(c.BaseURL+"/api/subscribers", c.Username, c.Password, email)
+	if err != nil {
+		stacktrace.Propagate(err, "")
+	}
+	// API endpoint expects an array of subscriber id as paarmeter
```
vishnukvmd: minor: typo in parameter

server/pkg/controller/mailing_lists.go
```diff
@@ -65,6 +66,31 @@ func NewMailingListsController() *MailingListsController {
...
+// and ListIDs (an array of integer values indicating the id of listmonk campaign mailing list
+// to which the subscriber needs to added)
+type ListmonkMailingListsController struct {
+	BaseURL  string
+	Username string
+	Password string
+	ListIDs  []int
+}
+
+// NewListmonkMailingListsController creates a new instance of ListmonkMailingListsController
```
vishnukvmd: Who calls this?

server/pkg/controller/mailing_lists.go
```diff
@@ -158,3 +186,56 @@ func (c *MailingListsController) doListAction(action string, email string) error
 
 	return stacktrace.Propagate(err, "")
 }
+
+// Add or subscribe an email to listmonk mailing list
+func (c *ListmonkMailingListsController) Subscribe(email string) error {
```
vishnukvmd: Who calls `Subscribe` and `Unsubscribe`?

server/pkg/controller/mailing_lists.go
```diff
@@ -89,32 +115,34 @@ func (c *MailingListsController) Subscribe(email string) error {
 	//   any confirmations.
 	//
 	// https://www.zoho.com/campaigns/help/developers/contact-subscribe.html
-	return c.doListAction("listsubscribe", email)
+	return c.doListActionZoho("listsubscribe", email)
```
vishnukvmd: We could trigger listmonk actions as well here

server/pkg/controller/mailing_lists.go
```diff
@@ -76,45 +92,68 @@ func NewMailingListsController() *MailingListsController {
...
+		// Need to set "Signup Form Disabled" in the list settings since we use this
+		// list to keep track of emails that have already been verified.
+		//
+		// > You can use this API to add contacts to your mailing lists. For signup
+		//   form enabled mailing lists, the contacts will receive a confirmation
+		//   email. For signup form disabled lists, contacts will be added without
+		//   any confirmations.
+		//
+		// https://www.zoho.com/campaigns/help/developers/contact-subscribe.html
+		return c.doListActionZoho("listsubscribe", email)
```
vishnukvmd: So the issue we discussed offline will get triggered here as well, since you're `return`-ing. If both Zoho and Listmonk are configured, we will return after performing the action for Zoho.

Instead, we should do Zoho, return an error if there's any. Then do Listmonk, return an error if there's any. And finally return `nil`.

mnvr (approved):

PR #1302, opened 2024-04-03
[Photos] Allow for configuring a custom server

vishnukvmd:
## Description
Users can now tap on the onboarding screen 7 times to bring up a page where they can configure the endpoint the app should be connecting to.
photos-selfhost

## Tests
- [x] Verified that production flows are working as expected
- [x] Verified that configuring the endpoint to a local instance lets you
  - [x] Connect to that instance
  - [x] Create an account
  - [x] Upload a photo
  - [x] Logout and log back in

### DIFF
--- a/mobile/ios/Podfile.lock
+++ b/mobile/ios/Podfile.lock
@@ -152,6 +152,8 @@ PODS:
   - path_provider_foundation (0.0.1):
     - Flutter
     - FlutterMacOS
+  - permission_handler_apple (9.1.1):
+    - Flutter
   - photo_manager (2.0.0):
     - Flutter
     - FlutterMacOS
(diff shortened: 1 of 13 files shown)


ua741 (approved):

alexandrestein:
Hello,

That works well for the phone application. (I tried with the Android)
But I was not able to do the same on the descktop applicaiton.

Is that supported?

mnvr:
It is not supported in the desktop app yet. We are planning to add that soon. You can watch this discussion to be notified when it is supported in the desktop app - https://github.com/ente-io/ente/discussions/504.

jo-chemla:
Currently, the endpoint string displayed in the text field before user inputs any text is the previous one - eg probably the one the user has setup like `https://localhost:8080`. If one want to switch back to ente, it could be useful to add a description for that field, which indicates the official ente api endpoint url in case one want to switch back to the ente API rather than the custom endpoint. 

For anyone stumbling on the issue here, replace that custom endpoint with `https://api.ente.io`

PR #1320, opened 2024-04-04
[server] Build ARM images when publishing to ghcr.io

mnvr:
Untested yet, will test alongside the next publish

Requested in https://github.com/ente-io/ente/discussions/1305

### DIFF
--- a/.github/workflows/server-publish.yml
+++ b/.github/workflows/server-publish.yml
@@ -32,6 +32,8 @@ jobs:
                   image: server
                   registry: ghcr.io
                   enableBuildKit: true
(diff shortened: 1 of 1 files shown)


mnvr:
This seems to be working, the new images got published with the new OS/arch combos - https://github.com/ente-io/ente/pkgs/container/server

It however has slowed down the github workflow for publishing by a lot, it takes around 20 mins now. This is expected - the documentation of the docker action we use calls out that doing cross platform builds using qemu emulation will be slow. Since this is a relatively infrequent workflow, the slowdown is acceptable if these alternative images are being useful to someone.

PR #1399, opened 2024-04-10
[web] Import ONNX-YOLO face changes from the web_face branch

mnvr:
Laurens has made the relevant changes to get ONNX-YOLO face detection working in a manner where the generated embeddings are the same as what get generated by the corresponding ML stack in the mobile client.

This PR cherry picks his ML related changes from the branch he was working in, but leaves out the surrounding scaffolding (We cannot merge that branch directly because it relies on wasm that we don't need and don't want to commit to main).

At this point this functionality is correct but not usable - the next step will be to tie this to the ONNX runtime that we already have on the Node.js layer of our desktop app.

**Tests**

- yarn dev + smoke test
- yarn build + smoke test
- yarn lint

### DIFF
--- a/web/apps/photos/src/constants/mlConfig.ts
+++ b/web/apps/photos/src/constants/mlConfig.ts
@@ -13,7 +13,6 @@ export const DEFAULT_ML_SYNC_CONFIG: MLSyncConfig = {
     imageSource: "Original",
     faceDetection: {
         method: "YoloFace",
-        minFaceSize: 32,
     },
     faceCrop: {
         enabled: true,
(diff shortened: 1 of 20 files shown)


web/apps/photos/src/constants/mlConfig.ts
```diff
@@ -12,7 +12,7 @@ export const DEFAULT_ML_SYNC_CONFIG: MLSyncConfig = {
     batchSize: 200,
     imageSource: "Original",
     faceDetection: {
-        method: "BlazeFace",
+        method: "YoloFace",
         minFaceSize: 32,
```
laurenspriem: Can remove this minFaceSize as well, we're not using it anymore for YOLOface

web/apps/photos/src/services/machineLearning/mlWorkManager.ts
```diff
@@ -30,7 +30,7 @@ class MLWorkManager {
 
     constructor() {
         this.liveSyncQueue = new PQueue({
-            concurrency: 1,
+            concurrency: 4,
```
laurenspriem: This change doesn't come from me. I remember in the past there were some issues with concurrency, so we should be careful with this.

laurenspriem (approved):

PR #1434, opened 2024-04-13
[mob][photos] Upgrade flutter

ashilkn:
## Description

- Update flutter
- Update packages
- Fix UI issues because of update (material 3)
- bump up version to v0.8.80

## Tests

- [x] Building on android and iOS.


mobile/android/app/build.gradle
```diff
@@ -6,10 +11,10 @@ if (localPropertiesFile.exists()) {
     }
 }
 
-def flutterRoot = localProperties.getProperty('flutter.sdk')
```
prateekmedia, line 9: why commented these instead of removing when they will not be used?

ashilkn: @ua741

mobile/pubspec.yaml
```diff
@@ -174,10 +174,11 @@ dependencies:
   widgets_to_image: ^0.0.2
 
 dependency_overrides:
-  # current fork of tfite_flutter_helper depends on ffi: ^1.x.x
-  # but we need ffi: ^2.0.1 for newer packages. The original tfite_flutter_helper
-  #
-  ffi: ^2.1.0
+  connectivity_plus: ^4.0.0
+  
+  #Remove this after  upgrading to flutter v3.19x
```
prateekmedia: This should be removed then as we are already bumping flutter to v3.19.x?

ua741: Seems like `flutter_sodium` depends on ^1.0.0. Once we remove dependency from that, we should be able to remove the override here.

ua741 (approved):

PR #1453, opened 2024-04-15
[web] Fix dev mode error on search input

daviddeepan:
## Description
Ran into error after clearing search bar of queries by hitting backspace.

>  The error is shown below
> Thank you to the team for considering 

## Tests
**Error**
Screenshot 2024-04-15 184234
**Fix**
Fixed by wrapping the line-causing error around by an `if` block.

PR #1457, opened 2024-04-16
[web] Clear search bar state when value is falsey

mnvr:
Updates https://github.com/ente-io/ente/pull/1453 so that we also clear the existing state by calling the setFoo methods even when value is falsey.


Jennjenn911:
Stop

PR #1458, opened 2024-04-16
[server] Specify a default for apps.public-albums

mnvr:
local.yaml is only loaded in the local environment, and will not be used in production. To avoid specifying it in both production and local configurations, use it as a default specified in the code instead (that's the pattern we're using in other places too from what I remember, maybe we should refactor that too, but that's a bigger task not tied to this change).

**Tested by**

On a local dev cluster

- [x] Not specifying anything and ensuring that the default is picked
- [x] Specifying an override in museum.yaml and verifying that is used instead

### DIFF
--- a/server/pkg/repo/public_collection.go
+++ b/server/pkg/repo/public_collection.go
@@ -23,7 +23,7 @@ type PublicCollectionRepository struct {
 // NewPublicCollectionRepository ..
 func NewPublicCollectionRepository(db *sql.DB, albumHost string) *PublicCollectionRepository {
 	if albumHost == "" {
-		panic("albumHost can not be empty")
+		albumHost = "https://albums.ente.io"
 	}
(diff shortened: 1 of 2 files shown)


server/pkg/repo/public_collection.go
```diff
@@ -23,7 +23,7 @@ type PublicCollectionRepository struct {
 // NewPublicCollectionRepository ..
 func NewPublicCollectionRepository(db *sql.DB, albumHost string) *PublicCollectionRepository {
 	if albumHost == "" {
-		panic("albumHost can not be empty")
+		albumHost = "https://albums.ente.io"
```
mnvr, line 26: @ua741 not sure if this is the best way, please tweak if needed 🙏

ua741 (approved):

PR #1489, opened 2024-04-19
[mob][photos] Use sqlite_async instead of sqlite3

ashilkn:
## Description

Using both `sqlite_async` and `sqlite3` for querying at the same time was throwing this error:
`error: SqliteException(5): while executing, database is locked, database is locked (code 5)`

So, have used `sqlite_async` everywhere.

## Tests

- [x] Tested all changes

### DIFF
--- a/mobile/lib/db/files_db.dart
+++ b/mobile/lib/db/files_db.dart
@@ -478,8 +478,8 @@ class FilesDB {
   }
 
   Future<EnteFile?> getFile(int generatedID) async {
-    final db = await instance.ffiDB;
-    final results = db.select(
+    final db = await instance.sqliteAsyncDB;
+    final results = await db.getAll(
       'SELECT * FROM $filesTable WHERE $columnGeneratedID = ?',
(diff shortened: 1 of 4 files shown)


ua741 (approved):

vishnukvmd (approved):

vishnukvmd (approved):

mobile/lib/db/files_db.dart
```diff
@@ -106,6 +106,7 @@ class FilesDB {
   static Future<sqlite3.Database>? _ffiDBFuture;
   static Future<sqlite_async.SqliteDatabase>? _sqliteAsyncDBFuture;
 
+  @Deprecated("Use ffiDB instead (sqlite_async)")
```
vishnukvmd, line 109: Should prompt to use sqliteAsyncDB instead?

ashilkn: Yes! Will fix and create a PR.

PR #1539, opened 2024-04-26
[FIX] treat steam host same as totp with 5 digits

prateekmedia:
## Description

Related: https://github.com/ente-io/ente/discussions/1038


auth/lib/models/code.dart
```diff
@@ -81,12 +86,26 @@ class Code {
   }
 
   static Code fromRawData(String rawData) {
+    bool isSteam = false;
+    if (rawData.startsWith("otpauth://steam")) {
+      rawData = rawData.replaceAll("://steam", "://totp");
```
ua741: changing the issuer here will break export. Ideally, on export we should be exporting it as otpauth://stream

prateekmedia: Done!

ua741 (approved):

ua741 (approved):

PR #1541, opened 2024-04-26
[FEAT] Pinning and Tags

prateekmedia:
## Description

Add pinning to auth


auth/lib/models/code.dart
```diff
@@ -216,7 +251,13 @@ class Code {
         secret.hashCode ^
         type.hashCode ^
         counter.hashCode ^
-        rawData.hashCode;
+        rawData.hashCode ^
+        display.hashCode;
+  }
+
+  @override
+  String toString() {
```
vishnukvmd: Not having a `toString` implementation will ensure we don't accidentally log sensitive data

auth/lib/models/code.dart
```diff
@@ -129,6 +146,28 @@ class Code {
...
+        display: CodeDisplay.fromJson(rawJson['display']),
+      );
+      return resultCode;
+    } catch (e) {
+      debugPrint("Failed to parse code from export json $e");
+      rethrow;
+    }
+  }
+
+  String toExportFormat() {
```
ua741: `toOTPAuthUrlFormat`?

auth/lib/models/code.dart
```diff
@@ -129,6 +146,28 @@ class Code {
...
 
+  static Code fromExportJson(Map rawJson) {
+    try {
+      Code resultCode = Code.fromOTPAuthUrl(
+        rawJson['rawData'],
+        display: CodeDisplay.fromJson(rawJson['display']),
+      );
+      return resultCode;
+    } catch (e) {
+      debugPrint("Failed to parse code from export json $e");
```
ua741: nit: user logger.severe instead of debugPrint, also let's log stacktrace?

auth/lib/ui/settings/data/import/plain_text_import.dart
```diff
@@ -101,20 +101,35 @@ Future<void> _pickImportFile(BuildContext context) async {
...
-    for (final code in splitCodes) {
-      try {
-        parsedCodes.add(Code.fromRawData(code));
-      } catch (e) {
-        Logger('PlainText').severe("Could not parse code", e);
+
+    if (codes.startsWith('otpauth://')) {
+      List<String> splitCodes = codes.split(",");
+      if (splitCodes.length == 1) {
+        splitCodes = codes.split("\n");
```
ua741: why remove `LineSplitter` IIRC, it was added to fix some customer issue.

auth/lib/ui/scanner_page.dart
```diff
@@ -66,11 +67,12 @@ class ScannerPageState extends State<ScannerPage> {
     }
     controller.scannedDataStream.listen((scanData) {
       try {
-        final code = Code.fromRawData(scanData.code!);
+        final code = Code.fromOTPAuthUrl(scanData.code!);
         controller.dispose();
         Navigator.of(context).pop(code);
       } catch (e) {
         // Log
+        showToast(context, context.l10n.invalidQRCode);
```
ua741, line 75: I think not showing a toast here was by choice. @vishnukvmd might know better as he had added this part of code.

auth/lib/onboarding/view/common/add_chip.dart
```diff
@@ -0,0 +1,27 @@
...
+    return GestureDetector(
+      onTap: onTap,
+      child: Padding(
+        padding: const EdgeInsets.symmetric(vertical: 8.0),
+        child: Icon(
+          Icons.add_circle_outline,
+          size: 30,
+          color: Theme.of(context).brightness == Brightness.dark
+              ? const Color(0xFF9610D6)
+              : const Color(0xFF8232E1),
```
ua741: These colors should ideally be defined in `EnteColorScheme` and we should use `getEnteColorScheme(context)` to get the color based on theme

auth/lib/store/code_store.dart
```diff
@@ -22,28 +23,51 @@ class CodeStore {
...
-      return compareAsciiLowerCaseNatural(a.account, b.account);
+      return compareAsciiLowerCaseNatural(
+        firstCode.account,
+        secondCode.account,
+      );
     });
-    return codes;
+    return AllCodes(
+      codes: codes,
+      state: hasError ? AllCodesState.error : AllCodesState.value,
```
ua741: This error state should be part of Code? Otherwise if any code parsing fails, we will end up showing error state for all valid codes too.
```dart
Code {
 Error? err;
 bool get hasError =? err != nil;
  Code withError(Error err) {
  }
```
}

auth/lib/ui/scanner_page.dart
prateekmedia: Actually it was done in previous pinning PR, so I thought it should be the same.

auth/lib/ui/settings/data/import/plain_text_import.dart
prateekmedia: Yes, I rebased some code from previous pinning PR, so this might be from it.

ua741: Okay, that PR might not have latest main. Should revert this change

auth/lib/store/code_store.dart
prateekmedia: I have kept both, store error and also store state of allcodes to keep it handy while checking if any error was there while fetching codes.

auth/lib/ui/settings/data/export_widget.dart
```diff
@@ -171,10 +171,11 @@ Future<void> _exportCodes(BuildContext context, String fileContent) async {
 }
 
 Future<String> _getAuthDataForExport() async {
-  final codes = await CodeStore.instance.getAllCodes();
+  final allCodes = await CodeStore.instance.getAllCodes();
   String data = "";
-  for (final code in codes) {
-    data += "${code.rawData}\n";
+  for (final code in allCodes.codes) {
+    data += "${code.rawDataWithoutDisplay}\n";
```
ua741: Should we export with display? User might expect their tags to be available.

auth/lib/ui/settings/data/export_widget.dart
```diff
@@ -171,10 +171,12 @@ Future<void> _exportCodes(BuildContext context, String fileContent) async {
...
 
 Future<String> _getAuthDataForExport() async {
-  final codes = await CodeStore.instance.getAllCodes();
+  final allCodes = await CodeStore.instance.getAllCodes();
   String data = "";
-  for (final code in codes) {
-    data += "${code.rawData}\n";
+  for (final code in allCodes) {
+    if (code.hasError) continue;
+    data += "${code.rawData.replaceAll(',', '%2C')}\n";
```
ua741, line 178: We are not exporting code display?

ua741 (approved):

PR #1561, opened 2024-04-30
[web] Revert "search-bar component added"

mnvr:
This reverts commit 4ed194f397fd4fdf0daf0238f93b08171f3528e5.

@daviddeepan I had to revert your PR unfortunately.

Two main functional issues was:

- When trying to upload files, if I would search something without results, the dialog would just close instantly.

- As a user, I expect to be able to move out of the search bar with escape. However, pressing escape closes the whole dialog, causing the user's drag and dropped or selected files to be lost. This is not specifically a problem with the code added in this commit, however this UX will need to be fixed in some way before introducing the search bar.

One medium issue:

- The dialog box kept changing size when I searched. This is visually jarring.

Other minor issues:

- The input element did not match the look and feel for the rest of the search input elements in the app. Aligning the search bar to the left and right of the album thumbnails and using the MUI Input element (as our code uses MUI  elements in other places) might've fixed it.

- There were various debugging console.logs left over, once of which even seemed like a case we need to handle.

- The strings were not localized.

---

Again, I appreciate your effort, but I think let us leave this for now, we can come back to it in the future after having thought more about the UX.

### DIFF
--- a/web/apps/photos/src/components/Collections/CollectionSelector/index.tsx
+++ b/web/apps/photos/src/components/Collections/CollectionSelector/index.tsx
@@ -39,7 +39,6 @@ interface Props {
     collections: Collection[];
     collectionSummaries: CollectionSummaries;
 }
-
 function CollectionSelector({
     attributes,
     collectionSummaries,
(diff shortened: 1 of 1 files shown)


daviddeepan:
Of course, @mnvr thank you for your review! 
I'll keep in mind to keep an eye out for such issues.

PR #1611, opened 2024-05-04
[web] Rework cast

mnvr:
* HEIC conversion is still pending, but that should be easy to add if the rest everything is working fine on Chromecast
* Have tested on manual pairing


web/apps/cast/src/services/pair.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    context.addCustomMessageListener(
+        namespace,
+        // We need to cast, the `senderId` is present in the message we get but
+        // not present in the TypeScript type.
+        incomingMessageListener as unknown as SystemEventHandler,
+    );
+
+    // Shutdown ourselves if the sender disconnects.
+    // TODO(MR): I assume the page reloads on shutdown. Is that correct?
```
mnvr, line 151: @ua741 I've left some TODOs in this file to add a few more comments around various customizations we're doing, if you know the reasons let me know.

ua741 (approved):

PR #1706, opened 2024-05-13
[desktop] --no-sandbox does not prevent the SUID abort when added from JS

mnvr:
- Revert the workaround
- Document the behaviour I observed
- (Unrelated) Add the other CHANGELOG entries in preparation for the release


mnvr:
For reference, a possible workaround for this scenario would be to do something like this - https://github.com/gergof/electron-builder-sandbox-fix/blob/master/lib/index.js - but unconditionally on Linux.

PR #1726, opened 2024-05-14
Make embedding bucket configurable

ua741:
## Tests
- [x] New ml data is doing to new bucket
- [x] For existing embedding, fall back logic is working as expected, we are returning the object immediately and copying the object to new bucket in an async manner. 
- [x] Verified that the dc values were getting updated correctly on copy or insert.
- [x] Verified that on deletion, we are deleting files from all dcs where the derived file is present.


server/pkg/controller/embedding/controller.go
```diff
@@ -368,32 +395,99 @@ func (c *Controller) getEmbeddingObjectsParallelV2(userID int64, dbEmbeddingRows
...
+			}
+			// Check if the error is due to context timeout or cancellation
+			if err == nil && fetchCtx.Err() != nil {
+				ctxLogger.Error("Fetch timed out or cancelled: ", fetchCtx.Err())
+			} else {
+				// check if the error is due to object not found
+				if s3Err, ok := err.(awserr.RequestFailure); ok {
+					if s3Err.Code() == s3.ErrCodeNoSuchKey {
+						if c.areEmbeddingAndHotBucketSame {
+							ctxLogger.Error("Object not found: ", s3Err)
```
mnvr: We should abort instead of retrying when this happens (Not fully aware of how the go control flow is working here, so please ignore if this is already what the code is doing)

mnvr (approved):

server/configurations/local.yaml
```diff
@@ -125,6 +125,16 @@ s3:
         endpoint:
         region:
         bucket:
+    wasabi-eu-central-2-embeddings:
```
mnvr: Changed this to `wasabi-eu-central-2-derived`. We should probably change the rest of the comments and `embeddings-bucket` config var too.

server/configurations/local.yaml
```diff
@@ -125,6 +125,16 @@ s3:
...
+        key:
+        secret:
+        endpoint:
+        region:
+        bucket:
+    # Embeddings bucket is used for storing embeddings and other derived data from a file.
+    # By default, it is the same as the hot storage bucket.
+    # embeddings-bucket: wasabi-eu-central-2-embeddings
```
mnvr: `derived-storage: wasabi-eu-central-2-derived`

mnvr (approved):

PR #1767, opened 2024-05-18
[meta] Prune gitignore

mnvr:
About the auth dbs, discussed in chat: These files got created once when debugging auth build for linux. They were probably accidental, so removing them for now (will add back if there is a workflow when they actively get recreated).


prateekmedia (approved):
LGTM!

PR #1809, opened 2024-05-22
Ente Auth: Add support for Steam domain (steampowered.com) as an issuer

pashynskykh:
## Description
I don't think I'm the only one using website domains as issuers of TOTP codes. This change will add support for the Steam domain (steampowered.com) as an issuer.

### DIFF
--- a/auth/lib/models/code.dart
+++ b/auth/lib/models/code.dart
@@ -128,7 +128,7 @@ class Code {
       final code = Code(
         _getAccount(uri),
         issuer,
-        _getDigits(uri, issuer),
+        _getDigits(uri),
         _getPeriod(uri),
         getSanitizedSecret(uri.queryParameters['secret']!),
         _getAlgorithm(uri),
@@ -201,11 +201,11 @@ class Code {
     }
   }
 
-  static int _getDigits(Uri uri, String issuer) {
+  static int _getDigits(Uri uri) {
     try {
       return int.parse(uri.queryParameters['digits']!);
     } catch (e) {
-      if (issuer.toLowerCase() == "steam" || issuer.toLowerCase().contains('steampowered.com')) {
+      if (uri.host == "steam") {
         return steamDigits;
       }
       return defaultDigits;
--- a/auth/lib/models/code.dart
+++ b/auth/lib/models/code.dart
@@ -205,7 +205,7 @@ class Code {
     try {
       return int.parse(uri.queryParameters['digits']!);
     } catch (e) {
-      if (issuer.toLowerCase() == "steam") {
+      if (issuer.toLowerCase() == "steam" || issuer.toLowerCase().contains('steampowered.com')) {
         return steamDigits;
       }
       return defaultDigits;
--- a/auth/lib/utils/totp_util.dart
+++ b/auth/lib/utils/totp_util.dart
@@ -4,7 +4,7 @@ import 'package:otp/otp.dart' as otp;
 import 'package:steam_totp/steam_totp.dart';
 
 String getOTP(Code code) {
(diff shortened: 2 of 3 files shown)


auth/lib/utils/totp_util.dart
```diff
@@ -4,7 +4,7 @@ import 'package:otp/otp.dart' as otp;
 import 'package:steam_totp/steam_totp.dart';
 
 String getOTP(Code code) {
-  if (code.issuer.toLowerCase() == 'steam') {
```
ua741, line 7: Hello,
Can you please change this to use `code.type == Type.steam`

And modify following line to handle domain as issuer

https://github.com/ente-io/ente/blob/main/auth/lib/onboarding/view/setup_enter_secret_key_page.dart#L243

```dart
final isStreamCode =
          issuer.toLowerCase() == "steam" || issuer.toLowerCase().contains('steampowered.com');
```

Thanks for the PR :)

prateekmedia, line 7: Other than what @ua741 suggested, you will also have to modify the logic of _getType and _getDigits in code.dart to support both the issuers.

pashynskykh: Thank you for checking my PR! I will change both IFs with `code.type == Type.steam` in [auth/lib/utils/totp_util.dart](https://github.com/ente-io/ente/pull/1809/files/954581093dd751fecf3d823985a8dac00ce74dfc#diff-a2a9b59a01e1ebc767cb4c397fcd346b91686783c9dd0a0cc5aceced09e553b0), update [auth/lib/onboarding/view/setup_enter_secret_key_page.dart#L243](https://github.com/ente-io/ente/blob/main/auth/lib/onboarding/view/setup_enter_secret_key_page.dart#L243), add changes suggested below by @prateekmedia and push a new commit.

pashynskykh: Thanks for the help! I updated _getDigits in code.dart. But I'm not quite sure what to do with _getType. It looks like _getType deals with an algorithm, but not an issuer.

For example, in `otpauth://steam/steampowered.com:MyUserName?secret=XXXXXXXXXXXXXXXXXX` "steam" is an algorithm and after that "steampowered.com" is used as an issuer. 

Please correct me if I wrong.

Thanks once again!

prateekmedia: Ok, let's proceed with that!

pashynskykh:
@ua741 @prateekmedia please review the new commit. I changed everything according to your recommendations.

auth/lib/models/code.dart
```diff
@@ -205,7 +205,7 @@ class Code {
     try {
       return int.parse(uri.queryParameters['digits']!);
     } catch (e) {
-      if (issuer.toLowerCase() == "steam") {
+      if (issuer.toLowerCase() == "steam" || issuer.toLowerCase().contains('steampowered.com')) {
```
prateekmedia: Can we also check host here if it is steam like _getType instead of checking individual issuer? Also remove the issuer parameter if not used elsewhere in this function.

prateekmedia (approved):
LGTM!

auth/lib/models/code.dart
pashynskykh: Done! Please check the new commit. Updated _getDigits to use uri.host instead of issuer and removed the issuer parameter from the function. This should be checked though. Unfortunately, I don't have time to go deeper into the code.

pashynskykh:
@prateekmedia, awesome! Thank you for your help with my PR! 

This is a small change, but I'm happy to make Ente Auth better. It is the only TOTP generator that is both FOSS and cross-platform.

PR #1945, opened 2024-05-31
[desktop] Show option to enable face indexing for beta users

mnvr:


I-I-IT:
Beta only for internal or external users can apply too ?

laurenspriem:
@I-I-IT the beta is for a limited number of external users, but lucky for you you're just in time.

Just send an email to team@ente.io with the email of your Ente account and we'll add you.

Issue #1909, opened 2024-05-28
Photos Desktop App Icon Not Showing on Linux


Jishnuraj9:
### Description

Despite successfully installing the app using AppImage , the app does not appear in the Linux "Finder" with an icon.

### Version

v1.7.0

### What product are you using?

Ente Photos

### What platform are you using?

Desktop - Linux

mnvr:
There are two different issues with icons on Linux.

* Firstly, the AppImage itself doesn't show an icon. There does not seem to be a reasonable workaround either currently. See: https://github.com/AppImage/AppImageKit/issues/346

* Secondly, and this is the problem that should be fixable, when the app is started it does not show a dock icon (Ubuntu 22) or shows the generic gear icon (Ubuntu 24). The issue possibly exists on other distributions too.

We've already added the call to `BrowserWindow.setIcon`, but that hasn't helped. If someone knows what we should do differently to solve the issue, please leave a comment.

residualbit:
This is an issue with the .rpm file as well, also introduced in 1.7.

It appears as though something is breaking when it creates the icon files in /usr/share/icons. As of this release, it dumps them into a new /usr/share/icons/0x0/apps folder rather than ones of the appropriate size and that folder is not referenced by the theme or fallbacks.

mnvr:
This won't be in the logs, this would be during the installation. Two possibilities are: either something changed in electron-builder, the library we use for generating RPMs etc (1.7 has an updated version), or something changed in how we're laying out the icons. 

I'll see if I can spot the difference. You've been a great help already, thank you @residualbit.

mnvr:
I've modified how we're laying out icons to rule out the second possibility, reverted this code back to exactly how it was in 1.6.63. @residualbit If you get a chance and want to test if it fixes the issue you can install the latest nightly build from here - https://github.com/ente-io/photos-desktop/releases. Otherwise I'll try to spin up a VM later and test the RPM.

residualbit:
@mnvr, can confirm, that worked! Icon behavior and directories are back to normal.

mnvr:
Tested on Ubuntu also, the app shows the correct dock icon when it runs. So I will close this, the fix will be out in the next version, 1.7.1.

There is a separate issue - the AppImage itself not showing an icon. As described [above](https://github.com/ente-io/ente/issues/1909#issuecomment-2144329330), there isn't a reasonable fix for it, it seems like a regression in AppImages themselves.

PR #1959, opened 2024-06-01
[desktop] (Maybe) fix the dock icon on Linux

mnvr:
Ref: https://github.com/ente-io/ente/issues/1909

build/icons/512-512.png is just a 1/2 copy of build/icons.png.


mnvr:
It didn't do anything in Ubuntu 22 either. Leaving the code in place for now to revisit later.

PR #2022, opened 2024-06-05
Created instructions to import from most providers

TomsDex:
## Description
Added detail to docs on what's able to be imported via the app

Give me a shout if anything needs updating!


mnvr (approved):
Thank you!

I'll also add this to the sidebar in a separate PR. Thanks again.

PR #2048, opened 2024-06-07
[docs] Add sharable URL configuration

xqus:
This PR adds a configuration option needed if you want sharable urls to work


mnvr (approved):
Thank you!

PR #2073, opened 2024-06-10
[mob][photos] Modify the code to use XFile.fromData(bytes) instead of saving the image to a file for share link placeholder

AmanRajSinghMourya:
1. Modified the` _createPlaceholder` function to return the image bytes instead of the image path, changed the return type to `Future<Uint8List?>`
2. Remove the call to `saveImage` on saving the image to a temporary directory
3. Modified the `shareImageAndUrl` function:
4. Replaced the usage of XFile(imagePath) with XFile.fromData(imageBytes).
5. Added metadata to the `XFile.fromData` method call, specifying the name and mimeType for better file identification during sharing.

### DIFF
--- a/mobile/lib/ui/viewer/actions/file_selection_actions_widget.dart
+++ b/mobile/lib/ui/viewer/actions/file_selection_actions_widget.dart
@@ -1,5 +1,4 @@
 import "dart:async";
-import "dart:io";
 
 import 'package:fast_base58/fast_base58.dart';
 import "package:flutter/cupertino.dart";
@@ -7,7 +6,6 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import "package:logging/logging.dart";
 import "package:modal_bottom_sheet/modal_bottom_sheet.dart";
-import "package:path_provider/path_provider.dart";
 import 'package:photos/core/configuration.dart';
 import "package:photos/core/event_bus.dart";
 import "package:photos/events/people_changed_event.dart";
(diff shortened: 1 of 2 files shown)


mobile/lib/ui/viewer/actions/file_selection_actions_widget.dart
```diff
@@ -610,40 +608,21 @@ class _FileSelectionActionsWidgetState
...
-      final file = await File(filePath).writeAsBytes(bytes);
-      path = file.path;
-    } catch (e) {
-      _logger.severe("Failed to save placeholder image", e);
-    }
-    return path;
-  }
-
-  Future<String?> _createPlaceholder(
+  Future<Uint8List?> _createPlaceholder(
```
ashilkn: `bytesOfImageToWidget` is of a non-nullable type. Can make the return type here a non-nullable future (`Uint8List`)

mobile/lib/ui/viewer/actions/file_selection_actions_widget.dart
```diff
@@ -610,40 +608,21 @@ class _FileSelectionActionsWidgetState
...
     List<EnteFile> ownedSelectedFiles,
   ) async {
     final Widget imageWidget = LinkPlaceholder(
       files: ownedSelectedFiles,
     );
-    await Future.delayed(const Duration(milliseconds: 100));
+    await Future.delayed(const Duration(milliseconds: 200));
     final double pixelRatio = MediaQuery.of(context).devicePixelRatio;
```
ashilkn: Change to ` MediaQuery.devicePixelRatioOf(context);`.

When ever using `MediaQuery.of(context)`, see if there an alternate static method that can be called from  https://api.flutter.dev/flutter/widgets/MediaQuery-class.html

ashilkn (approved):

PR #2111, opened 2024-06-12
[server] Reduce passkey JWT duration + API to get token via sessionID

ua741:
## Tests
Monkey tested locally


server/pkg/controller/user/jwt.go
```diff
@@ -13,11 +13,15 @@ import (
 const ValidForDays = 1
 
 func (c *UserController) GetJWTToken(userID int64, scope enteJWT.ClaimScope) (string, error) {
+	tokenExpirty := time.NDaysFromNow(1)
```
mnvr: typo expiry

mnvr (approved):

PR #2130, opened 2024-06-14
[mob][photos] Implemeted Applock to set custom pin/password

AmanRajSinghMourya:
This PR add the following features
1. **Device Lock:** If the user's device has a screen lock (PIN, fingerprint, etc.), it will be used for initial authentication.
2. **Custom PIN/Passwprd**: Users can set a custom PIN/Password for app access, independent of the device lock.

<br/>

<br/>


brog-io:
Would it be possible to have this for Auth as well?

ashilkn:
> Would it be possible to have this for Auth as well?

Yes, will be adding it to auth as well :)

mobile/lib/services/local_authentication_service.dart
```diff
@@ -33,6 +37,43 @@ class LocalAuthenticationService {
     return true;
   }
 
+  Future<bool> requestEnteAuthForLockScreen(BuildContext context) async {
+    final String? savedPin = await _configuration.getPin();
+    final String? savedPassword = await _configuration.getPassword();
```
ashilkn: Can avoid redundant `getPin()` and `getPassword()` and instead pass pin and password from where this is called since we do read pin and password before calling this.

mobile/lib/ui/components/text_input_widget.dart
```diff
@@ -43,6 +46,7 @@ class TextInputWidget extends StatefulWidget {
   final ValueNotifier? isEmptyNotifier;
   final List<TextInputFormatter>? textInputFormatter;
   final TextInputType? textInputType;
+  final bool? fillColor;
```
ashilkn: Can make this non-nullable. It's always either true (default) or false. 
Slightly feel it's name can be changed to `enableFillColor`.

mobile/lib/ui/settings/TEMP/lock_screen_option.dart
```diff
@@ -0,0 +1,230 @@
...
+  bool isPasswordEnabled = false;
+
+  @override
+  void initState() {
+    isPasswordEnabled = _configuration.isPasswordSet();
+    isPinEnabled = _configuration.isPinSet();
+    appLock = isPinEnabled ||
+        isPasswordEnabled ||
+        _configuration.shouldShowLockScreen();
+    super.initState();
```
ashilkn: Should move this up to the first line of `initState()`.

mobile/lib/ui/settings/TEMP/lock_screen_option.dart
```diff
@@ -0,0 +1,230 @@
...
+class LockScreenOption extends StatefulWidget {
+  const LockScreenOption({super.key});
+
+  @override
+  State<LockScreenOption> createState() => _LockScreenOptionState();
+}
+
+class _LockScreenOptionState extends State<LockScreenOption> {
+  final Configuration _configuration = Configuration.instance;
+  bool? appLock;
```
ashilkn: Can make this non-nullable. Either use late keyword or assign an initial boolean value.

mobile/lib/ui/settings/TEMP/lock_screen_option.dart
```diff
@@ -0,0 +1,230 @@
...
+    });
+  }
+
+  Future<void> _onToggleSwitch() async {
+    if (appLock == false && !(isPasswordEnabled || isPinEnabled)) {
+      AppLock.of(context)!.setEnabled(!appLock!);
+      await Configuration.instance.setShouldShowLockScreen(!appLock!);
+    } else {
+      AppLock.of(context)!.setEnabled(!appLock!);
+      await Configuration.instance.setShouldShowLockScreen(!appLock!);
```
ashilkn: Both if and else block is doing the same thing?

mobile/lib/ui/settings/TEMP/lock_screen_option_confirm_password.dart
```diff
@@ -0,0 +1,166 @@
...
+  final String password;
+
+  @override
+  State<LockScreenOptionConfirmPassword> createState() =>
+      _LockScreenOptionConfirmPasswordState();
+}
+
+class _LockScreenOptionConfirmPasswordState
+    extends State<LockScreenOptionConfirmPassword> {
+  final _confirmPasswordController = TextEditingController(text: null);
```
ashilkn: Make sure this is disposed.

mobile/lib/ui/settings/TEMP/lock_screen_option_confirm_password.dart
```diff
@@ -0,0 +1,166 @@
...
+  final Configuration _configuration = Configuration.instance;
+  final _focusNode = FocusNode();
+  final _isFormValid = ValueNotifier<bool>(false);
```
ashilkn: Make sure this valueNotifier is disposed.

mobile/lib/ui/settings/TEMP/lock_screen_option_confirm_password.dart
```diff
@@ -0,0 +1,166 @@
...
+    }
+    await HapticFeedback.vibrate();
+    throw Exception("Incorrect password");
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final colorTheme = getEnteColorScheme(context);
+    final textTheme = getEnteTextTheme(context);
+    final isKeypadOpen = MediaQuery.of(context).viewInsets.bottom > 100;
```
ashilkn: Use `MediaQuery.viewInsetsOf(context)`

mobile/lib/ui/settings/TEMP/lock_screen_option_confirm_pin.dart
```diff
@@ -0,0 +1,386 @@
...
+    width: 48,
+    decoration: BoxDecoration(
+      border: Border.all(color: const Color.fromRGBO(45, 194, 98, 1.0)),
+      borderRadius: BorderRadius.circular(15.0),
+    ),
+  );
+  @override
+  void initState() {
+    super.initState();
+  }
```
ashilkn: Can remove this.

mobile/lib/ui/settings/TEMP/lock_screen_option_confirm_pin.dart
```diff
@@ -0,0 +1,386 @@
...
+  @override
+  State<LockScreenOptionConfirmPin> createState() =>
+      _LockScreenOptionConfirmPinState();
+}
+
+class _LockScreenOptionConfirmPinState
+    extends State<LockScreenOptionConfirmPin> {
+  final _confirmPinController = TextEditingController(text: null);
+  final Configuration _configuration = Configuration.instance;
+  Key _pinputKey = UniqueKey();
```
ashilkn: What is the use of this key? I haven't checked, just curious. Should be removed if it's unnecessary.

mobile/lib/ui/settings/TEMP/lock_screen_option_confirm_pin.dart
```diff
@@ -0,0 +1,386 @@
...
+    super.initState();
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
+    _confirmPinController.dispose();
+  }
+
+  String _pin = "";
```
ashilkn: Do we need this variable? I haven't checked all uses of `_pin`, but wouldn't just using `_confirmPinController.text` instead be enough?

mobile/lib/ui/settings/TEMP/lock_screen_option_password.dart
```diff
@@ -0,0 +1,179 @@
...
+  });
+  final bool isAuthenticating;
+  final String? authPass;
+  @override
+  State<LockScreenOptionPassword> createState() =>
+      _LockScreenOptionPasswordState();
+}
+
+class _LockScreenOptionPasswordState extends State<LockScreenOptionPassword> {
+  final _passwordController = TextEditingController(text: null);
```
ashilkn: Make sure the controller is disposed.

mobile/lib/ui/settings/TEMP/lock_screen_option_password.dart
```diff
@@ -0,0 +1,179 @@
...
+      );
+      _passwordController.clear();
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final colorTheme = getEnteColorScheme(context);
+    final textTheme = getEnteTextTheme(context);
+    final isKeypadOpen = MediaQuery.of(context).viewInsets.bottom > 100;
```
ashilkn: Use `MediaQuery.viewInsetsOf(context)`

mobile/lib/ui/settings/TEMP/lock_screen_option.dart
AmanRajSinghMourya: `if (appLock == false && !(isPasswordEnabled || isPinEnabled))` 
   
then `Configuration.instance.setShouldShowLockScreen(!appLock)` is set to `true` by doing `!appLockk`
else
`Configuration.instance.setShouldShowLockScreen(!appLock)` is set to `false` by doing `!appLockk`

mobile/lib/ui/settings/TEMP/lock_screen_option_confirm_password.dart
AmanRajSinghMourya: When i try to dispose this, i get an error for `TextEditingController was used after being disposed.` 
I checked other `TextInputWidget` where the `textEditingController` is used, found that the controller is not disposed there also. We also discussed about this error before

ashilkn: Okay. Just mention why this isn't disposed in a comment over it, just for context in future.

mobile/lib/ui/settings/TEMP/lock_screen_option.dart
ashilkn: Either the `if` or `else` block will execute. If both blocks are doing the same thing, let's remove the unnecessary `if` and `else` conditions? 

```
if (appLock == false && !(isPasswordEnabled || isPinEnabled)) {
      AppLock.of(context)!.setEnabled(!appLock);
      await Configuration.instance.setShouldShowLockScreen(!appLock);
    } else {
      AppLock.of(context)!.setEnabled(!appLock);
      await Configuration.instance.setShouldShowLockScreen(!appLock);
    }
```

and just

```
  AppLock.of(context)!.setEnabled(!appLock);
  await Configuration.instance.setShouldShowLockScreen(!appLock);
```

does the same thing. So let's remove the unnecessary conditional statements?

mobile/lib/ui/components/text_input_widget.dart
ashilkn: This is still nullable.

mobile/lib/ui/settings/TEMP/lock_screen_option_password.dart
ashilkn: Does disposing this controller also throw an error/exception? If yes, mention it in a comment over it.

AmanRajSinghMourya: Yes this one also throws an exception

ashilkn (approved):

PR #2207, opened 2024-06-18
Dashboard improvement

atyabbin:
After clicking the FETCH button, if the data is fetched successfully then a new button labelled as 'MORE' appears, clicking on this opens a dropdown menu which has 3 options: Disable2fa, Disable Passkeys and Close Family, clicking on any of the option directly performs the action required.


server/configurations/local.yaml
```diff
@@ -303,7 +303,8 @@ internal:
     #     local-domain-suffix: "@example.org"
     #     local-domain-value: 123456
     # List of user IDs that can use the admin API endpoints.
-    admins: []
+    admins:
```
ua741: undo? you can create a `museum.yaml` file in server/ directory and add your local configurations in that.

atyabbin: Done in the new commit

ua741 (approved):

PR #2240, opened 2024-06-21
Warning about icon sizes and instructions on matching the icon to the issuer.

vbenincasa:
## Description

Some contributors are submitting unoptimized/huge PNG/JPG icons encapsulated in SVG, resulting in an unnecessary increase in the size of installation packages, as well as increased memory consumption and bandwidth when users use these icons.

Icon matching instructions are based on function:
https://github.com/ente-io/ente/blob/30dd41e3cae11ed7d1eeccf8e5f99c671e3f863a/auth/lib/ui/utils/icon_utils.dart#L147

### DIFF
--- a/auth/docs/adding-icons.md
+++ b/auth/docs/adding-icons.md
@@ -7,6 +7,15 @@ If you would like to add your own custom icon, please open a pull-request with
 the relevant SVG placed within `assets/custom-icons/icons` and add the
 corresponding entry within `assets/custom-icons/_data/custom-icons.json`.
 
+Please be careful to upload small and optimized icon files. If your icon file 
+is over 50KB, it is certainly not optimized.
+
+Note that the correspondence between the icon and the issuer is based on the name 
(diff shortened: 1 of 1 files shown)


auth/docs/adding-icons.md
```diff
@@ -7,6 +7,15 @@ If you would like to add your own custom icon, please open a pull-request with
 the relevant SVG placed within `assets/custom-icons/icons` and add the
 corresponding entry within `assets/custom-icons/_data/custom-icons.json`.
 
+Please be careful to upload small and optimized icon files. If your icon file 
+is over 50KB, it is certainly not optimized.
```
mnvr: Can we change "certainly" to "likely" so that it comes off as a bit more friendlier? Otherwise LGTM!

vbenincasa: Definitely a good observation, please go ahead.

ua741 (approved):

PR #2254, opened 2024-06-22
[server] Allow replication without Cloudflare worker

nicegamer7:
## Description
I recently started self hosting an instance of Ente. All went well until I tried to enable replication.

This PR adds an option that allows replication to be enabled without a Cloudflare worker, useful for self hosting.

## Tests
It doesn't look like there are any automated tests that I can add to, but let me know if I'm wrong. Otherwise, I tested this on my own instance and it's working as expected.

### DIFF
--- a/server/pkg/controller/replication3.go
+++ b/server/pkg/controller/replication3.go
@@ -87,10 +87,11 @@ func (c *ReplicationController3) StartReplication() error {
 
 	workerURL := viper.GetString("replication.worker-url")
 	if workerURL == "" {
-		return fmt.Errorf("replication.worker-url was not defined")
+		log.Infof("replication.worker-url was not defined, files will downloaded directly during replication")
+	} else {
+		log.Infof("Worker URL to download objects for replication v3 is: %s", workerURL)
 	}
 	c.workerURL = workerURL
-	log.Infof("Worker URL to download objects for replication v3 is: %s", workerURL)
 
 	c.createMetrics()
 	err := c.createTemporaryStorage()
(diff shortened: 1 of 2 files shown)


mnvr (approved):
Thank you!

server/pkg/controller/replication3.go
```diff
@@ -85,12 +85,19 @@ func (c *ReplicationController3) StartReplication() error {
 		return fmt.Errorf("v3 replication can currently only run when the primary hot data center is Backblaze. Instead, it was %s", hotDC)
 	}
 
+	// If bypassWorker is true, we set c.workerURL to "", which makes downloadFromB2ViaWorker
+	// bypass the download worker. c.workerURL should not be empty otherwise
+	bypassWorker := viper.GetBool("replication.bypass-worker")
```
mnvr: On reading the existing code again, I'm wondering if we should just make the worker-url optional. @nicegamer7 what are your thoughts on this?  There is extra cognitive and complexity cost of adding a new option, so if we just instead make the worker-url optional and use its absence as a signal for direct replication.

nicegamer7: I agree that adding a new option adds unneeded complexity, especially since it seems there aren't any automated tests to ensure this will stay working as intended. For that reason I think it'd be better to just make worker-url optional.

The only reason I didn't do that in the first place was to prevent changing the server's existing behaviour, since I wasn't sure if that'd be welcome.

I'll get this changed.

mnvr: > The only reason I didn't do that in the first place was to prevent changing the server's existing behaviour, since I wasn't sure if that'd be welcome.

Yes, I understood you'd intentionally tried to minimize the diff. Thank you!

mnvr (approved):

PR #2286, opened 2024-06-25
[mob][photos] Fix file bottom bar not disappearing when in fullscreen

ashilkn:
## Description

When viewing photos/videos and enabling fullscreen (with a tap on scree), the bottom bar wasn't disappearing like it used to. This PR fixes this regression.


ua741 (approved):

ashilkn:
@prateekmedia This regression is from https://github.com/ente-io/ente/pull/1748
Was there any specific reason why `IgnorePointer` and `AnimatedOpacity` was removed, was this causing any issues with the video editor? I've added it back and couldn't spot any issues with the video editor.

PR #2367, opened 2024-07-05
[mob][photos] Lockscreen Fixes

AmanRajSinghMourya:


mobile/lib/ui/tools/lock_screen.dart
```diff
@@ -203,36 +195,20 @@ class _LockScreenState extends State<LockScreen>
...
-      ],
-    );
-
-    await showDialog(
-      context: context,
-      builder: (BuildContext context) {
-        return alert;
-      },
+    _logger.info("Auto logout on max invalid attempts");
+    await _lockscreenSetting.setInvalidAttemptCount(0);
```
ashilkn: I think this can be removed, since we clear `sharedPreferences` on logout here
https://github.com/ente-io/ente/blob/543fb835f34e96bc825b9955ac8870c8d98ff6d5/mobile/lib/core/configuration.dart#L202

ashilkn (approved):

Issue #1328, opened 2024-04-05
When trying to run ente cli from a server it fails on dbus-launch


I-Would-Like-To-Report-A-Bug-Please:
### Description

Tried to run ente cli on my docker server however I can't due to missing dbus-launch.

``` bash
root@docker:~/docker/ente# ./ente
2024/04/05 10:52:54 error getting password from keyring: exec: "dbus-launch": executable file not found in $PATH
``` 

### Version

cli-v0.1.13

### What platform are you using?

Server

AlexandarY:
Issue appears to be coming from the use of https://github.com/zalando/go-keyring . This module requires the secret service dbus interface. To get it to work on my end (in a Debian Bookworm container), I had to do the following:

1. Install these packages - `apt install gnome-keyring libsecret dbus-x11`
2. Create the default collection. Setting this up without an X11 Server is a bit tricky, but the steps described [here](https://unix.stackexchange.com/questions/473528/how-do-you-enable-the-secret-tool-command-backed-by-gnome-keyring-libsecret-an/548005#548005) do work: 

```
eval "$(dbus-launch --sh-syntax)"

mkdir -p ~/.cache
mkdir -p ~/.local/share/keyrings # where the automatic keyring is created

# 1. Create the keyring manually with a dummy password in stdin
eval "$(printf '\n' | gnome-keyring-daemon --unlock)"

# 2. Start the daemon, using the password to unlock the just-created keyring:
eval "$(printf '\n' | /usr/bin/gnome-keyring-daemon --start)"
```

Also if you are trying to use the CLI to create an account, make sure that the yaml config file has the `hardcoded-ott.emails` value set for the user you are trying to use, otherwise you won't be able to use the CLI or even log in.

diva-in-STEM:
Hi there, I'm trying to get this working but I'm running into this error:
` error getting password from keyring: failed to unlock correct collection '/org/freedesktop/secrets/aliases/default'`
I'm trying to run it on a raspberry pi 5 using the 64 bit raspbian lite OS (I'm trying to run the arm64 version BTW)

raphaelm:
Installing these packages is not an option on all systems, e.g. I try to download photos to my Synology NAS. Any other workarounds?

diva-in-STEM:
Hi, sorry, I have fixed it! Turns out I had the IP address wrong for the s3 buckets and running the CLI on my windows machine worked wonderfully!

raphaelm:
[...]

Issue #722, opened 2024-01-26
error getting password from keyring


31ank:
I'm running a headless Ubuntu 22.04.3 server and tried installing the deb package from the release tab. When I try to execute a ente command I get the error `error getting password from keyring: failed to unlock correct collection '/org/freedesktop/secrets/collection/login'`.
I installed the gnome-keyring-daemon since that seems to be required but that didn't fix the problem.

I'm out of ideas since you don't find much information online regarding this error, does someone know what the issue could be?

RappyTV:
+1

AlexandarY:
I encountered the error from #1328, but at one point it evolves into the error above. The steps described [here](https://unix.stackexchange.com/questions/473528/how-do-you-enable-the-secret-tool-command-backed-by-gnome-keyring-libsecret-an/548005#548005) helped me fix the error. TL;DR on steps:

```
eval "$(dbus-launch --sh-syntax)"

mkdir -p ~/.cache
mkdir -p ~/.local/share/keyrings # where the automatic keyring is created

# 1. Create the keyring manually with a dummy password in stdin
eval "$(printf '\n' | gnome-keyring-daemon --unlock)"

# 2. Start the daemon, using the password to unlock the just-created keyring:
eval "$(printf '\n' | /usr/bin/gnome-keyring-daemon --start)"
```

RappyTV:
> I encountered the error from #1328, but at one point it evolves into the error above. The steps described [here](https://unix.stackexchange.com/questions/473528/how-do-you-enable-the-secret-tool-command-backed-by-gnome-keyring-libsecret-an/548005#548005) helped me fix the error. TL;DR on steps:
> 
> 
> 
> ```
> 
> eval "$(dbus-launch --sh-syntax)"
> 
> 
> 
> mkdir -p ~/.cache
> 
> mkdir -p ~/.local/share/keyrings # where the automatic keyring is created
> 
> 
> 
> # 1. Create the keyring manually with a dummy password in stdin
> 
> eval "$(printf '\n' | gnome-keyring-daemon --unlock)"
> 
> 
> 
> # 2. Start the daemon, using the password to unlock the just-created keyring:
> 
> eval "$(printf '\n' | /usr/bin/gnome-keyring-daemon --start)"
> 
> ```

does not work for me unfortunately

RxnNode:
The workaround doesn't work for me either.

pinpox:
[...]

PR #2426, opened 2024-07-11
CLI: Allow setting secrets path via env var

pinpox:
## Description

Adds a new enviroment variable `ENTE_CLI_SECRETS_PATH`, that allows specifying the path to a `secrets.txt` like the one used in the dockerized version. If no path is provided, the old behaviour is maintained.

The configuration variable and behaviour is analog to the already present `ENTE_CONFIG_PATH` variable, which allows specifying a configuration. 

This provides a solution for people running without gnome-keyring (https://github.com/ente-io/ente/issues/722 and https://github.com/ente-io/ente/issues/1328)

### DIFF
--- a/cli/pkg/secrets/secret.go
+++ b/cli/pkg/secrets/secret.go
@@ -26,10 +26,15 @@ const (
 func GetOrCreateClISecret() []byte {
 	// get password
 	secret, err := keyring.Get(secretService, secretUser)
+
 	if err != nil {
 		if !errors.Is(err, keyring.ErrNotFound) {
+
+			if secretsFile := os.Getenv("ENTE_CLI_SECRETS_PATH"); secretsFile != "" {
+				return GetSecretFromSecretText(secretsFile)
+			}
 			if IsRunningInContainer() {
-				return GetSecretFromSecretText()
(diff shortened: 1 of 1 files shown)


cli/pkg/secrets/secret.go
```diff
@@ -26,10 +26,15 @@ const (
 func GetOrCreateClISecret() []byte {
 	// get password
 	secret, err := keyring.Get(secretService, secretUser)
+
 	if err != nil {
 		if !errors.Is(err, keyring.ErrNotFound) {
+
+			if secretsFile := os.Getenv("ENTE_CLI_SECRETS_PATH"); secretsFile != "" {
+				return GetSecretFromSecretText(secretsFile)
+			}
```
ua741, line 35: Thanks for the change.
Should we put this change inside the else block.
is isRunningContainser() {
} else {
   			if secretsFile := os.Getenv("ENTE_CLI_SECRETS_PATH"); secretsFile != "" {
				return GetSecretFromSecretText(secretsFile)
			}
			log.Fatal(fmt.Errorf("error getting password from keyring: %w", err))
}

pinpox: But then it only works if it is running inside a container. I want to use this env var outside of a container aswell.

ua741 (approved):

PR #2465, opened 2024-07-16
[mob][auth] Lockscreen option auth

AmanRajSinghMourya:
### Pull Request Description

*Features Added:*

1. *App Lock Options:*
- *Device Lock*: Users can set the app lock to utilize the device's default lock mechanism (e.g., fingerprint, face recognition, or device PIN/password).
- *Custom PIN/Password*: Users have the option to set a custom PIN or password for app lock. 

2. *Authentication for App Lock Settings:*
- When users attempt to change the app lock settings, they must authenticate using the currently set PIN/password or, if none is set, the device lock.

3. *App Lock on Background and Quit:*
- The app will prompt for authentication when it is brought back to the foreground after being in the background or when the app is restarted.
4. *Incremental Lock Timer:*
- A timer mechanism is implemented which starts at 30 seconds after few incorrect unlock attempts.
- The lockout time doubles after each incorrect attempt, increasing delay penalties.
- After 10 incorrect attempts, the user is automatically logged out.


auth/lib/utils/lock_screen_settings.dart
```diff
@@ -35,14 +35,16 @@ class LockScreenSettings {
...
     _secureStorage = const FlutterSecureStorage();
     _preferences = await SharedPreferences.getInstance();
-    await shouldHideAppContent(isContentVisible: getShouldHideAppContent());
+
+    ///Workaround for privacyScreen not working when app is killed and opened.
+    await setHideAppContent(getShouldHideAppContent());
   }
 
-  Future<void> shouldHideAppContent({bool isContentVisible = true}) async {
+  Future<void> setHideAppContent(bool showContent) async {
```
ashilkn: Use the same names as on photos to make this part easier to understand

ashilkn (approved):

PR #2532, opened 2024-07-24
[mob][panorama] update panorama checking mechanism

prateekmedia:
## Description

Panorama was getting checked every time and was not getting cached. This PR creates mediaType to classify image whether it's panorama or not.

- If a image can be panorama (Any side Dimensions >8000px and 2 * height >= width or 2 * width >= height) it will try to check metadata if image could be panorama.
- Opening details pane will check metadata if image is panorama or not.
- Gyroscope button is introduced along with proper panoramic view instead of 360* view for Images that have Cropped metadata.


mobile/lib/utils/panorama_util.dart
```diff
@@ -1,28 +1,60 @@
...
+
+// guardedCheckPanorama() method is used to check if the file is a panorama image.
+Future<void> guardedCheckPanorama(EnteFile file) async {
+  if (file.isPanorama() != null) {
+    return;
+  }
+  final result = await checkIfPanorama(file);
+
+  // Update the metadata if it is not updated
+  if (file.isPanorama() == null && file.canEditMetaInfo) {
```
ua741: file.canEditMetaInfo check can happen at the start of the method

mobile/lib/models/file/file.dart
```diff
@@ -179,6 +181,15 @@ class EnteFile {
...
           }
+          mediaUploadData.isPanorama = checkPanoramaFromEXIF(null, exifData);
+
+          try {
+            final xmpData = XMPExtractor()
+                .extract(mediaUploadData.sourceFile!.readAsBytesSync());
+            mediaUploadData.isPanorama = checkPanoramaFromXMP(xmpData);
+          } catch (_) {}
+
+          mediaUploadData.isPanorama ??= false;
```
ua741: These check should only happen for Normal Images (not for Live photo and video).
Also, if exifSays it's panorama, no need to check for xmp?

prateekmedia: For the first part, it is under a if where it only runs for file type image, hence file type live photo and video are eliminated.

ua741 (approved):

PR #2553, opened 2024-07-26
[mob][photos] Add preview to the album link

AmanRajSinghMourya:


mobile/lib/ui/sharing/manage_links_widget.dart
```diff
@@ -318,6 +330,52 @@ class _ManageSharedLinkWidgetState extends State<ManageSharedLinkWidget> {
     );
   }
 
+  Future<Uint8List> _createAlbumPlaceholder(
+    List<EnteFile> ownedSelectedFiles,
```
ashilkn: Rename this parameter?

mobile/lib/ui/sharing/manage_links_widget.dart
```diff
@@ -318,6 +330,52 @@ class _ManageSharedLinkWidgetState extends State<ManageSharedLinkWidget> {
     );
   }
 
+  Future<Uint8List> _createAlbumPlaceholder(
```
ashilkn: Since `_createAlbumPlaceholder` and `_shareAlbumLink` is written twice in both on `manage_links_widget.dart` and `share_collection_page.dart`, let move both to `share_util.dart` to avoid duplicate code. Can rename `shareAlbumLink` to `shareAlbumLinkWithPlaceholder` in `share_util.dart` and call it from both files.

mobile/lib/utils/share_util.dart
```diff
@@ -261,3 +267,70 @@ Rect _sharePosOrigin(BuildContext? context, GlobalKey? key) {
   }
   return rect;
 }
+
+/// The [ScreenshotController] is used to capture the image of a widget that serves as
+/// a placeholder for an album link. The placeholder is then shared along with the album link.
+final ScreenshotController screenshotController = ScreenshotController();
```
ashilkn: Better to create create an instance of `ScreenshotController` inside `shareAlbumLinkWithPlaceholder` and pass that `ScreenshotController` to `_createAlbumPlaceholder` from inside `shareAlbumLinkWithPlaceholder`.

Why: This variable now has a global scope, which is unnecessary in this case and could lead to issues in future. For example, someone could use this accidentally in another file.

mobile/lib/utils/share_util.dart
```diff
@@ -261,3 +267,70 @@ Rect _sharePosOrigin(BuildContext? context, GlobalKey? key) {
...
+
+Future<Uint8List> _createAlbumPlaceholder(
```
ashilkn: It's good practice and followed in this codebase to have public methods first and then private methods. So should move this method to end the of file (together with other private methods) and `shareAlbumLinkWithPlaceholder` before all private methods.

ashilkn (approved):

ashilkn (approved):

PR #2607, opened 2024-08-02
allow setting extra parameter for dsn

winston0410:
## Description
For https://github.com/ente-io/ente/discussions/2603


mnvr (approved):
Thank you!

PR #2631, opened 2024-08-07
[mob][photos] Swipe lock for multiple files

AmanRajSinghMourya:


mobile/lib/ui/viewer/actions/file_selection_actions_widget.dart
```diff
@@ -559,6 +566,23 @@ class _FileSelectionActionsWidgetState
...
   }
 
+  Future<void> _onGuestViewClick() async {
+    final List<EnteFile> selectedFiles = widget.selectedFiles.files.toList();
+    final page = DetailPage(
+      DetailPageConfiguration(
+        selectedFiles,
+        null,
+        0,
+        "guest_view",
```
ashilkn, line 576: Could you check what tag to pass here so that hero animations work just like when photos or videos are viewed from gallery without the guest view?

We use this tag to make up the tag for `Hero` widget.

mobile/lib/ui/viewer/file/detail_page.dart
```diff
@@ -235,10 +237,13 @@ class _DetailPageState extends State<DetailPage> {
         } else {
           _selectedIndexNotifier.value = index;
         }
+        Bus.instance.fire(FileSwipeLockEvent(isGuestView, swipeLocked));
         _preloadEntries();
       },
-      physics: _shouldDisableScroll || isFileSwipeLocked
-          ? const NeverScrollableScrollPhysics()
+      physics: _shouldDisableScroll || isGuestView
+          ? swipeLocked
```
ashilkn: If `_shouldDisableScroll` is true, scrollPhysics should be `NeverScrollableScrollPhysics()`, always. Here if `swipeLocked` is false, it won't work. 

Where it has affected: try zooming in an image and go to another area of the image.

mobile/lib/ui/viewer/file/file_app_bar.dart
```diff
@@ -69,7 +69,7 @@ class FileAppBarState extends State<FileAppBar> {
     _fileSwipeLockEventSubscription =
```
ashilkn: When adding a listener inside methods like `didUpdateWidget`, `didChangeDependencies` which could be called more than once in a Stateful Widget's lifecycle unlike `initState` or `dispose`, there is a chance of adding multiple identical listeners. 

So before adding a listener here, it's safe to cancel any existing listener. 
Can remove `late` keyword from `_fileSwipeLockEventSubscription` and make it nullable. If `_fileSwipeLockEventSubscription` is not null, that means a listener already exists, so cancel it before adding a new listener.

mobile/lib/ui/viewer/file/file_app_bar.dart
```diff
@@ -413,9 +413,9 @@ class FileAppBarState extends State<FileAppBar> {
     }
   }
 
-  Future<void> _onSwipeLock() async {
+  Future<void> _onTapGuestView() async {
     if (await LocalAuthentication().isDeviceSupported()) {
-      Bus.instance.fire(FileSwipeLockEvent(!_isFileSwipeLocked));
+      Bus.instance.fire(FileSwipeLockEvent(!isGuestView, true));
```
ashilkn: Isn't the value of `isGuestView` insignificant here? Won't the value that's passed be always `true`? 
Passing just `true` would make the code a little more good.

From where I'm coming: The moment I saw this, I was trying to figure out when will be passed value be `false` in this line, since that could be a possibility since we're passing a boolean _variable_ here.

mobile/lib/events/file_swipe_lock_event.dart
```diff
@@ -1,7 +1,7 @@
 import "package:photos/events/event.dart";
 
 class FileSwipeLockEvent extends Event {
```
ashilkn: Let's rename this to `GuestViewEvent`, since we're renaming 'Swipe lock' to 'Guest view' in the app.

ashilkn (approved):

PR #2695, opened 2024-08-14
[server] Send 5xx if query to validate token fails

ua741:


ua741:
Lint check is working fine locally, action seems to failing 
> compile: version "go1.22.6" does not match go tool version "go1.21.13" (compile)

PR #2705, opened 2024-08-14
Update README.md

brog-io:
I updated the sentence "Ente Photos is a paid service, but we offer a free trial. You can also clone this repository and choose to self host." to reflect that Ente now offers a permanent free option with 5GB of free storage.


mnvr (approved):
Thank you!

PR #2738, opened 2024-08-17
[mob][auth] Option to add pin/passcode without having a system lock

AmanRajSinghMourya:


auth/lib/utils/lock_screen_settings.dart
```diff
@@ -83,6 +84,14 @@ class LockScreenSettings {
     await _preferences.setInt(keyInvalidAttempts, count);
   }
 
+  Future<void> isAppLockSet(bool value) async {
```
ashilkn: There is a disconnect between the name of this function and what it does.

auth/lib/ui/settings/lock_screen/lock_screen_options.dart
```diff
@@ -89,27 +104,30 @@ class _LockScreenOptionsState extends State<LockScreenOptions> {
...
-            _configuration.shouldShowSystemLockScreen();
+        appLock = _lockscreenSetting.getIsAppLockSet();
       }
     });
   }
 
   Future<void> _onToggleSwitch() async {
     AppLock.of(context)!.setEnabled(!appLock);
-    await _configuration.setSystemLockScreen(!appLock);
+    await _configuration.setSystemLockScreen(false);
```
ashilkn: Do enable system lock by default if available.

PR #2752, opened 2024-08-19
[mob][auth] local_auth upgrade & work around for new applock key

AmanRajSinghMourya:


auth/lib/utils/lock_screen_settings.dart
```diff
@@ -38,6 +41,23 @@ class LockScreenSettings {
 
     ///Workaround for privacyScreen not working when app is killed and opened.
     await setHideAppContent(getShouldHideAppContent());
+
+    await setHasMigratedLockScreenChanges();
+  }
+
+  Future<void> setHasMigratedLockScreenChanges() async {
+    if (_preferences.getBool(keyHasMigratedLockScreenChanges) == null ||
```
ashilkn: Can change this to just `if (_preferences.getBool(keyHasMigratedLockScreenChanges) == null`.
I couldn't understand why we need to check `getUsAppLockSet` here.

auth/lib/utils/lock_screen_settings.dart
```diff
@@ -38,6 +41,23 @@ class LockScreenSettings {
...
+        getIsAppLockSet() == false) {
+      await _preferences.setBool(keyHasMigratedLockScreenChanges, true);
```
ashilkn: Move this to the end. So that only after everything is done we set this.
Why: let's say `isPinSet()` throws an error, the migration never completes but we set the key to true.

ashilkn (approved):

PR #2754, opened 2024-08-19
Updated Mastodon links

brog-io:
Changed the links in the social section and readme to Fosstodon instead of mstdn


mnvr (approved):
Thank you! 

cc @vishnukvmd

PR #2796, opened 2024-08-21
[mob][auth] Change dialog copy

ashilkn:


vishnukvmd (approved):

mobile/lib/l10n/intl_en.arb
ChiefMikeK, line 1: 🫨 Shouldn't there be this code inserted/replacing the deleted line ⸳ 1291
https://github.com/ente-io/ente/blob/0ae080ea6a906c39788cea5a3a1095f5b69de163/mobile/lib/l10n/intl_en.arb#L1291

+ `"deviceLockEnablePreSteps": "To enable app lock, please setup device passcode or screen lock in your system settings."`

PR #2817, opened 2024-08-22
Ml copy

laurenspriem:
## Description

Copy changes for ML


laurenspriem:
Can remove "all results will be synced automatically" if it makes the string too long

vishnukvmd (approved):

PR #2853, opened 2024-08-23
[docs] Authy guide obsolete notice

greeeen-dev:
## Description
As Authy no longer supports its desktop apps, moving from Authy to Ente Auth using the methods are not possible. This PR adds this notice to the guide.

it was fun while it lasted :pensive:


mnvr (approved):
😇🙏

PR #2856, opened 2024-08-23
[web][photos] Fix vertical mirror icon

Johannes7k75:
## Description
This PR changes the rotation of the vertical mirror icon to be rotated by 90deg

image
It's German but should show the fix


mnvr (approved):
Thank you!

PR #2911, opened 2024-08-25
[web][photos] Fix search icon layout & color

saibotk:
This fixes the search icon in the search box not being centered properly and also not having the muted color style.

The container is a flex-box layout which already tries to center its items. But due to the icons in MUI having a display of `inline-block`, our `Box` container also needs `display: inline-flex` to properly calculate the height. Now the box container is the same height as the icon and it gets centered correctly again.
This is the same methodology that the MUI native `IconButton` also uses
as a display value. See https://github.com/mui/material-ui/issues/33020#issuecomment-1149971549

Current web version:
image

Proposed fix:
image

This also removes the `icon` CSS class from the Box container since it does not exist anyway. And we fixed the theme color not being applied. This was otherwise just passed as a raw string to the color property and now correctly uses the intended theme color.

Thanks!


mnvr (approved):
Thank you, both for the commit and the explanation.

PR #2912, opened 2024-08-25
[web][photos] Fix web navbar button sizes

saibotk:
Two minor style fixes for the nav bar. 

1. Sidebar 3-lines IconButton had its left padding set to 0, which causes issues.
2. Upload icon on mobile had the wrong height and thus weird ripple / hover effects.

Please see the respective commit for details & the reasoning.

How those currently look:
image
image

How they look with the proposed fix:
image
image

Thanks!


mnvr (approved):
Thank you!

PR #2940, opened 2024-08-26
Remove extra space before comma

vishnukvmd:


mnvr (approved):
Thanks. And jfyi to avoid invalidating the existing translations, we can use the "Push Crowdin translations (web)" action, this uploads both the sources and the (remaining existing) translations.


I've already triggered it from this branch (before merging it) so that the sync action (which uploads only source strings) is a no-op.
```
gh workflow run web-crowdin-push-both.yml --ref trim-space
```

PR #2955, opened 2024-08-26
Remove X.com custom icon

araghon007:
Both Twitter and X.com already have a simple icon defined, making this custom icon redundant. Plus it's nice to be able to change which icon you prefer just by changing the account name between X and Twitter 

## Screenshot
image


ua741 (approved):

Eseltwift:
Hi there @araghon007!

I was going through some PRs related to icons and thought I’d share my thoughts on this one. Originally, the simple-icon library didn’t include the new X logo, which is why I added it. But you made the right call to remove it now that simple-icons has been updated.

That said, I’m a bit hesitant about removing the alt-name that merged X and Twitter. Since the Twitter brand no longer exists to the public and redirects to X.com, it seems a little unusual to offer users the option to use a logo that’s no longer around. But it's really not a big deal—everyone has their preferences, and I just wanted to share mine.

araghon007:
I assumed as such, and the name merger is fair, but I'd say for as long as the Twitter icon stays in simple icons, the separate name should probably stay as well. Thank you for replying, and you better not stop adding more icons :p Greatly appreciate the effort.

It might also be worth going through the rest of simple icons and removing any custom icons if they match.

PR #2991, opened 2024-08-27
Revert "[auth] Add & change icons"

Vantesh:
Reverts ente-io/ente#2954

GitHub icon is already fine as it is and removing hex leads to the icon being invisible on light theme.


Vantesh:
Ok...

Eseltwift:
Hi there, I've noticed that some of my changes have been reverted because the icons weren't visible in either dark or light mode. I wanted to suggest looking into the `mix-blend-mode:difference` property for SVGs. This property allows the color of an SVG to automatically invert based on the background color, which can help ensure visibility in both modes. For example, the Proxmox logo in the live version of the app uses this approach effectively (at least on iOS, which I can confirm).

Thanks for considering this!

Vantesh:
Hey, I don't think the changes were reverted since i closed the request and after looking into the mix blend property i found it actually amazing, only that on some editors the icon is invisible but as long as it works on ente it's good 👍

PR #3037, opened 2024-08-29
[mob][photos] Fix: audio not playing on iOS when in silent mode

ashilkn:
## Tests

Tested, everything works fine. 
If an audio is playing in the background from a different app and once a video starts playing on ente, the background audio is paused.


ashilkn:
> Otherwise, let's say something was playing in the background, and the device was on mute—will calling this method result in some audio getting unmuted?

No, I've already tested this and it works as expected by letting the background audio play. The background audio only stops when a video is played.

PR #3059, opened 2024-08-30
[mob][photos] Add option to enable or disable looping of videos

ashilkn:
https://github.com/user-attachments/assets/e6635b09-d0d2-4f7e-b75f-513899c996ad


mobile/lib/core/configuration.dart
```diff
@@ -661,6 +662,14 @@ class Configuration {
     await _preferences.setBool(hasSelectedAllFoldersForBackupKey, value);
   }
 
+  Future<void> setShouldLoopVideo(bool value) async {
```
ua741: Let's move this to localSettings? The key value can be `video.loop` or at least with `video.should_loop`

ua741 (approved):

PR #3108, opened 2024-09-03
[auth] Add Appbar Lock Icon for Manual App Locking on Desktop + fix bugs

AmanRajSinghMourya:
## Description

When app lock is enabled on Auth's desktop app and when closed/minimised, the app is not getting locked in certain cases. As a workaround for that, have added a button to manually lock the app.


auth/lib/ui/home_page.dart
```diff
@@ -199,6 +203,28 @@ class _HomePageState extends State<HomePage> {
...
 
+  Future<void> navigateToLockScreen() async {
+    final bool shouldShowLockScreen =
+        await Configuration.instance.shouldShowLockScreen();
+    if (shouldShowLockScreen) {
+      await AppLock.of(context)!.showLockScreen();
+    } else {
+      await showDialogWidget(
+        context: context,
+        title: "No app lock enabled",
```
ashilkn: Can change this to "App lock not enabled"

auth/lib/ui/home_page.dart
```diff
@@ -199,6 +203,28 @@ class _HomePageState extends State<HomePage> {
...
+        body: "Please enable app lock from Security > App Lock",
```
ashilkn: Need to extract all new strings.

ashilkn (approved):

PR #3131, opened 2024-09-04
[mob][photos] Require Authentication on App Reopen from Guest View

AmanRajSinghMourya:
This PR introduces a security enhancement that requires authentication when the app is reopened after being closed while in guest view.

### DIFF
--- a/mobile/lib/core/configuration.dart
+++ b/mobile/lib/core/configuration.dart
@@ -71,7 +71,6 @@ class Configuration {
       "has_selected_all_folders_for_backup";
   static const anonymousUserIDKey = "anonymous_user_id";
   static const endPointKey = "endpoint";
-  static const guestViewKey = "key_guest_view";
   static final _logger = Logger("Configuration");
 
   String? _cachedToken;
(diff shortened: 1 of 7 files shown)


mobile/lib/core/configuration.dart
```diff
@@ -71,6 +71,7 @@ class Configuration {
       "has_selected_all_folders_for_backup";
   static const anonymousUserIDKey = "anonymous_user_id";
   static const endPointKey = "endpoint";
+  static const guestViewKey = "key_guest_view";
```
ashilkn: Can move this to the local settings file

mobile/lib/core/configuration.dart
```diff
@@ -633,6 +634,14 @@ class Configuration {
     return _preferences.setBool(keyShowSystemLockScreen, value);
   }
 
+  Future<void> setGuestView(bool value) {
```
ashilkn: These too

mobile/lib/core/configuration.dart
ashilkn: Can rename this to `onGuestViewKey = "on_guest_view"`

mobile/lib/core/configuration.dart
```diff
@@ -633,6 +634,14 @@ class Configuration {
...
+    return _preferences.setBool(guestViewKey, value);
+  }
+
+  bool getGuestView() {
```
ashilkn: Can use better name `isOnGuestView`

mobile/lib/core/configuration.dart
ashilkn: `setOnGuestView`

ashilkn (approved):

PR #3132, opened 2024-09-04
[Auth] Add icon for GOV.UK

bsian03:
Adds the icon for UK government online services (Government Gateway)


Eseltwift:
Hey, I think you should add a fill color for the `mix-blend-mode` mode to work properly.

You could do something like this:
```diff
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 30">
<g style="mix-blend-mode:difference">
- <path d="M22.6 ... 12.3-.8"/>
+ <path fill="white" d="M22.6 ... 12.3-.8"/>
</g>
</svg>
```

Without an actual background coded in (as it would be in a web browser, for example), the SVG will appear white on white, but in the app, it will display correctly.

bsian03:
Thanks, must have missed it looking at the github svg as the example

ua741 (approved):

ua741 (approved):

PR #3141, opened 2024-09-05
[mob](upload-status): fixes

prateekmedia:
## Description

These PR does following fixes to upload status:
- Always maintain order or backup status.
  - failed -> in queue -> in background -> uploading -> uploaded
- Show previous uploaded items in this too


ua741 (approved):

mobile/lib/services/search_service.dart
```diff
@@ -67,6 +67,7 @@ class SearchService {
 
   Future<List<EnteFile>> getAllFiles() async {
     if (_cachedFilesFuture != null) {
+      _logger.fine("Reading all files from cache");
```
ua741: remove?

mobile/lib/ui/settings/backup/backup_status_screen.dart
```diff
@@ -20,12 +23,47 @@ class BackupStatusScreen extends StatefulWidget {
...
+          (e) {
+            return BackupItem(
+              status: BackupItemStatus.uploaded,
+              file: e,
+              collectionID: e.collectionID ?? 0,
+              completer: null,
+            );
+          },
+        )
+        .toList()
```
ua741: We need to ignore files that are not uploaded and also the one that are not owned by the user. Also, we can have multiple files with same uploadID.

ua741 (approved):

PR #3184, opened 2024-09-09
Minio Cold Storage

heftyzauk:
## Description
This PR is to allow use of Minio for cold storage, removing the GLACIER storage class that is usually set and is unsupported by Minio. 

By default if you set use "are_local_buckets" this breaks TLS on all buckets, so it is impossible to use public hot storage and self hosted cold storage. 

This option now allows for a hybrid approach. 

Note: TLS must be configured on the Minio instance.

### DIFF
--- a/server/configurations/local.yaml
+++ b/server/configurations/local.yaml
@@ -162,6 +162,12 @@ s3:
     #    not support them, specifically it doesn't support GLACIER).
     #
     #are_local_buckets: true
+
+    # Allow use of Minio (Configured with TLS) as cold storage, as Minio does not support glacier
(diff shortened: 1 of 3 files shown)


mnvr:
We're already working on the sides to make the replication more configurable (e.g. see this comment - https://github.com/ente-io/ente/discussions/3167#discussioncomment-10585970). As such, adding this ad-hoc option, while useful for your use case, is just going to increase the configuration surface that we'll have to deprecate when we provide the new configuration mechanism.

It's nice that you opened this PR though, thank you. This way people can find this patch and apply it locally if they wish (until the more configurable configuration is in-place).

PR #3192, opened 2024-09-09
[desktop][photos] Use antialias with clip image

laurenspriem:
## Description

Adds antialias to clip image preprocessing

## Tests

Only tested the mobile implementation which follows exactly the same logic. So review carefully.


mnvr:
Have pushed some mostly non-functional commits: the Array construction for the kernel needed a zero-fill to initialize, have fixed that + some minor syntactic tweaks to satisfy the typechecks.

mnvr (approved):

PR #3198, opened 2024-09-10
Added Capacities logo to icons and to custom-icons.json

legomstr1:


legomstr1:
Added in the Capacities logo and added the Capacities to the json for ente auth

ua741 (approved):

Issue #1510, opened 2024-04-21
[cli] Failed to encrypt invalid key when adding account


vulongm:
### Description

I'm currently using ente-cli-v0.1.13-linux-amd64 on Fedora Kinoite 40 and ran into an error when trying to add an account.
```
$ ./ente account add
Enter app type (default: photos): 
Use default app type: photos
Enter export directory: /var/home/user/Photos/
Enter email address: email-here
Enter password: 
Please wait authenticating...
Enter TOTP: totp-here
2024/04/21 12:23:19 failed to encrypt invalid key
```

### Version

v0.1.13

### What product are you using?

Ente Photos

### What platform are you using?

Desktop - Linux

Tritri1029:
Have you found a solution?

vulongm:
I have not.

maffinca69:
Any solutions?

shoetten:
I have the same problem, but found a workaround. Ente saves a so called "Device Key" in your system keyring / wallet and expects it to be 32 bytes. That key somehow got corrupted in my keyring. Deleting it helped (ente-cli will create a new one if none exists) and i authorized my account successfully.

Try looking for `ente` in your system keyring (e.g. KDE Wallet or i believe GnomeKeyring / Seahorse under Gnome) and deleting the entry.

**Background**

While trying to debug the issue, i found that the underlying cause is that the key stored in my keyring is `66` bytes instead of the expected `32`. I don't know why, but it's obvious that `NewEncryptor` will [fall over with the invalid key error](https://github.com/ente-io/ente/blob/main/cli/internal/crypto/stream.go#L80) when that is the case.

Hope that helps someone with actual understanding of Go / this codebase, to fix the underlying issue.

shoetten:
Unfortunatley, that only gets you half way there, because the next operation you execute will fail again with a similar error. Apparently what is stored in the keyring is not the same that comes out. I expect something with the `string(secret)` and `[]byte(secret)` casting being wrong. I will try to dig deeper..

shoetten:
I made a minimal reproduction here, to better understand the issue:
https://github.com/shoetten/ente-1510-repro

One can clearly see, that what you store in the keyring in not what comes out (at least when running under Linux, i suspect the other platforms work, otherwise this issue would surly have a higher priority..)

The problem always occurs when handling characters, that are not represented by a single byte, so basically everything not ASCII. I don't know whether that's a bug in go-keyring or even deeper in the stack.

[...]

PR #3208, opened 2024-09-10
[cli] Base64 Encode the CLI secrets

ua741:
## Description

Related to #1510 & [#3163](https://github.com/ente-io/ente/pull/3163)

This potential fix was only possible due to amazing investigation done by @shoetten

## Tests
Did basic sanity testing on local machine


shoetten:
Nice, much better than my quick hack ;) especially with the recovery for users with exisiting broken keys.

PR #3241, opened 2024-09-12
[auth] added issuer name in confirmation dialog when deleting the code

### LINKED ISSUE #459: Mention name of item to delete in confirmation dialogues
**Is your feature request related to a problem? Please describe.**
When deleting an item from ente Auth, there are two confirmation screens: The first asks for my fingerprint, the seconds one is a final confirmation. However, in both dialogues the name of the item I want to delete is missing. So I do not know if I really selected the correct one.

**Describe the solution you'd like**
Both screen should mention the name of the item to delete.

kishan-dhankecha:
Fixes #459 

I don't know how the contribution to the translations is handled. I just updated the keys with description and placeholder.


ua741:
Hello, we are already working on updating this while working on other features for auth. As this require copy change, will discuss internally and fix it. Thanks for the PR :)

PR #3243, opened 2024-09-12
[auth] Fix duplicate tags entry

### LINKED ISSUE #1794: [Bug] [Auth] Duplicate tags visual bug
### Description

When creating a new tag, it is possible to uncheck it and then create another one with the same name (or alternatively, create a tag with the same name as one that already exists but is unused by the current service).
These duplicated tags will then function as a unit (either all checked, or all unchecked).

See the video below for further context:

https://github.com/ente-io/ente/assets/86177399/3f3fa926-49d1-4d70-b5e0-06ed2cfc71a4

This is only a visual bug and doesn't affect the state of the application when going back to the home screen, or refreshing.

The bug was reproduced on Linux Desktop ([AUR](https://aur.archlinux.org/packages/ente-auth-bin) v3.0.1) and Android Mobile (Google Playstore v3.0.0)

### Version

v3.0.1

Ente Auth

Mobile - Android, Desktop - Linux

kishan-dhankecha:
fixes #1794

### DIFF
--- a/auth/lib/onboarding/view/setup_enter_secret_key_page.dart
+++ b/auth/lib/onboarding/view/setup_enter_secret_key_page.dart
@@ -31,7 +31,7 @@ class _SetupEnterSecretKeyPageState extends State<SetupEnterSecretKeyPage> {
   late TextEditingController _accountController;
   late TextEditingController _secretController;
   late bool _secretKeyObscured;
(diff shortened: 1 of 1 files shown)


prateekmedia (approved):
LGTM! Thanks.

PR #3251, opened 2024-09-12
[server] Optional SMTP authentication

tobiasge:
## Description
When using a local relay server, not authentication is needed. To support this use-case authentication is only send to the server when an user and password are set in the configuration file.

See error message from https://github.com/ente-io/ente/discussions/3179


mnvr (commented):
There is a `server/museum` binary in the PR. Please force push to reset the PR with only a clean commit.

While we can merge this, I'll mention that we're generally averse to adding branching logic to the server side code. For example, with this change, it is possible that someone might accidentally omit credentials and not be notified of it by an error message.

mnvr (approved):
Thank you!

Issue #2581, opened 2024-07-30
Can't play videos on firefox


pinpox:
### Description

When visiting an album in using the latest Firefox on Windows 11, videos can't be played back. Here is an example album I uploaded for demostration

https://albums.ente.io/?t=hcSStFTf#FFffhDgScXVMD6joprQEcX6XYXCrNh89xQ7JFo2R4btJ

The video transcoding using ffmpeg-wasm seems to be broken, chrome works because it does probably not need to transcode as the native format is supported. The video can be played fine when downloaded. It was recorded using Iphone 12 Mini. 

I'm using the official hosted version as you can see, the same issue happens on selfhosted aswell though. In addition this are the debug logs: 
[debug_logs_1722370846323.txt](https://github.com/user-attachments/files/16433160/debug_logs_1722370846323.txt)

### Version

Current official Web UI (2024-07-30)

### What product are you using?

Ente Photos

### What platform are you using?

Web - Firefox

Floehe:
Hi,
I have the same issue in current Safari and Firefox on macOS. Also in Safari on the iPhone.
The video is playing fine in the macOS Desktop App and iPhone app.
The backend and web apps are self-hosted.

Floehe:
Hi @mnvr ,
I just checked out the new code for the web app. 

It is now working in Safari on iPhone. (although it takes much longer to convert the video to a playable format then in the app).
I still can not get it to display in Firefox on Windows, it says: "No video with supported format an MIME type found."
On Windows In MS Edge, the video player starts up really quick, but I only get sound, no picture.
On macos in Safari, I get a video player after some loading time, but it stays empty, nothing is playing when pressing the play button, also no sound.
On macos in Firefox, it also takes longer than in the app to transcode, but it is working fine then!

Windows:
* Firefox: ❌
* MS Edge: quick opening of player, only sound, no video ❌

macos:
* Firefox: ✔
* Safari: ❌

If I can provide any extra information, just let me know.

mirh:
Some vague info on the codec wouldn't hurt? It's an mp4, but if it's H265 encoded then it was normal for firefox not to support it.

PR #3355, opened 2024-09-20
[web] Transcode videos if they are smaller than 100 MB

mnvr:
Fixes: https://github.com/ente-io/ente/issues/2581

### DIFF
--- a/web/apps/photos/src/components/PhotoFrame.tsx
+++ b/web/apps/photos/src/components/PhotoFrame.tsx
@@ -492,7 +492,9 @@ const PhotoFrame = ({
             );
             fetching[item.id] = true;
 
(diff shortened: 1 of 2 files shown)


pinpox:
Why the limitation to 100 MB? Could this be made configurable for self-hosted instances or set to transcode everything into a sensible format?

mnvr:
> Why the limitation to 100 MB?

Web transcoding uses Wasm, which is going to be unusably slow for large videos. 100 MB is an arbitrary threshold to attempt to give a stop gap good experience for (majority) small clips while we come up with a different approach.

> Could this be made configurable for self-hosted instances 

You can configure the number to be anything if you build from source (change the constant you see in the PR)

> set to transcode everything into a sensible format?

We are working on supporting this - both video conversion and streaming support.

PR #3386, opened 2024-09-21
[mob][photos] Implement deep linking for public links

AmanRajSinghMourya:
## Description

Opening an Ente photos public link in the app if the app is installed on device.


mobile/lib/services/collections_service.dart
```diff
@@ -1030,6 +1035,84 @@ class CollectionsService {
     }
   }
 
+  Future<Collection> getPublicCollection(Uri uri) async {
+    final String authToken = uri.queryParameters["t"] ?? "Not found";
```
ashilkn: Can write `uri.queryParameters["t"]!` instead of assigning "Not found" here. 
This is assuming `uri.queryParameters["t"]` is never null.

ashilkn: Even if it's nullable, this is not the right way to handle it. We should log and throŵ an `Exception` instead.

mobile/lib/services/collections_service.dart
```diff
@@ -1030,6 +1035,84 @@ class CollectionsService {
...
+          headers: {"X-Auth-Access-Token": authToken},
+        ),
+      );
+
+      final collectionData = response.data["collection"];
+      final Collection collection = Collection.fromMap(collectionData);
+      final Uint8List collectionKey =
+          Uint8List.fromList(Base58Decode(albumKey));
+
+      _logger.severe("Public collection fetched: $collectionData");
```
ashilkn: Let's remove logging this. 
If we're keeping it, need to use `_logger.info` and not log `collectionData`

mobile/lib/services/collections_service.dart
```diff
@@ -1030,6 +1035,84 @@ class CollectionsService {
...
+        "https://api.ente.io/public-collection/verify-password",
+        data: {"passHash": passwordHash},
+        options: Options(
+          headers: {
+            "X-Auth-Access-Token": authToken,
+            "Cache-Control": "no-cache",
+          },
+        ),
+      );
+      final jwtToken = response.data["jwtToken"];
```
ashilkn, line 1108: Do not log these

mobile/lib/services/collections_service.dart
```diff
@@ -1030,6 +1035,84 @@ class CollectionsService {
...
+      _logger.severe("TOKEN $authToken");
+      _logger.severe("JWT TOKEN $jwtToken");
+      if (response.statusCode == 200) {
+        await setPublicAlbumTokenJWT(collectioID, jwtToken);
+        return true;
+      }
+      return false;
+    } catch (e) {
+      _logger.severe("Failed to verify public collection password $e");
```
ashilkn: -> `_logger.warning("Failed to verify public collection password", e)`

mobile/lib/ui/tabs/home_widget.dart
```diff
@@ -224,6 +231,89 @@ class _HomeWidgetState extends State<HomeWidget> {
...
+
+  Future<void> _initDeeplinkPublicAlbum() async {
+    // Handle the terminated state
+    final Uri? uri = await getInitialUri();
+    if (uri != null) {
+      await _handlePublicAlbumLink(uri);
+    }
+
+    // Handle the background state
+    uriLinkStream.listen((Uri? uri) async {
```
ashilkn: Should cancel this stream subscription (When app is terminated I think? Cancelling it in home widget's dispose method should do the work).

mobile/lib/ui/tabs/home_widget.dart
```diff
@@ -224,6 +231,89 @@ class _HomeWidgetState extends State<HomeWidget> {
...
+        title: S.of(context).enterPassword,
+        submitButtonLabel: S.of(context).ok,
+        alwaysShowSuccessState: false,
+        onSubmit: (String text) async {
+          if (text.trim() == "") {
+            return;
+          }
+          try {
+            final hashedPassword = await CryptoUtil.deriveKey(
+              utf8.encode(text) as Uint8List,
```
ashilkn: Unnecessary type cast, try removing it.

mobile/lib/ui/tabs/home_widget.dart
```diff
@@ -224,6 +231,89 @@ class _HomeWidgetState extends State<HomeWidget> {
...
+              CryptoUtil.base642bin(publicUrl.nonce!),
+              publicUrl.memLimit!,
+              publicUrl.opsLimit!,
+            );
+            _logger.info(
```
ashilkn: Can remove logging this.

mobile/lib/ui/tabs/home_widget.dart
```diff
@@ -224,6 +231,89 @@ class _HomeWidgetState extends State<HomeWidget> {
...
+            );
+
+            result = await CollectionsService.instance
+                .verifyPublicCollectionPassword(
+              CryptoUtil.bin2base64(hashedPassword),
+              collection.id,
+            );
+            if (!result) {
+              await showErrorDialog(
+                context,
```
ashilkn: These strings are already extracted for translation, use them.

mobile/lib/ui/viewer/gallery/collection_page.dart
```diff
@@ -24,13 +24,17 @@ class CollectionPage extends StatelessWidget {
...
+  final List<EnteFile> sharedLinkFiles;
 
   CollectionPage(
     this.c, {
     this.tagPrefix = "collection",
     this.hasVerifiedLock = false,
     this.isFromCollectPhotos = false,
     Key? key,
+    this.isFromPublicShareLink = false,
+    this.sharedLinkFiles = const [],
```
ashilkn: I feel it's better to make this nullable and null be its default value. So that it can be easily identified if `sharedLinkFiles` was passed (non-null) or not (null). 

If we use an empty list by default, it would be hard to know if the link has no files or if a `sharedLinkFiles` was never passed. In the future the need to know this might arise.  

We can also add an `assert` that if `isFromPublicShareLink == true`, `sharedLinkFiles` cannot be null. 

https://dart.dev/language/error-handling#assert

mobile/lib/utils/file_download_util.dart
```diff
@@ -24,14 +24,96 @@ import "package:photos/utils/file_util.dart";
...
 
+Future<File?> downloadAndDecryptPublicFile(
+  EnteFile file,
+  String authToken, {
+  ProgressCallback? progressCallback,
+}) async {
+  final String logPrefix = 'Public File-${file.uploadedFileID}:';
+  _logger
+      .info('$logPrefix starting download ${formatBytes(file.fileSize ?? 0)}');
+  _logger.severe("File id ${file.uploadedFileID}");
```
ashilkn: Should remove this log

ashilkn (approved):

PR #3449, opened 2024-09-24
[auth] landscape bottom actions padding

prateekmedia:
## Description

Remove bottom action landscape padding


ua741:
I had merged a fix for this yesterday and tested on iPad Simulator. Does it fix something else?

PR #3467, opened 2024-09-25
Clip with preprocessing inside ONNX on mobile

laurenspriem:
## Description

Switched to new clip ONNX model on mobile, where preprocessing is done inside ORT. This means it's more performant and more advanced image processing.

## Tests

Tested in debug mode on my Pixel 8.

### DIFF
--- a/mobile/lib/services/machine_learning/semantic_search/clip/clip_image_encoder.dart
+++ b/mobile/lib/services/machine_learning/semantic_search/clip/clip_image_encoder.dart
@@ -1,12 +1,10 @@
-import "dart:typed_data" show Uint8List;
+import "dart:typed_data" show Int32List, Uint8List;
 import "dart:ui" show Image;
 
 import "package:logging/logging.dart";
-// import "package:onnx_dart/onnx_dart.dart";
+import "package:onnx_dart/onnx_dart.dart";
 import "package:onnxruntime/onnxruntime.dart";
 import "package:photos/services/machine_learning/ml_model.dart";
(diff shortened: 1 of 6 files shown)


ua741 (approved):

mobile/lib/services/machine_learning/semantic_search/clip/clip_image_encoder.dart
```diff
@@ -1,15 +1,15 @@
...
 import "package:logging/logging.dart";
 import "package:onnx_dart/onnx_dart.dart";
 import "package:onnxruntime/onnxruntime.dart";
 import "package:photos/services/machine_learning/ml_model.dart";
-import "package:photos/utils/image_ml_util.dart";
 import "package:photos/utils/ml_util.dart";
 
 class ClipImageEncoder extends MlModel {
-  static const kRemoteBucketModelPath = "mobileclip_s2_image.onnx";
+  static const kRemoteBucketModelPath =
```
ahxxm, line 11: is this model generating embeddings in a same space as before? if not so, existing embeddings will need invalidation, `++clipMlVersion`

laurenspriem: Yes it's essentially the same model operating in the same embedding space. It's just the preprocessing that has moved to a different layer. So totally compatible with existing embeddings.

ahxxm: I see!

PR #3486, opened 2024-09-26
Added pritunl icon

AgentT30:
Added pritunl icon

@ua741 please review

### DIFF
--- a/auth/assets/custom-icons/_data/custom-icons.json
+++ b/auth/assets/custom-icons/_data/custom-icons.json
@@ -5,7 +5,7 @@
     },
     {
       "title": "23andme"
-    },    
+    },
     {
       "title": "3Commas"
     },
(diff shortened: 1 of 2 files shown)


Eseltwift:
Hey 👋 

Allow me to make a few suggestions for improvement.

First, the section on Pritunl should be added between PostNL and Privacy Guides to maintain the alphabetical order of the JSON file.

Additionally, the `altNames` field is only required if several names can refer to the same logo. **Icon names are case insensitive.**

Here is an example:
```json
"title": "Battle.net",
"slug": "battlenet",
"altNames": [
  "Battle net",
  "Blizzard"
]
```

The main/official name is `Battle.net`, but the file name is different, so we're adding a `slug` field to point to the correct file. Finally, we can assume that some people will call their entry in a different way, so we're planning alternative names such as `Battle net` and `Blizzard`.

So there's no need to add an alternative name in your case, since it's identical (not case-sensitive) to the name field.

ua741:
@AgentT30  I am closing this PR, feel free to open a new one with the improvements suggested by @Eseltwift

PR #3502, opened 2024-09-28
[auth] fix x64 installer on arm64 Windows

omove:
Portable x64 Ente Auth works on arm64 Windows, but the installer version fails due to not supporting arm64. As per Inno Setup's documentation the 'x64' option will only allow installing on x64 Windows, changing to 'x64compatible' allows the x64 installer to work on arm64 as well.


prateekmedia (approved):
LGTM! Thanks.

PR #3522, opened 2024-09-29
[mob] Misc bug fixes

ua741:


vishnukvmd (approved):

laurenspriem (commented):
I don't know if the `Fix hide cluster properly` is correct here. iirc we on purpose decided to not use the `isHidden` flag and instead reserve that for when we want to implement proper hiding of people (i.e. photos with person not showing up in gallery at all). I remember we had a discussion on this and decided to use empty name instead.

PR #3545, opened 2024-10-01
[auth] Auth lockscreen fixes

AmanRajSinghMourya:
## Description

1. Removed the logout button from the lockscreen if app in without backup state
2. Show a warning dialog when user try to set the app lock for the 1st time

### DIFF
--- a/auth/lib/ui/settings/security_section_widget.dart
+++ b/auth/lib/ui/settings/security_section_widget.dart
@@ -151,7 +151,7 @@ class _SecuritySectionWidgetState extends State<SecuritySectionWidget> {
         onTap: () async {
           ButtonResult? result;
           if (_config.hasOptedForOfflineMode() &&
-              !LockScreenSettings.instance.getOfflineModeWarningStatus()) {
+              LockScreenSettings.instance.getOfflineModeWarningStatus()) {
             result = await showChoiceActionSheet(
               context,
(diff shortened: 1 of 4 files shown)


ashilkn:
@AmanRajSinghMourya Can change the copy to "You have chosen to proceed without backups. If you forget your applock, you will be locked out from accessing your data."

auth/lib/ui/settings/security_section_widget.dart
```diff
@@ -146,10 +149,27 @@ class _SecuritySectionWidgetState extends State<SecuritySectionWidget> {
         trailingIcon: Icons.chevron_right_outlined,
         trailingIconIsMuted: true,
         onTap: () async {
+          ButtonResult? result;
+          if (_config.hasOptedForOfflineMode() &&
+              !LockScreenSettings.instance.getOfflineModeWarningStatus()) {
```
ashilkn: Names aren't used coherently. `getOfflineModeWarningStatus` returns a `bool` but isn't clear how the status and boolean returns are related. 

`getOfflineModeWarningStatus` returns the key `ls_show_offline_mode_warning`. So it's expected if `getOfflineModeWarningStatus` returns false, we should not show the warning. But the use of `getOfflineModeWarningStatus` here isn't following that expectation.

ashilkn (approved):

PR #3577, opened 2024-10-04
[auth] update dependencies for ios

prateekmedia:

### DIFF
--- a/auth/lib/ui/settings/language_picker.dart
+++ b/auth/lib/ui/settings/language_picker.dart
@@ -181,8 +181,6 @@ class _ItemsWidgetState extends State<ItemsWidget> {
         return '日本語';
       case 'ko':
         return '한국어';
-      case 'ar':
-        return 'العربية';
       case 'fa':
         return 'فارسی';
       default:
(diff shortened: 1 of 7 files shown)


auth/lib/ui/settings/language_picker.dart
```diff
@@ -181,8 +181,6 @@ class _ItemsWidgetState extends State<ItemsWidget> {
         return '日本語';
       case 'ko':
         return '한국어';
-      case 'ar':
-        return 'العربية';
```
vishnukvmd, line 185: Wondering if this is supposed to be a part of this change?

prateekmedia: This is a part of lint fixes, if you see 'ar' is repeated key, it is the first case also.

prateekmedia: See: https://github.com/ente-io/ente/blob/83990e11601e52caf01dadc6cb31aad26047c705/auth/lib/ui/settings/language_picker.dart#L127-L128

vishnukvmd (approved):

PR #3582, opened 2024-10-05
[auth][macos] show hidden window when user clicks on the dock

tmikus:
## Description

In the current version of Ente Auth, nothing happens when the user closes the Ente Auth window on macOS and then clicks on the Ente Auth icon in the dock.

This PR changes this behaviour and brings it in-line with other macOS applications by showing the main application window when the user clicks on the dock icon.


macgeeknz:
@tmikus thank you so much for these commits! They address the two main peeves I have with Ente Auth and make it so much better! I would have created them myself if I had the skills, but alas I do not, so I am very appreciative of your efforts :)

PR #3631, opened 2024-10-08
[web] add favorite button to SelectedFileBar to bulk favorite selection

Apalepex:
## Description
This allows on the web app to bulk favorite multiple photos in one go, by adding a favorite button to the SelectedFileOptions.tsx component. Just like for the Android App.
This could be enhanced by having the button plain or empty depending the selection and by setting/unsetting favarite depending previous status. For now it just set as favorite whatever the status (like the Android app).

## Tests
- Select at least 2 pictures in another album than hidden, favorited, trash, ...
- Click at top right of the screen the favorite icon
- the favorite icon on the previously selected thumbnails should appear

### DIFF
--- a/web/apps/photos/src/utils/file/index.ts
+++ b/web/apps/photos/src/utils/file/index.ts
@@ -24,7 +24,7 @@ import { LS_KEYS, getData } from "@ente/shared/storage/localStorage";
 import type { User } from "@ente/shared/user/types";
 import { downloadUsingAnchor } from "@ente/shared/utils";
 import { t } from "i18next";
-import { addMultipleToFavorites, addToFavorites, moveToHiddenCollection } from "services/collectionService";
+import { addMultipleToFavorites, moveToHiddenCollection } from "services/collectionService";
 import {
     deleteFromTrash,
     trashFiles,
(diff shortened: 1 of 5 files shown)


web/apps/photos/src/utils/file/index.ts
```diff
@@ -681,3 +690,11 @@ const fixTimeHelper = async (
 ) => {
     setFixCreationTimeAttributes({ files: selectedFiles });
 };
+
+const setBulkFavorite = async (files: EnteFile[]) => {
+    try {
+        await addMultipleToFavorites(files);
+    } catch (e) {
+        log.error("Could not add to favorites", e);
```
mnvr, line 698: This is swallowing the error, and will cause the local state to get out of sync with remote. I'll merge and fix.

As a side note, we might have other (sometimes temporary) reasons why a particular change has been made yet. And with code changes, is also hard for you to test exhaustively (e.g. your PR yesterday was causing the [shared albums to stop loading](https://github.com/ente-io/ente/commit/ac9e9079619fb8968a6ef254b1ddc4d4b13c778d)).

Apalepex: Ok thank you for you review and time. I'll check your fix in following commits so I'll learn.
Indeed, hard for me to make exhaustive tests. If your prefer I can tell you beforehand starting a new change so you can advice not to do it, or to pay attention to some point while testing.
Or I can continue like this. Up to you.

PR #3926, opened 2024-11-04
make the docs more verbose

mngshm:


cli/internal/promt.go
```diff
@@ -225,7 +225,7 @@ func ValidateDirForWrite(dir string) (bool, error) {
 }
 
 func ResolvePath(path string) (string, error) {
-	if path[:2] != "~/" {
+	if path[:1] != "~/" {
```
ua741, line 228: Is there a reason for changing this? Don't know if this will be backward compatible change

PR #3952, opened 2024-11-05
[server] Improvements to Dockerfile and compose.yaml

pzerone:
## Description
Gets rid of `libsodium-dev` installation on the final image build, instead copy the runtime libs over. shaves around 10 seconds on a clean `docker compose up -build` and around 3 MB on the server image. Not much, but hey..

This commit also switches the postgres container image from default `bookworm-slim` to `alpine`


ua741:
Hey, Thank you digging into this, and making this particular change.
For the time-being, I am afraid that we won't merge it as the delta isn't much. We want to avoid unnecessary change that can potential affect production.

PR #3963, opened 2024-11-06
[mob] add workmanager for android

prateekmedia:
## Description

Add support for work manager as it is more advanced and supports more range of devices in android side.
The iOS side is still experimental for the periodic tasks so background_fetch could be kept for now.


mobile/lib/app.dart
```diff
@@ -87,7 +84,11 @@ class _EnteAppState extends State<EnteApp> with WidgetsBindingObserver {
         : MediaExtentionAction(action: IntentAction.main);
     AppLifecycleService.instance.setMediaExtensionAction(mediaExtentionAction);
     if (mediaExtentionAction.action == IntentAction.main) {
-      _configureBackgroundFetch();
+      if (!enableWorkManager) {
+        _configureBackgroundFetch();
+        return;
+      }
+      _configureWorkManager();
     }
```
ua741: nit:
Can use simple if else here instead of adding return;

ua741 (approved):

ua741:
@prateekmedia Closing this PR. Once video streaming is done, then we will pick up background sync on both platform

PR #3985, opened 2024-11-08
[mob][photos] Concatenate strings more efficiently

ashilkn:
## Description

If number of iterations are high, concatenating a list of values to a string using `str1 += str2` can block the UI thread. 
Using `List.join()` is a much more efficient way to do this.


ua741 (approved):

ua741:
Can we add a linter for same?

ashilkn:
@ua741 Will check.

PR #3995, opened 2024-11-09
An attempt to fix the Custom Icon for Coinbase

NanoBunTV:
The icon currently only displays as a small black circle in the app.
I've replaced the SVG file entirely, and hopefully that'll fix the issue.


NanoBunTV:
Looks like this was already fixed in a recent update, but my fix would still fix the icon size at least, as now it's no longer a really small circle, it's just a really small Coinbase logo.

ua741 (approved):

PR #4028, opened 2024-11-13
[cli] Don't perform CLI secret initialisation unless needed

tigattack:
## Description

Calling `GetOrCreateClISecret()` can cause issues in some unattended scenarios.

I'm creating a Brew formula for ente CLI with the following test:

```
  test do
    assert_match "Version #{version}", shell_output("#{bin}/ente version")
  end
```

When running `brew test` for the formula, ente would fail with the following error: `2024/11/13 11:49:56 error setting password in keyring: exit status 154`

And I would see this pop-up:

image

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -49,14 +49,28 @@ func main() {
 			panic(err)
 		}
 	}
+
+	// Define a set of commands that do not require KeyHolder initialisation.
+	skipKeyHolderCommands := map[string]struct{}{"version": {}, "docs": {}, "help": {}}
+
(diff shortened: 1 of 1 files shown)


S33G:
nice

mathieutu:
@tigattack Hey, sorry I'm just seeing your message.
I already made the [PR](https://github.com/Homebrew/homebrew-core/pull/201919) to Homebrew. 😬 
Linux test does not pass, feel free to update it if you understand why. 🙂

@ua741 Could we release this change? We won't probably be able to see the formula merged until the error is fixed. 

Thanks!

PR #4039, opened 2024-11-15
Ml banner

laurenspriem:
## Description

Add banner in home gallery for enabling ML. 

## Tests

Tested in debug mode on my Pixel 8.

### DIFF
--- a/mobile/lib/generated/intl/messages_ar.dart
+++ b/mobile/lib/generated/intl/messages_ar.dart
@@ -31,6 +31,8 @@ class MessageLookup extends MessageLookupByLibrary {
             MessageLookupByLibrary.simpleMessage("Checking models..."),
         "decrypting": MessageLookupByLibrary.simpleMessage("فك التشفير..."),
         "email": MessageLookupByLibrary.simpleMessage("البريد الإلكتروني"),
+        "enableMachineLearningBanner": MessageLookupByLibrary.simpleMessage(
+            "Enable machine learning for magic search and face recognition"),
         "enterYourEmailAddress":
             MessageLookupByLibrary.simpleMessage("أدخل عنوان بريدك الإلكتروني"),
         "enterYourRecoveryKey":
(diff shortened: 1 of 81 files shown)


simondubrulle:
Hi @laurenspriem , I am trying to contribute to the translations of Ente Photos (#4040 ) and I see in existing PRs that whenever you introduce a new string, then you also introduce them in each arb file. I therefore guess any contributor should follow that practice. I am trying to understand and summarise this knowledge in a doc. My questions are:

1. Is there any command you are using to do that automatically? Or is it something you are doing manually?
2. I tried to check the workflows/Github actions and it seems like there is no automation for generating the messages_XX.dart file (for example after Crowdin sync). Is it something that has to be generated and comitted by a developer? Even though the commit was initially fixing something else? My interrogation is that is seems I might end up generating some messages_XX.dart files with information not related to the introduced new string (only because there is new content in arb files and the .dart files have not been generated yet).

Thanks in advance.

laurenspriem:
Hey, if you're trying to contribute to the translations then the best way to do that is [directly through Crowdin](https://github.com/ente-io/ente/blob/main/CONTRIBUTING.md#translate). Ideally the arb files should only be directly changed by the developers.

simondubrulle:
Thank you for the answer. I have actually already contributed to the remaining translations in Crowdin, but then while playing a bit with the app I started to see some texts which were still in English with no entry in Crowdin. That's where I started to check the code and saw there was still a series of hardcoded english strings. I just wanted to help extract those missing strings and make them available in Crowdin (#4040). I also understood those other arb files (not the en template) will eventually be synced in the repo via the weekly Crowdin Github action (hence the question above related to the manual creation).

Is it then fair to say that external contributors should not try to extract those strings? Would creating an issue (or discussion) be a better solution then?

laurenspriem:
Understood.

First of all, thanks for the contributions you already made with the translations!

Secondly, thinking on it more I don't see any issues with external contributors extracting those strings. You could create a PR for it and then tag me or one of the other developers in it. 

So let me tell you how I usually extract the strings, and then you can see if that works for you. 
1. Open Visual Studio Code, with the [Flutter Intl](https://marketplace.visualstudio.com/items?itemName=localizely.flutter-intl) extension installed, in the mobile directory
2. Select the string you want to extract, including the quotation marks
3. Press `Ctrl + .` to open quick actions, and press "Extract to ARB". That should replace the string and generate/modify the arb files
4. Commit the changes

Please let me know if this doesn't work for you, and thanks again for the contributions!

PR #4197, opened 2024-11-27
[API] Add Cache-control: no-store to API responses

mnvr:
...instead of adding it to ad-hoc API requests.

---
Tests: Smoke tested photos, shared albums and cast on localhost.


ua741 (approved):

mnvr:
The lint error for mobile seems unrelated to this PR, merging.

PR #4215, opened 2024-11-27
[mob][photos] Update change log

ashilkn:


simondubrulle:
Hi, shouldn't it be "Public links tapped from outside the app now **open** in-app." ("open" instead of "opens")?

ashilkn:
> Hi, shouldn't it be "Public links tapped from outside the app now **open** in-app." ("open" instead of "opens")?

Yes, but too late to change it now.

PR #4228, opened 2024-11-28
Workaround for Admins list in Environment Variables by specifying singular admin

Inumedia:
## Description

This allows working around an issue with Viper where it doesn't properly unmarshal GetIntSlice from environment variables. Seems like this is a known issue and is quite old. I'm assuming there isn't plans to fix it on the Viper side.

https://github.com/spf13/viper/issues/339

## Tests

I'll be brutally honest, I don't know how to run Go, but this seemed like a pretty quick low hanging fix.

### DIFF
--- a/server/pkg/middleware/auth.go
+++ b/server/pkg/middleware/auth.go
@@ -93,6 +93,12 @@ func (m *AuthMiddleware) AdminAuthMiddleware() gin.HandlerFunc {
 		}
 		// if no admins are set, then check if the user is first user in the system
 		if len(admins) == 0 {
+			// Fallback for specifying a singular admin ID when encountering https://github.com/spf13/viper/issues/1611
+			if viper.GetInt("internal.admin") == userId {
+				c.Next()
+				return
+			}
+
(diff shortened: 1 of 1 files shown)


ua741:
Hello,
This would break the behaviour for people who have multiple admins. Is there a specific reason why can't use the config file to list the admin ids?

Inumedia:
It doesn't alter the admins list functionality, and instead uses a different property `admin` vs `admins`.

It's purely intended as a work-around when using environment variables instead of a config file.

For context, in my environment, I prefer to primarily use environment variables and not have config files laying around. Most of the environment variables work **except for this one**.

Maybe the pull request should be renamed to clarify

mnvr (approved):
Not very keen on bloating up the configuration with one-off workarounds, esp for cases which could be worked around in other ways (e.g. by using the first-user-as-admin functionality). Since you've already put in the effort, will merge this, but going forward please keep feature requests to discussions instead of direct code changes.

(will add the documentation for this new option in a separate PR)

PR #4235, opened 2024-11-28
[cli][setup] add: public-albums-url subcommand to configure albums url

mngshm:
Adds a command `ente setup public-albums-url`. 

Capture-Nov-28-17-12-36


ua741:
Should we maintain a separate branch for all these changes related to self-hosting setup. We can merge the changes in the main when it's in a state where we can release it to cater almost all of the initial use-cases that we can think of?

mngshm:
Yes, please go ahead. I will work on and push majority of my changes there.

PR #4238, opened 2024-11-28
Add `internal.disable-registration` to config

neferin12:
## Description
As discussed in #2476, I added an option `internal.disable-registration` to allow disabling the registration of new users on selfhosted instances.
Users can still go through the registration flow, but when entering the ott they received via mail, they get an unauthorized error.

Should this be documented in any of the self-hosting documentation?

### DIFF
--- a/server/pkg/controller/user/userauth.go
+++ b/server/pkg/controller/user/userauth.go
@@ -362,13 +362,13 @@ func (c *UserController) onVerificationSuccess(context *gin.Context, email strin
 	userID, err := c.UserRepo.GetUserIDWithEmail(email)
 	if err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
-			if !viper.GetBool("internal.disable-registration") {
+			if viper.GetBool("internal.disable-registration") {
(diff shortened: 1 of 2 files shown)


server/pkg/controller/user/userauth.go
```diff
@@ -362,9 +362,13 @@ func (c *UserController) onVerificationSuccess(context *gin.Context, email strin
 	userID, err := c.UserRepo.GetUserIDWithEmail(email)
 	if err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
-			userID, _, err = c.createUser(email, source)
-			if err != nil {
-				return ente.EmailAuthorizationResponse{}, stacktrace.Propagate(err, "")
+			if !viper.GetBool("internal.disable-registration") {
```
mnvr: Let us swap the conditions to avoid the double negation
```
if viper.GetBool("internal.disable-registration") {
 ... err
} else {
  ... normal
```

mnvr (approved):
Thank you. I'll also mention this in the self-hosting channel.

mnvr:
Have also added a mention in the self-hosting docs - https://github.com/ente-io/ente/pull/4259

PR #4253, opened 2024-11-29
[mob][preview] add support for video streaming

prateekmedia:
## Description

- Add support for encrypted HLS streaming of videos using video_player(Android) and media_kit (iOS)


prateekmedia:
Hi @simondubrulle , This PR is just for the mobile end. The backend is done separate on video_file_preview branch.

Regarding the rotation part it, it was due to the source of video_player, I have reverted it to original.

Also regarding the "Cache Preview", it will work only when you spin up the local backend using the said branch.

mobile/lib/ui/viewer/file/video_view_widget.dart
```diff
@@ -0,0 +1,157 @@
...
+  Future<void> _checkForPreview() async {
+    if (!flagService.internalUser) return;
+
+    final Set<int> filesWithFDStatus =
+        await MLDataDB.instance.getFileIDsWithFDData();
+
+    if (!filesWithFDStatus.contains(widget.file.uploadedFileID)) {
+      isCheckingForPreview = false;
+      setState(() {});
+    }
```
ua741: We need to check for fileIDs that have vid_preview type, this will also return files that have ml indexes.

mobile/lib/ui/viewer/file/file_widget.dart
```diff
@@ -40,16 +40,8 @@ class FileWidget extends StatelessWidget {
...
-      // playing certain videos on iOS simulator
-      // if (kDebugMode && Platform.isIOS) {
-      //   return VideoWidget(
-      //     file,
-      //     tagPrefix: tagPrefix,
-      //     playbackCallback: playbackCallback,
-      //   );
-      // }
-      return VideoWidgetNative(
+      return VideoViewWidget(
```
ua741: We should still use the native player for non-preview videos.
Also, let's add the logic here to check if it's a internal User and the video has preview, then use the PreviewVideoWidget otherwise stick to existing logic. Also, if the video is not uploaded, we just stick to existing logic here.

Inside FileDataService, we can cache the fileIDs for all videos that have preview (that cache can be populcated in sync manner during app startup).

mobile/lib/ui/viewer/file/video_view_widget.dart
prateekmedia: Resolved!

mobile/lib/ui/viewer/file/file_widget.dart
prateekmedia: Resolved!

mobile/lib/services/preview_video_store.dart
```diff
@@ -167,31 +170,36 @@ class PreviewVideoStore {
   }
 
   String _getCacheKey(EnteFile file) {
-    return "vid_preview_${file.uploadedFileID}";
+    return "video_playlist_${file.uploadedFileID}";
+  }
+
+  String _getVideoPreviewKey(EnteFile file) {
+    return "video_preview_${file.uploadedFileID}";
   }
```
ua741: Here, the preview key can be the object_id. If we have to regenerate a preview for whatever reason, the object_id will change.

mobile/lib/models/metadata/file_magic.dart
```diff
@@ -14,6 +14,7 @@ const latKey = "lat";
 const longKey = "long";
 const motionVideoIndexKey = "mvi";
 const noThumbKey = "noThumb";
+const previewVersionKey = "previewVersion";
```
ua741: We can remove this?

prateekmedia: Yes

mobile/lib/services/preview_video_store.dart
prateekmedia: Outdated review!

mobile/lib/utils/file_uploader.dart
```diff
@@ -469,6 +472,16 @@ class FileUploader {
     }
   }
 
+  void _uploadPreview(EnteFile file) {
```
ua741, line 477: This should only be called for internal users.

prateekmedia: Resolved!

mobile/lib/utils/file_download_util.dart
```diff
@@ -267,10 +267,7 @@ Future<void> _saveLivePhotoOnDroid(
   savedAsset = (await (PhotoManager.editor.saveVideo(
     video,
     title: videoTitle,
-  )));
-  if (savedAsset == null) {
-    throw Exception("Failed to save video of live photo");
-  }
+  )).catchError((_) => throw Exception("Failed to save video of live photo")));
```
ua741, line 270: Is there a reason to add these changes?
Instead of catchError, i would probably prefer try catch for better readability.

mobile/lib/services/preview_video_store.dart
```diff
@@ -0,0 +1,525 @@
...
+    if (!isVideoStreamingEnabled) return;
+
+    final cutoff = videoStreamingCutoff;
+    if (cutoff == null) return;
+
+    final files = await FilesDB.instance.getAllFilesAfterDate(
+      fileType: FileType.video,
+      beginDate: cutoff,
+    );
+    await Future.delayed(const Duration(seconds: 5));
```
ua741: Please add comment/reason for these wait times.

mobile/lib/services/preview_video_store.dart
```diff
@@ -0,0 +1,525 @@
...
+      return response.data["url"];
+    } catch (e) {
+      _logger.warning("Failed to get preview url", e);
+      rethrow;
+    }
+  }
+
+  // get all files after cutoff date and add it to queue for preview creation
+  // only run when video streaming is enabled
+  Future<void> putFilesForPreviewCreation() async {
```
ua741, line 578: This can be triggered inside init instead of calling this method seperately. Ideally, inside init, I will trigger this method after 5 or 10 seconds to avoid blocking app start. With with ignore, this can end up taking some CPU cycle.

prateekmedia: Done!

mobile/lib/services/preview_video_store.dart
prateekmedia: Removed the wait times. Reasoning was to wait for previews to be loaded, which are loaded by the ML services.

mobile/lib/utils/file_download_util.dart
prateekmedia: It is due to the upgrade, basically now savedAsset will never be null, so I had to catch and throw exceptions myself.

mobile/lib/services/preview_video_store.dart
```diff
@@ -0,0 +1,549 @@
...
+    required int objectSize,
+  }) async {
+    _logger.info("Pushing playlist for ${file.uploadedFileID}");
+    try {
+      final encryptionKey = getFileKey(file);
+      final playlistContent = playlist.readAsStringSync();
+      final result = await gzipAndEncryptJson(
+        {
+          "playlist": playlistContent,
+          'type': 'hls_video',
```
ua741, line 372: The view preview height and size should be preserved here instead of file metadata as we might generate more previews of same file

mobile/lib/models/metadata/file_magic.dart
```diff
@@ -96,6 +107,9 @@ class PubMagicMetadata {
       mvi: map[motionVideoIndexKey],
       noThumb: map[noThumbKey],
       mediaType: map[mediaTypeKey],
+      previewWidth: map[previewWidthKey],
+      previewHeight: map[previewHeightKey],
+      previewSize: map[previewSizeKey],
```
ua741: Size is already available (including the metadata + video) in the fileData. Also, these previewHeight and Width should be stored along with the playlist as we might generate multiple previews for single video.

mobile/lib/main.dart
```diff
@@ -280,13 +286,20 @@ Future<void> _init(bool isBackground, {String via = ''}) async {
...
       entityService,
       MLDataDB.instance,
       preferences,
     );
+    unawaited(
+      Future.delayed(
+        const Duration(seconds: 10),
+        PreviewVideoStore.instance.putFilesForPreviewCreation,
+      ),
+    );
```
ua741: This can be moved inside init of PreviewVideoStore

PR #4266, opened 2024-11-30
[auth] Add "Favorites" Tag to Display Pinned Codes

AmanRajSinghMourya:
## Description
This PR introduces a new Favorites tag, allowing users to quickly view all their pinned codes in one place.

### DIFF
--- a/auth/lib/ui/home_page.dart
+++ b/auth/lib/ui/home_page.dart
@@ -79,6 +79,9 @@ class _HomePageState extends State<HomePage> {
   bool hasTrashedCodes = false;
   bool hasNonTrashedCodes = false;
   bool isCompactMode = false;
+  bool _isFavouriteOpen = false;
+  bool hasFavouriteCodes = false;
+  bool hasNonFavouriteCodes = false;
 
   @override
   void initState() {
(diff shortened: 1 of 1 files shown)


auth/lib/ui/home_page.dart
```diff
@@ -110,13 +113,24 @@ class _HomePageState extends State<HomePage> {
...
           hasTrashedCodes = true;
         } else {
           hasNonTrashedCodes = true;
         }
-        if (hasNonTrashedCodes && hasTrashedCodes) {
+        if (c.isPinned) {
+          hasFavouriteCodes = true;
+        } else {
+          hasNonFavouriteCodes = true;
+        }
```
ua741: This check we should do for non-trashed codes only. If only pinCodes are in trash, we don't want to show the tab.

auth/lib/ui/home_page.dart
```diff
@@ -110,13 +113,24 @@ class _HomePageState extends State<HomePage> {
...
+        if (hasTrashedCodes && hasNonTrashedCodes) {
+          break;
+        }
+        if (hasFavouriteCodes && hasNonFavouriteCodes) {
           break;
         }
```
ua741: This should be if (hasTrashedCodes && hasNonTrashedCodes && hasFavouriteCodes && hasNonFavouriteCodes) {
break;
}

ua741 (approved):

PR #4269, opened 2024-11-30
Dashes in ENV var names not supported by docker compose

flaki:
When I was trying to set this up, I was getting errors from `docker compose`:

```
failed to read /var/hosts/ente/.credentials.env: line 11: unexpected character "-" in variable name "ENTE_APPS_PUBLIC-ALBUMS=https://<redacted>"
```

There was, apparently, a brief period in time when `docker compose` supported dashes in ENV var names, [but this is no longer the case](https://github.com/docker/compose/issues/12302), hence some vars need to be defined in a `museum.yml` and mounted into the instance. Maybe there is some rewriting in place that allows to redefine `museum.yml` entries with dashes in ENV vars, however I couldn't find it (and if there isn't, there probably should be, but this change should make the guide usable again at least).

### DIFF
--- a/docs/docs/self-hosting/guides/external-s3.md
+++ b/docs/docs/self-hosting/guides/external-s3.md
@@ -72,6 +72,7 @@ services:
             - ./.credentials.env
         volumes:
             - custom-logs:/var/logs
+            - museum.yaml:/museum.yaml:ro
         networks:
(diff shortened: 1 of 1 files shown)


mnvr (approved):
Thank you! 🙏

PR #4303, opened 2024-12-04
[cli] adds some checks for env vars and reference to cli secrets doc.

mngshm:


cli/main.go
```diff
@@ -119,16 +119,25 @@ func initConfig(cliConfigDir string) {
...
+
+  if configDir == "" {
+    // for backward compatibility, check for ENTE_CLI_CONFIG_PATH
+    configDir = os.Getenv("ENTE_CLI_CONFIG_PATH")
+  }
+
+  // Throw instructoin if none of the environment variables has been set.
+  if configDir == "" && oldConfigPath == "" {
+    fmt.Println(`Environment variable for the config dir is not set. Please setup 
+the environment variable with export ENTE_CLI_CONFIG_DIR="/path/to/config"
```
ua741: What is the need for this log line? Will it print on each command?
Asking because for normal Ente Customer, ideally they should not need to configure anything as long as KeyRing is available.

mngshm: Yes, wanted this to be a reminder to export the ENV variable. But makes sense, totally forgot about the other part. I will add something to handle the other scenario (if keyring is available)

mngshm: https://github.com/ente-io/ente/pull/4303/commits/08357e1d7354426c0d6dc68648705e859a22549e

Pushed some changes adding a check for keyring availability. please check and merge if looks good

P.S: Mistakenly pushed with wrong commit message in the above because of reverse-i-search, amended with new message

mngshm:
Removed the function and log line as discussed

cli/main.go
```diff
@@ -119,16 +119,19 @@ func initConfig(cliConfigDir string) {
 // GetCLIConfigDir returns the path to the .ente-cli folder and creates it if it doesn't exist.
 func GetCLIConfigDir() (string, error) {
 	var configDir = os.Getenv("ENTE_CLI_CONFIG_DIR")
-	if configDir == "" {
-		// for backward compatibility, check for ENTE_CLI_CONFIG_PATH
-		configDir = os.Getenv("ENTE_CLI_CONFIG_PATH")
-	}
+  // For backward compatibility 
+  var oldConfigPath = os.Getenv("ENTE_CLI_CONFIG_PATH")
+
```
ua741: Can remove this?

PR #4348, opened 2024-12-08
Update json for custom icons

k3kk07:
## Description
Update json for custom icons for an error and add new nord account icon.


k3kk07:
i don't know why microsoft icon is not working in the last version so i update the name in the json with lowercase name.

ua741 (approved):

PR #4369, opened 2024-12-10
Fix #4296 + download UI improvement + fixes

### LINKED ISSUE #4296: [mob][Photos] CreationDate gets (randomly?) changed when downloading remote photo/video
### Description

It is now the second time in only a few trials that this behaviour occurs. 

Whenever I download remote files that are in my timeline (and not from a public shared album) but not yet on my device, the files get propely downloaded in the "Pictures" folder (assets are created) but some of the tiles will (likely) move to the top of the timeline in the "today" section. Looking at the photos info, the creation time is now today (at the time of the download). If I check the web app, it will also appear in the "today" section, like if it has been changed in the DB as well.

**Steps to reproduce :**

0. Delete application cache (to make sure the remote files are not cached somewhere)
1. From the timeline, or from an album view, Multi select files which you know are not yet on your device (and let's assume photos were taken in 2019)
2. Click "Download" from the selection tab and let it download them all
3. Some of them will "likely" have their creation time changed to today, at the time you clicked the download button
4. You will therefore find some of these downloaded photos in "today" section of the album/timeline.

**Expected behaviour :**
Since Ente was already "aware" of the correct date, it should not change it when I download the photo locally.

**Extra info :**
I also noticed that in the "Pictures" album (which is not selected for backup), some of the downloaded photos will have the proper creation date assigned (in 2019, for example) but most of them will be in the "Today" section as well. At first I thought it was only pictures with no EXIF data but I can find some with Exif data as well. Actually, some photos from the same moment (few seconds apart) with both Exif data, same filename structure,  do have both behaviour (one is in the "today" section, and the other one in the right day in 2019). 

I was downloading +- 130 photos. 

It also occured for videos.

I tried to have a look at the code, but apparently all it does is downloading the asset and update the file local_id with the one from the asset:

https://github.com/ente-io/ente/blob/7b002c857e446b3692951abc383b9c2933135691/mobile/lib/utils/file_download_util.dart#L219-L227

I am not sure of how the sync process is handling those events but it might be that some events are not processed properly.

### Version

v0.9.65

Ente Photos

Mobile - Android

simondubrulle:
### Background 
Proposal to fix #4296 (@ua741). High level, there is a race condition which is generating wrong entries in the FilesDB (creating orphan files). This issue is easily replicable if you select multiple files and try to download them (the more files, the more likely wrong entries will be created). This issue is also messing up with the date of the files we see in the timeline and is creating "orphan" (not linked to uploaded files) files in the device collections. 

To my understanding, this is quite a critical bug and any user is very likely to encounter it in the current implementation. 

There is some self healing already in place that will reupload all those files and this should correct the dates, but this will only do it 10 files at a time, and based on potential candidates, without fixing the orphan entries. I was also able to produce files which were never corrected.

### This pull request
After extensive tests and lots of debugging, I was able to fully fix this issue (meaning I am not able to reproduce it, at all, and I tried a lot..). I also do not see any orphan in the device collection ("Pictures" folder for Android), while i could see 20+ (even more) each time I was trying to replicate the issue. To fix it, I have:

- created a new class `PhotoManagerSafe` which is safely making sure that the app will not react to file changes when it has been asked not to do so (previous implementation allowed scenarios where a caller was executing code thinking no change notification would be sent).
- created a Lock in the `LocalSyncService` that is used to make sure no local synchronisation can be performed at the same time a download is performed (this was the main culprit). The main side effect is that when you download multiple files, they will be processed sequentially. I think this can be improved but the main "side" advantage is that whenever there is a failure, not a lot will be lost and it is more robust in that way if you want to download a big collection. It would be better to have at least 2 threads downloading/decrypting at the same time, maybe (maybe not needed as well)?
- improved the "Downloading files..." dialog, which is now showing the progress (for example "Downloading files... (235/494)"), so that the user knows if something is actually happening and what is the current progress.
- Added a missing call on the Video Editor page where the change notify was stopped but never started again at the end when task was performed.

Feel free to close this PR if not suitable according to you, I can also take feedback and try to implement them as best as I can.

## Tests

Tested on my Pixel 6a

Built with :

Flutter 3.24.3
JDK 17.0.2
Gradle 7.2


mobile/lib/utils/file_download_util.dart
```diff
@@ -189,50 +190,54 @@ Future<void> downloadToGallery(EnteFile file) async {
...
-    } else if (type == FileType.livePhoto) {
-      final File? liveVideoFile =
-          await getFileFromServer(file, liveVideo: true);
-      if (liveVideoFile == null) {
-        throw AssertionError("Live video can not be null");
+    // We use a lock to prevent synchronisation to occur while it is downloading
+    // as this introduces wrong entry in FilesDB due to race condition
+    // This is a fix for https://github.com/ente-io/ente/issues/4296
+    await LocalSyncService.instance.getLock().synchronized(() async {
+      final File? fileToSave = await getFile(file);
```
ua741: We can move this outside. This will fetch the file from remote, and decrypt it.

mobile/lib/utils/photo_manager_util.dart
```diff
@@ -10,3 +13,128 @@ Future<PermissionState> requestPhotoMangerPermissions() {
...
 }
+
+final _logger = Logger("PhotoManagerUtil");
+// This is a wrapper for safe handling of PhotoManager.startChangeNotify() and
+// PhotoManager.stopChangeNotify(). Since PhotoManager is globally shared, we want
+// to make sure no notification is sent while it should not. The logic is that it will
+// only start if no other asset (or task) requested to stop changes, or if the expiration
+// time for the asset (task) expired. '_processingAssets' should be seen as a queue of
+// open requests.
+class PhotoManagerSafe {
```
ua741: Sorry, but to me this feels like additional complexity that we like prefer to avoid adding.
Is it possible to simplify this part either via debouncer or leveraging the fact that all calls to PhotoManager start/stop happens on the same thread?

simondubrulle: I agree, actually it is not even needed with the Local Service Lock (I tested it, but it was already making sense). I have removed it at the moment as it is trying to solve a bigger issue IMO.

mobile/lib/utils/file_download_util.dart
simondubrulle: I tried moving it outside, and it is technically working (it is not triggering this specific issue), but it starts all selected downloads at same time with following effects:
- app gets very laggy
- likely to have network issues (I have had time outs and network errors because of too many parallel/open requests). Issues which I did not get with the sequential approach.
- it should be faster but in my opinion less robust when dowloading lots of files/complete collection.

At the moment I kept it inside as it is solving the issue while not introducing other ones but I leave it up to you to decide based on this feedback.

A further improvement might be to move it outside and limit the number of concurrent downloads, but maybe not part of this fix/PR?

ua741: Ack. May be we can use different lock to restrict one download. This will reduce the potential time period during which sync might be blocked while files are getting downloaded.
```dart
await downloadLock.synchronized(() async {
final File? fileToSave = await getFile(file);
});
```

PS: We can keep the previous change to surface dialog progress on UI.

simondubrulle: I tried multiple other approaches, including the one you suggested and at the end the following was the simplest and with the best results (to my opinion) : 
https://github.com/ente-io/ente/pull/4369/commits/32ed84f48da01d706f59fc70b88cadd3b9d144c9
- moving getFile outside the lock as you initially suggested
- created simple queue for downloads (max 5 concurrent downloads)

Concurrent downloads does make it significantly faster (hence my attempt with the queue)

ua741 (approved):

ua741:
Once again, thank you for your patience and contribution @simondubrulle  :)

PR #4384, opened 2024-12-12
[auth] Fixing taskbar overlapping window panel

FluffyPunk:
## Description

Pull request was made because of some user percentage has taskbar connected on top, so it overlaps auth window for WinApp at start.
After some measuring, I found that taskbar occupies:

- 50px on FHD with 125% scaling (standard values on some notebooks)
- 40px on FHD with 100% scaling
- 30px on FHD with 100% scaling and small taskbar icons

### DIFF
--- a/auth/windows/runner/main.cpp
+++ b/auth/windows/runner/main.cpp
@@ -72,7 +72,7 @@ int APIENTRY wWinMain(_In_ HINSTANCE instance, _In_opt_ HINSTANCE prev,
   project.set_dart_entrypoint_arguments(std::move(command_line_arguments));
 
   FlutterWindow window(project);
-  Win32Window::Point origin(50, 50);
+  Win32Window::Point origin(70, 70);
   Win32Window::Size size(1280, 720);
   if (!window.Create(L"Ente Auth", origin, size))
   {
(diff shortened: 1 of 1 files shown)


ua741:
Hey @FluffyPunk,
Thank you for creating this PR :).
Can you please attach screenshots so that we can better understand the effect of this change.

FluffyPunk:
Yeah, sorry.
This one was FHD with 100% scaling and small taskbar. According to ShareX, it goes about 8px for window top panel, so it is barely draggable
image 

And this was FHD with big taskbar, which fully overlaps window panel and it can't be dragged or resized except doing it by keyboard shortcuts

image

According to this change, it should open window on 70, 70(i made it bigger to be sure of all sizes) and it doesn't go under taskbar on start.
That's the only point i saw about creating window at 10,10 coords

ua741 (approved):

chrissv:
Just changing the windows offsets is no better than the original implementation. It is very short-sighted to make an assumption as to how people have configured the task bar and the location. With this change, my task bar still overlaps part of the window (albeit less than before).

A better implementation would be to actually detect the size and adjust dynamically. Or just don't hard-code the window location in the first place.

PR #4395, opened 2024-12-13
[auth] Add Support for Custom Icons 

AmanRajSinghMourya:
## Description
This PR introduces a new feature allowing users to select and assign custom icons to their codes.

### DIFF
--- a/auth/lib/models/code_display.dart
+++ b/auth/lib/models/code_display.dart
@@ -1,6 +1,6 @@
 import 'dart:convert';
 
-import 'package:ente_auth/ui/utils/icon_utils.dart';
+import 'package:ente_auth/models/all_icon_data.dart';
 import 'package:flutter/foundation.dart';
 import 'package:logging/logging.dart';
 
--- a/auth/lib/models/code_display.dart
+++ b/auth/lib/models/code_display.dart
@@ -1,5 +1,6 @@
 import 'dart:convert';
 
+import 'package:ente_auth/ui/utils/icon_utils.dart';
 import 'package:flutter/foundation.dart';
 import 'package:logging/logging.dart';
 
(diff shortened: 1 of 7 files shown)


auth/lib/ui/utils/icon_utils.dart
```diff
@@ -162,3 +215,19 @@ class CustomIconData {
...
+  final String? color;
+  final String? slug;
+
+  AllIconData({
+    required this.title,
+    required this.type,
+    required this.color,
+    this.slug,
+  });
+}
```
ua741: let's define these types under models?

auth/lib/models/code_display.dart
```diff
@@ -12,6 +13,9 @@ class CodeDisplay {
   String note;
   final List<String> tags;
   int position;
+  String customIconData;
+  bool isCustomIcon;
+  IconType iconType;
```
ua741: Instead of defining type, let's just use src
String iconSrc;
String iconID;

isCustomIcon can just be a getter => iconSrc != '' && iconID |=''l

auth/lib/models/code_display.dart
```diff
@@ -21,8 +23,12 @@ class CodeDisplay {
     this.tags = const [],
     this.note = '',
     this.position = 0,
+    this.iconSrc = 'ente',
```
ua741: the default value can be empty? Any reason to prefer 'ente' over empty?

AmanRajSinghMourya: Yes so when add a new code and then go to edit it, there we show 2 icons
1. ente icon 
2. icon based on issuer name 
so when there is new code added we need to show some icon for 1. so put ente icon as default, took reference from 2FAS
image

ua741 (approved):

PR #4458, opened 2024-12-21
[server] Return various app urls as part of relevant API response.

ua741:
## Description

For the cast dialog, that we show on the app, before making the API call to pair. Have parked that change for now.

### DIFF
--- a/server/pkg/api/user.go
+++ b/server/pkg/api/user.go
@@ -7,6 +7,7 @@ import (
 	"github.com/ente-io/museum/pkg/controller/emergency"
 	"github.com/gin-contrib/requestid"
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
 	"net/http"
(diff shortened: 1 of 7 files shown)


server/pkg/api/user.go
```diff
@@ -485,6 +487,7 @@ func (h *UserHandler) GetAccountsToken(c *gin.Context) {
 	}
 	c.JSON(http.StatusOK, gin.H{
 		"accountsToken": token,
+		"accountUrl":    viper.GetString("apps.account"),
```
mnvr: Typo, I think, the config key is "apps.accounts" (s). While we're at it, can we also change the JSON accountUrl => accountsUrl - no functional impact, just for consistency (familyUrl is fine, don't need to change that, just changing  accounts would be good since the rest of the PR uses the plural)

mnvr (approved):

PR #4488, opened 2024-12-23
World Cube Association icon

Alf3rez:
Hi, I just want to contribute adding the World Cube Association official logo. I hope every thing is ok, this is my first pull request ever and I don't want to mess anything.

Thank you.


Alf3rez:
Sorry, if everything is ok with this, before merging wait for my confirmation with World Cube Association marketing members, because maybe this use is outside is own regulation, I assume it isn't, but after reading both legal sides from you and from them I believe is better to have an explicit "yes" from them to be completely sure.

Sorry again and thank you for your patience.

ua741:
I will close this to reduce the noise from open PR. Feel free to create another pull request when it's ready for review.

Issue #2003, opened 2024-06-04
[Auth] Constant 20-25% CPU usage on macOS


DrBrynzo:
### Description

Whether the window is hidden or visible, Ente Auth is always at between 20% to 25% CPU usage.
![Screenshot 2024-06-04 at 11 07 51 AM](https://github.com/ente-io/ente/assets/37386883/15e56fcd-ca2f-4dd7-b9b9-b4fe3e508eed)

### Version

3.0.6

### What product are you using?

Ente Auth

### What platform are you using?

Desktop - macOS

m-ammar:
Hello. I too face the same issue. However, on my M1 Macbook pro 16", it is consistently between 45-55%! 
<img width="455" alt="image" src="https://github.com/ente-io/ente/assets/34128636/580f788e-efca-4280-8177-c0d11dae2b1f">

I use Ente Auth Desktop app on macOS.
Apple M1 Pro
macOS Sonoma Version 14.5 (23F79)

Fredouye:
Same here : macOS 14.6.1, M1 Pro 10 cores, and Ente Auth 3.1.3 is constantly using 50-55% of CPU.

Moulick:
Same!

Fredouye:
I still have the issue with macOS 15.0 and Ente Auth 4.0.0

prateekmedia:
Related to Flutter:
https://github.com/flutter/flutter/issues/59327

pshifrin:
Same on macOS 15.0.1 and Ente Auth 4.0.2 and M3 MBP

OscarVanL:
Same issue here, I have a M4 Max processor, ente is minimised and still using ~40% CPU.

<img width="562" alt="image" src="https://github.com/user-attachments/assets/7b78acb4-a5cd-464e-b235-f854ad22055f">

simonryf:
I understand - this is open source and free and everything. And besides CPU load - it IS pretty nice :-) And IMO the only option if you want desktop app since the EOL of Authy. So I cannot throw banana skins in your garden.

Having said that. The solution is IMO relatively easy: Don't generate the codes when the window is not visible.

As a cumbersome workaround you can **search for nonexistent string so no codes are displayed and CPU drops to 0**.

I don't know how many mac-users of ente auth we are. But I bet if you combine all the wasted power - you could probably mine a few Bitcoins a day for it.

So I'm offering (pretty good) chocolate for a fix :-P

ua741:
[...]

PR #4517, opened 2024-12-27
[auth] Reduce progress bar refresh rate to lower CPU usage

ua741:
## Description
Related https://github.com/ente-io/ente/issues/2003#issuecomment-2563380828


OscarVanL:
Thank you! I just installed the new updated version of Ente with this patch and on my M4 Max Macbook Pro, Ente's CPU usage has dropped to ~6% while minimised and 7-10% while focused. Before it was ~40% (see [here](https://github.com/ente-io/ente/issues/2003#issuecomment-2465944451)).

This still feels a little higher than it should be (for example, my current Chrome window is using 2-5% CPU with ~30 several tabs open, while Ente sitting minimised in the background is using more), but it's a big improvement, so thank you for this!

PR #4565, opened 2025-01-01
[docs] Update self-hosting docs for configurable app urls

mngshm:


mngshm:
I was a bit unsure of what worked and what not, so I built the image from scratch one by one myself to see which env vars work and which don't. Hopefully ready to merge now.

ua741 (approved):

Issue #4536, opened 2024-12-29
enteauth deleting the entire ~/.cache folder on linux


gepbird:
### Description

A few minutes after running enteauth, I noticed other programs misbehaving, failing to read files in ~/.cache, and I was surprised to see my cache directory being pretty much empty (~300KB) when it's usually hundreds of MB. This happened for the 3rd time.

Unfortunately I can't consistently reproduce this by opening and closing enteauth (maybe it does it only after a specific amount of time has passed?).

By peeking at the code, looks like it's [deleting `tempDirectory`](https://github.com/ente-io/ente/blob/auth-v4.1.6/auth/lib/core/configuration.dart#L84) that is [being set to `cacheHome`](https://github.com/ente-io/ente/blob/auth-v4.1.6/auth/lib/utils/directory_utils.dart#L30) which [resolves to $XDG_CACHE_HOME](https://pub.dev/packages/xdg_directories) that is `~/.cache`.

It should only delete enteauth's data, something like `~/.cache/enteauth`.

Looks like this bug was introduced in 2e01a964775 (https://github.com/ente-io/ente/pull/3334), cc @prateekmedia

Related: https://github.com/ente-io/ente/issues/2563

### Version

4.1.6

### What product are you using?

Ente Auth

### What platform are you using?

Desktop - Linux

prateekmedia:
Will be fixed in next release.

stancel:
Wow, thank you for opening this issue as well as for the upcoming fix! I have been going nuts trying to track down why my ~/.cache folder contents are being periodically wiped out. It is killing the ibus socket files stored there that eventually causes the keyboard input not to work, among many other issues. I have been searching through system logs and putting scripts in place on cron jobs that backup my ~/.cache directory and then restore it has been wiped. The timestamps of this logging I put in place of this happening matches my times using and closing out enteauth. I was also able to just recreate the issue, confirming it is wiping out all contents under ~/.cache/*. Thank you for finding and diagnosing this issue and for the upcoming fix. Will check for an update now. Was on 4.1.6 when adding this comment and testing the issue.

stancel:
[...]

Issue #4464, opened 2024-12-21
[Ubuntu] Keyboard is locked from using after launching ente on some apps


omer-bar:
### Description

So, I noticed a weird bug today—if I launch Ente on my Ubuntu machine, I cannot type in some apps anymore(Terminal, Firefox, etc.) but in some apps, I can.
Even after closing the app completely, I need to log out in order to regain the ability to type in Terminal/Firefox. Now, if the apps were already open, everything would be fine, but once I close them and reopen them, I can't type at all.

A video of the problem:
https://streamable.com/c9dkn1

### Version

v4.1.6

### What product are you using?

Ente Auth

### What platform are you using?

Desktop - Linux

dburba:
The same for me. I used flatpak before noticing this. After removing flatpak and installing deb no changes, the bug still persists.

stancel:
The same behavior was happening for me. I was desperately trying to find the root cause and then noticed that another user here correctly identified that the contents under the ~/.cache folder in Linux at least Ubuntu 24.04 were being periodically deleted when EnteAuth was opened. You will find more details in issue  #4536. I have also placed a shell script there that you can use to triage the issue until a permanent fix is pushed out in a future release of EnteAuth. Hope that helps.

PR #4597, opened 2025-01-05
[auth] Fix ~/.cache being deleted on linux

gepbird:
## Description

Fixes https://github.com/ente-io/ente/issues/4536 and https://github.com/ente-io/ente/issues/4464.

## Tests

After opening the app on linux, my 1.3G cache folder remained intact.

## Technical details

Now the path resolves to `~/.cache/io.ente.auth` instead of `~/.cache`. [Under the hood](https://github.com/flutter/packages/blob/ca4671cc0f2d0a347db662144978f06ae9c017e8/packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart#L75-L82) it uses the app's ID and the `xdg_directories` package, so it will work when the user has `XDG_CACHE_HOME` set to something else.

### DIFF
--- a/auth/lib/utils/directory_utils.dart
+++ b/auth/lib/utils/directory_utils.dart
@@ -27,9 +27,9 @@ class DirectoryUtils {
     Directory? directory;
     if (Platform.isLinux) {
       try {
-        return cacheHome;
+        return await getApplicationCacheDirectory();
       } catch (e) {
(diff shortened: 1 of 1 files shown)


gepbird:
cc @prateekmedia

prateekmedia (approved):
LGTM! Thanks.

stancel:
Thanks to both of you! Much appreciated!

PR #4677, opened 2025-01-11
[server] Make SMTP sender name configurable

tobiasge:

### DIFF
--- a/server/configurations/local.yaml
+++ b/server/configurations/local.yaml
@@ -228,6 +228,9 @@ smtp:
     # The email address from which to send the email. Set this to an email
     # address whose credentials you're providing.
     email:
+    # Optional override for the sender name in the emails. If specified, it will
+    # be used for all emails sent by the instance (default is email specific).
+    sender-name:
 
 # Zoho Zeptomail config (optional)
 #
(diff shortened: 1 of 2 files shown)


server/pkg/utils/email/email.go
```diff
@@ -62,10 +63,14 @@ func sendViaSMTP(toEmails []string, fromName string, fromEmail string, subject s
 		emailAddresses += email
 	}
 
-	// If an sender email is provided use it instead of the fromEmail.
+	// If a sender email is provided use it instead of the fromEmail.
 	if smtpEmail != "" {
 		fromEmail = smtpEmail
 	}
+	// If a sender name is provided use it instead of the fromName.
+	if smtpEmail != "" {
```
ua741: `smtpSender != nil` ?

tobiasge: Copy & paste error; is now fixed.

server/configurations/local.yaml
```diff
@@ -228,6 +228,7 @@ smtp:
     # The email address from which to send the email. Set this to an email
     # address whose credentials you're providing.
     email:
+    sender:
```
mnvr: Can we rename this to sender-name, and add a comment, e.g.
```
# Optional override for the sender name in the emails. If specified, it will
# be used for all emails sent by the instance (default is email specific).
sender-name:
```

Otherwise looks good to merge.

mnvr: EDIT: Update casing to "sender-name"

PR #4727, opened 2025-01-14
[server] update request_logger.go to ignore /ping

11notes:
Don’t spam the log with messages about requests on /ping

## Description
The logger should simply ignore requests on /ping to not spam the log every {n} seconds for this trivial test. If this is somehow required maybe consider ignoring /ping if request was made by localhost.

### DIFF
--- a/server/pkg/middleware/request_logger.go
+++ b/server/pkg/middleware/request_logger.go
@@ -47,6 +47,13 @@ func shouldSkipBodyLog(method string, path string) bool {
 	return false
 }
 
+func shouldSkipLog(method string, path string) bool {
+	if method == "GET" && path == "/ping" {
+		return true
+	}
+	return false
+}
+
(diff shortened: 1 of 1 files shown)


11notes:
With a [health check every 5 seconds](https://github.com/11notes/docker-ente/blob/32d6e574c2b52e226e4a6fd7f1b2b2e547ba542e/arch.dockerfile#L84) the log is basically full of SPAM since the ente local.yaml does not support  a LOGLEVEL (or at least I’m not aware of it?).

mnvr:
Thank you for the pull request, but as mentioned in the contributing guide, for anything beyond trivial bug fixes please create a feature request or discussion instead of performing code changes directly. 

That way, we can gauge community interest and get feedback first before adding more code and branching logic that we don’t really need and doesn’t unblock a common use case of self hosting.

Issue #2326, opened 2024-06-30
On Linux, ente auth fails to run randomly. Issue with both app images and debian.


asheshstha:
### Description

I was on `ente-auth-v3.0.13-x86_64.deb` and suddlenly ente_auth stopped working. Tried running updated version v3.0.17 of .deb, AppImage. None of it worked.

# Issues

1. On .deb files from terminal ente_auth
``` bash
flutter:  [ente_logging] [INFO] [2024-06-30 10:04:29.358095] log file for today: File: '/home/your-user/.local/share/ente_auth/logs/2024-6-30.txt' with prefix 
flutter: [ente_logging] [INFO] [2024-06-30 10:04:29.358095] app version: '3.0.17+317'
flutter: [ente_logging] [INFO] [2024-06-30 10:04:29.358095] 
flutter:  [main] [INFO] [2024-06-30 10:04:29.358149] Starting app in foreground
flutter:  [Configuration] [INFO] [2024-06-30 10:04:29.363663] Skipping temp folder clear
Segmentation fault (core dumped)
```

2. export LIBGL_ALWAYS_SOFTWARE=1 also didn't work for .deb and .AppImages
3. running AppImages - opening onces opens the app in process but shows no UI. Need to open it twice and the UI appears as `no gl implementation is available ente`

# Solution that worked for me

Delete /home/your-user/.local/share/ente_auth/ folder which deletes the stored credentials also. Don't do it if you are not signed in with an account. Wait for better solution.

### Version

v3.0.17

### What product are you using?

Ente Auth

### What platform are you using?

Desktop - Linux

cbmiller2610:
I also see the same issue on openSUSE tumbleweed (snapshot 20240621) and Ente Auth 3.0.17+317 Appimage, randomly you will have to delete the credentials db or else the app will segfault every time on launch. More info on the segfault from journalctl:

```
Jul 03 23:35:03 localhost.localdomain kernel: ente_auth[11391]: segfault at 7f4e6dd1fddc ip 00007f4e6dc8c94e sp 00007fff6904f560 error 7 in libcairo.so.2.11800.0[7f4e6dc30000+e3000] >
Jul 03 23:35:03 localhost.localdomain kernel: Code: 47 30 02 75 5f 44 8b 66 1c 45 85 e4 75 5e f6 46 30 02 75 50 48 8b 06 48 85 c0 74 06 8b 00 85 c0 74 2a 41 bc 0d 00 00 00 31 c0 <f0>>
[...]

Issue #2004, opened 2024-06-04
Invalid windowHeight causes SIGSEGV on Linux


robxu9:
### Description

Ente Auth 3.0.7 on Linux throws a segfault when there's an invalid `windowHeight` in `shared_preferences.json`:

```json
{"flutter.windowWidth":450.0,"flutter.windowHeight":165207184.0}
```

I get a `windowHeight` like that a few times after opening Ente Auth, without doing anything else. (Just opening Ente Auth and closing it.)

Stacktrace:

```
                Stack trace of thread 788112:
                #0  0x00007d29381b7402 cairo_surface_unmap_image (libcairo.so.2 + 0x6a402)
                #1  0x00007d2939aba8aa n/a (libgdk-3.so.0 + 0x328aa)
                #2  0x00007d2939ac53d2 n/a (libgdk-3.so.0 + 0x3d3d2)
                #3  0x00007d2939ac54f3 gdk_window_end_draw_frame (libgdk-3.so.0 + 0x3d4f3)
                #4  0x00007d29385e661f gtk_main_do_event (libgtk-3.so.0 + 0x1e661f)
                #5  0x00007d2939aaf3c7 n/a (libgdk-3.so.0 + 0x273c7)
                #6  0x00007d2939ac1dc2 n/a (libgdk-3.so.0 + 0x39dc2)
                #7  0x00007d2939ac65df n/a (libgdk-3.so.0 + 0x3e5df)
                #8  0x00007d2939ac67fa n/a (libgdk-3.so.0 + 0x3e7fa)
                #9  0x00007d2937f53732 n/a (libgobject-2.0.so.0 + 0x32732)
```

Another project that had the issue, plus their solution: https://github.com/AppFlowy-IO/AppFlowy/issues/4989

### Version

v3.0.7

### What product are you using?

Ente Auth

### What platform are you using?

Desktop - Linux

JmaJeremy:
I encountered the exact same issue, and was bashing my head against a wall trying to figure out why it kept throwing a segmentation fault. Even deleting and reinstalling would fix it because that settings file was being retained. 
This is happening in version 3.0.17 for me and I'm running Manjaro Linux.

neurolit:
Same problem with Ente Auth 4.0.2 on Linux:

```"flutter.windowHeight":76456912.0```

Replacing it by 500.0 solves the problem (no more segmentation fault).

cbmiller2610:
[...]

PR #4731, opened 2025-01-15
[auth] Potential fix for segfault on Linux due to bad window size

ua741:
## Description
Related 
https://github.com/ente-io/ente/issues/2004
https://github.com/ente-io/ente/issues/2326


BrianG61UK (commented):
@ua741 Is this change why Auth now opens with a ridiculously wide window in Windows 11?
I'd like it back the way it was before, where it remembered how wide I like to have it.
I'm staying on v4.2.7 for now as a workaround.

PR #4757, opened 2025-01-17
[desktop] Add MetaInfo

NN708:
## Description

Add AppStream MetaInfo for https://github.com/flathub/flathub/pull/6032, similar to [enteauth.appdata.xml](https://github.com/ente-io/ente/blob/main/auth/linux/packaging/enteauth.appdata.xml).


mnvr (approved):
Thank you!

PR #4800, opened 2025-01-21
web-app docs update

gsabater:
## Description
Updated web-app docs to fix the pm2 ecosystem json and include a yarn version aclaration.


mnvr:
Thank you for telling us which file had the issue. We'll update the docs.

PR #4918, opened 2025-01-30
Add Twitter icon and update custom-icons.json

bessx:
Add twitter icon with X and X.com as alt names.


ua741:
Sorry, we will have to close the PR as it doesn't make sense to keep the old bird icon given the branding change by the company.

PR #4922, opened 2025-01-31
[mob][photos] Ask "open public album link in app" permission once

AmanRajSinghMourya:
## Description
1. Don't show the permission dialog if self-hosted ente
2. Show dialog only once

### DIFF
--- a/mobile/lib/ui/tabs/home_widget.dart
+++ b/mobile/lib/ui/tabs/home_widget.dart
@@ -246,7 +246,8 @@ class _HomeWidgetState extends State<HomeWidget> {
 
     if (Platform.isAndroid &&
         !localSettings.hasConfiguredInAppLinkPermissions() &&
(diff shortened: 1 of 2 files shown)


mobile/lib/ui/tabs/home_widget.dart
```diff
@@ -246,7 +246,8 @@ class _HomeWidgetState extends State<HomeWidget> {
 
     if (Platform.isAndroid &&
         !localSettings.hasConfiguredInAppLinkPermissions() &&
-        RemoteSyncService.instance.isFirstRemoteSyncDone()) {
+        RemoteSyncService.instance.isFirstRemoteSyncDone() &&
+        Configuration.instance.getHttpEndpoint() == "https://api.ente.io") {
```
ashilkn: Better to use `kDefaultProductionEndpoint` instead from `constants.dart`.

ashilkn (approved):

PR #4956, opened 2025-02-03
Add support for Takeout's new .supplemental-metadata.json files

### LINKED ISSUE #4953: Google Takeout photo import doesn't read the new ".supplemental-metadata.json" files
### Description

At some point recently, Google Takeout changed the naming for the photo metadata JSON files. Instead of just appending ".json" to the end of the photo/video filename, they now append ".supplemental-metadata.json". Ente's import doesn't handle these new files yet.

The new filenames don't seem to be used all of the time. I requested a few Takeouts (different albums in each Takeout request) two weeks ago, and some of the requests produced zip files with the old naming whereas others produced zip files with the new names. There are reports online of these new filenames going back to October 2024 ([1], [2], [3]), though I didn't see them showing up in my Takeouts until this month.

[1]: https://github.com/TheLastGimbus/GooglePhotosTakeoutHelper/issues/353
[2]: https://exiftool.org/forum/index.php?topic=16634.0
[3]: https://github.com/simulot/immich-go/issues/652

There are some quirks around how these filenames are clipped when they're longer than 46 characters. I'll put up a PR shortly that has more details on the quirks and example filenames.

### Version

v1.7.8

Ente Photos

Desktop - Linux

garyp:
In recent Google Takeout archives, the metadata JSON files are named
"${original_filename}.supplemental-metadata.json" instead of
"${original_filename}.json", as before.

I refactored the previous code so that `getMetadataJSONMapKeyForJSON()`
only removes the ".json" suffix from the metadata filename and does not
make any other changes. All of the filename munging is now done to the
name of the media file. That was the only way I could make the process
deterministic. As far as I can figure out, there's no deterministic way
of deriving the media filename from the metadata filename -- it's only
deterministic going from the media filename to the metadata filename.

These new names are still subject to the 46-character clipping limit,
with some specific rules about how the filename is clipped:

- The ".json" suffix is never clipped, only the ".supplemental-metadata"
  portion is.
- If the original filename is longer than 46 characters, then the
  ".supplemental-metadata" suffix gets completely removed during the
  clipping, along with a portion of the original filename (as before).
- The numbered suffix (if present) is also never clipped. It is however
  added at the end of the clipped ".supplemental-metadata" portion,
  instead of after the original filename. E.g. "IMG_1234(1).jpg" would
  previously use a metadata filename of "IMG_1234.jpg(1).json". Now it
  uses a metadata filename of
  "IMG_1234.jpg.supplemental-metadata(1).json". But if the filename is
  too long, it gets turned into something like
  "IMG_1234.jpg.suppl(1).json".
- Worth noting is that if the original filename is 45 characters long,
  then everything except for the "." from ".supplemental-metadata" will
  get clipped. So the metadata file ends up with a filename like
  "filename_that_is_45_chars_long.jpg..json".

I added a bunch of additional test cases in `upload.test.ts` based on
actual filenames I have in my Google Photos Takeout archives. The new
code passes all of the new test cases, as well as the original ones.

Fixes #4953

### DIFF
--- a/web/apps/photos/src/services/upload/takeout.ts
+++ b/web/apps/photos/src/services/upload/takeout.ts
@@ -23,65 +23,81 @@ export interface ParsedMetadataJSON {
     description?: string;
 }
 
+export interface FileNameComponents {
+    originalName: string;
+    numberedSuffix: string;
+    extension: string;
+    isEditedFile: boolean;
+}
+
 export const MAX_FILE_NAME_LENGTH_GOOGLE_EXPORT = 46;
+const EDITED_FILE_SUFFIX = "-edited";
+const METADATA_SUFFIX = ".supplemental-metadata";
 
 export const getMetadataJSONMapKeyForJSON = (
     collectionID: number,
     jsonFileName: string,
 ) => {
-    let title = jsonFileName.slice(0, -1 * ".json".length);
-    const endsWithNumberedSuffixWithBrackets = /\(\d+\)$/.exec(title);
-    if (endsWithNumberedSuffixWithBrackets) {
-        title = title.slice(
-            0,
-            -1 * endsWithNumberedSuffixWithBrackets[0].length,
-        );
-        const [name, extension] = nameAndExtension(title);
-        return `${collectionID}-${name}${endsWithNumberedSuffixWithBrackets[0]}.${extension}`;
-    }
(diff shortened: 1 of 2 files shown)


jfly:
Is this redundant with https://github.com/simulot/immich-go/pull/657?

mnvr (approved):
Thank you

PR #4961, opened 2025-02-04
[server] Allow setting storageLimit for family members

mngshm:


server/ente/family.go
```diff
@@ -18,7 +18,8 @@ const (
 )
 
 type InviteMemberRequest struct {
-	Email string `json:"email" binding:"required"`
+	Email        string `json:"email" binding:"required"`
+	StorageLimit int64  `json:"storageLimit" binding:"omitempty"`
```
ua741: This should be `*int64`. When the value is null, then only we should let the family member take as such space the the family usage allows.

server/pkg/controller/family/admin.go
```diff
@@ -118,6 +118,22 @@ func (c *Controller) InviteMember(ctx *gin.Context, adminUserID int64, email str
...
+		if err != nil {
+			return stacktrace.Propagate(err, "failed to Set Storage Limit")
+		}
+	} else {
+		var defaultLimit int64 = 100 * 1024 * 1024 * 1024 * 1024
+		err := c.FamilyRepo.SetStorageLimit(ctx, defaultLimit, activeInviteToken)
+		if err != nil {
+			stacktrace.Propagate(err, "Cannot Set Default Storage Limit")
+		}
+	}
```
ua741: This method should accept storageLimit *int64 as input.
And we can just modify the existing invite method to add storageLimit during invite..

```go
// AddMemberInvite inserts a family invitation entry for this given pair of admin & member and return the active inviteToken
// which can be used to accept the invite
func (repo *FamilyRepository) AddMemberInvite(ctx context.Context, adminID int64, memberID int64, inviteToken string, storeLimit *int64) (string, error) {
	if adminID == memberID {
		return "", stacktrace.Propagate(errors.New("memberID and adminID can not be same"), "")
	}
	// on conflict, we should not change the status from 'ACCEPTED' to `INVITED`.
	// Also, the token should not be updated if the user is already in `INVITED` state.
	_, err := repo.DB.ExecContext(ctx, `INSERT INTO families(id, admin_id, member_id, status, token, storage) 
			VALUES($1, $2, $3, $4, $5, $6) ON CONFLICT (admin_id,member_id) 
			    DO UPDATE SET(status, token) = ($4, $5, $6) WHERE  NOT (families.status = ANY($6))`,
		uuid.New(), adminID, memberID, ente.INVITED, inviteToken, pq.Array([]ente.MemberStatus{ente.INVITED, ente.ACCEPTED}), storeLimit)
		.....
}
```

server/pkg/controller/usage.go
```diff
@@ -103,5 +104,24 @@ func (c *UsageController) CanUploadFile(ctx context.Context, userID int64, size
...
 	}
+
+	// Get particular member's storage and check if the file size is larger than the size of the storage allocated
+	// to the Member and fail if its too large.
+	var memberStorage *int64
+	if subscriptionAdminID != userID {
+		familyMembers, err := c.FamilyRepo.GetMembersWithStatus(*familyAdminID, repo.ActiveFamilyMemberStatus)
+		if err != nil {
+			stacktrace.Propagate(err, "Failed to Get Family Members")
+		}
```
ua741: We are already fetching membersWithStatus, we should fill the value of `memberStorage` after fetching the members instead of fetching it again.

server/pkg/controller/usage.go
```diff
@@ -103,5 +104,24 @@ func (c *UsageController) CanUploadFile(ctx context.Context, userID int64, size
...
+		for _, familyMember := range familyMembers {
+			if familyMember.MemberUserID == userID {
+				memberStorage = familyMember.StorageLimit
+				break
+			}
+		}
+		if memberStorage != nil {
+			return stacktrace.Propagate(ente.ErrStorageLimitExceeded, "Size of Files is larger than Family Members Allocated Storage")
+		}
```
ua741: As discussed, this won't work.

At the bottom, before returning nil (as no error), let's add a check
```go
if subscriptionAdminID != userID && memberUsage != nil {
      memberUsage , mUsageErr := c.UsageRepo.GetUsage(userID);
       // err handling
       if  size != nil {
         memberUsage += *size
         }
         if memberUsage > (memberUsage + StorageOverflowAboveSubscriptionLimit) {
           // return error here for ErrStorageLimitExceeded
         }
 }
 ```

File: server/ente/family.go (hunk 1 above)
Comment (mngshm) [reply server/ente/family.go]: Resolved in the below commit

File: server/pkg/controller/family/admin.go (hunk 2 above)
Comment (mngshm) [reply server/pkg/controller/family/admin.go]: Solved in below commit also removed the SetStorage Method

File: server/pkg/controller/usage.go (hunk 4 above)
Comment (mngshm) [reply server/pkg/controller/usage.go]: fixed in #[2da115b](https://github.com/ente-io/ente/pull/4961/commits/2da115b6b77ac2f6265dc5d7cd1c6498d48da4d5)

File: server/pkg/controller/family/admin.go [hunk 5]
```diff
@@ -114,10 +114,20 @@ func (c *Controller) InviteMember(ctx *gin.Context, adminUserID int64, email str
...
+
+	// If storageLimit is greater than 0, set the StorageLimit assigned to the Member
+	// or else set no storage limit for that particular Family member.
+	if StorageLimit != nil {
+		_, err := c.FamilyRepo.AddMemberInvite(ctx, adminUserID, potentialMemberID, activeInviteToken, StorageLimit)
+		if err != nil {
+			return stacktrace.Propagate(err, "failed to Set Storage Limit")
+		}
+	}
+
```
ua741: This is not required now.

server/pkg/controller/usage.go
```diff
@@ -45,11 +46,11 @@ func (c *UsageController) CanUploadFile(ctx context.Context, userID int64, size
 	var subscriptionUserIDs []int64
 	// if user is part of a family group, validate if subscription of familyAdmin is valid & member's total storage
 	// is less than the storage accordingly to subscription plan of the admin
+	familyMembers, err := c.FamilyRepo.GetMembersWithStatus(*familyAdminID, repo.ActiveFamilyMemberStatus)
+	if err != nil {
+		return stacktrace.Propagate(err, "failed to fetch family members")
+	}
```
ua741: This will cause panic if user is not part of any family plan, pls test for user who is not part of any family plan.

server/pkg/controller/usage.go
```diff
@@ -103,5 +104,28 @@ func (c *UsageController) CanUploadFile(ctx context.Context, userID int64, size
...
 	}
+
+	// Get particular member's storage and check if the file size is larger than the size of the storage allocated
+	// to the Member and fail if its too large.
+	var memberStorage *int64
+	for _, familyMember := range familyMembers {
+		memberStorage = familyMember.StorageLimit
+		break
+	}
+	if subscriptionAdminID != userID {
```
ua741: Here also we should have a check for memberStorage not being nil. For existing users and users with no limit, this value will be nil. You can test by setting storage to null in families db

server/pkg/controller/family/admin.go
```diff
@@ -57,7 +57,7 @@ func (c *Controller) CreateFamily(ctx context.Context, adminUserID int64) error
 }
 
 // InviteMember invites a user to join the family plan of admin User
-func (c *Controller) InviteMember(ctx *gin.Context, adminUserID int64, email string) error {
+func (c *Controller) InviteMember(ctx *gin.Context, adminUserID int64, email string, StorageLimit *int64) error {
```
ua741: storageLimit (camel case)

server/pkg/controller/usage.go
mngshm: fixed in [26a407c](https://github.com/ente-io/ente/pull/4961/commits/26a407cc81cb6db8ae1d8132b0e53856ab9d1568)

server/pkg/controller/usage.go
mngshm: fixed in [26a407c](https://github.com/ente-io/ente/pull/4961/commits/26a407cc81cb6db8ae1d8132b0e53856ab9d1568)

ua741 (approved):

PR #5028, opened 2025-02-11
[server] API for modifying family member storage limit

mngshm:
## Tests 

Tested in followin cases 
1. various statuses "REJECTED", "REVOKED" and "SELF" etc
2. user not part of any family & admin user
3. storage limit check on bulk upload
4. reduce storage check 
5. set storage bigger than anything available in the world
6. admin shouldn't be able to set his own storage.

- [x] check for potential refactoring

### DIFF
--- a/server/ente/errors.go
+++ b/server/ente/errors.go
@@ -94,6 +94,13 @@ var ErrLockUnavailable = errors.New("could not acquire lock")
 // ErrActiveLinkAlreadyExists is thrown when the collection already has active public link
 var ErrActiveLinkAlreadyExists = errors.New("Collection already has active public link")
 
+// ErrCannotModifyAdminStoragLimit is thrown when the admin is modifying their own storage limit
+var ErrCannotModifyAdminStoragLimit = errors.New("Cannot modify admin storage limit")
+
+// ErrFailedReducingStorageLimit is thrown when the admin is setting storage limit lesser than
+// members pre-consumed storage
+var ErrFailedReducingStorageLimit = errors.New("Members Usage is more than potential storage limit")
+
 // ErrNotImplemented indicates that the action that we tried to perform is not
 // available at this museum instance. e.g. this could be something that is not
 // enabled on this particular instance of museum.
--- a/server/ente/errors.go
+++ b/server/ente/errors.go
@@ -95,11 +95,11 @@ var ErrLockUnavailable = errors.New("could not acquire lock")
 var ErrActiveLinkAlreadyExists = errors.New("Collection already has active public link")
 
 // ErrCannotModifyAdminStoragLimit is thrown when the admin is modifying their own storage limit
-var ErrCannotModifyAdminStoragLimit = errors.New("Cannot modify admin storage limit")
+var ErrCannotModifyAdminStoragLimit = errors.New("cannot modify admin storage limit")
 
 // ErrFailedReducingStorageLimit is thrown when the admin is setting storage limit lesser than
 // members pre-consumed storage
-var ErrFailedReducingStorageLimit = errors.New("Members Usage is more than potential storage limit")
+var ErrFailedReducingStorageLimit = errors.New("members usage is more than potential storage limit")
 
 // ErrNotImplemented indicates that the action that we tried to perform is not
 // available at this museum instance. e.g. this could be something that is not
--- a/server/ente/errors.go
+++ b/server/ente/errors.go
@@ -94,13 +94,6 @@ var ErrLockUnavailable = errors.New("could not acquire lock")
 // ErrActiveLinkAlreadyExists is thrown when the collection already has active public link
 var ErrActiveLinkAlreadyExists = errors.New("Collection already has active public link")
 
-// ErrCannotModifyAdminStoragLimit is thrown when the admin is modifying their own storage limit
-var ErrCannotModifyAdminStoragLimit = errors.New("cannot modify admin storage limit")
-
-// ErrFailedReducingStorageLimit is thrown when the admin is setting storage limit lesser than
-// members pre-consumed storage
-var ErrFailedReducingStorageLimit = errors.New("members usage is more than potential storage limit")
-
 // ErrNotImplemented indicates that the action that we tried to perform is not
 // available at this museum instance. e.g. this could be something that is not
 // enabled on this particular instance of museum.
--- a/server/pkg/controller/family/admin.go
+++ b/server/pkg/controller/family/admin.go
@@ -230,11 +230,11 @@ func (c *Controller) ModifyMemberStorage(ctx context.Context, actorUserID int64,
 		if memberUsage > *storageLimit {
 			return stacktrace.Propagate(ente.NewBadRequestWithMessage("Failed to reduce storage"), "User's current usage is more")
 		}
+	}
 
-		modifyStorageErr := c.FamilyRepo.ModifyMemberStorage(ctx, actorUserID, member.ID, storageLimit)
-		if modifyStorageErr != nil {
-			return stacktrace.Propagate(modifyStorageErr, "Failed to modify members storage")
-		}
+	modifyStorageErr := c.FamilyRepo.ModifyMemberStorage(ctx, actorUserID, member.ID, storageLimit)
+	if modifyStorageErr != nil {
+		return stacktrace.Propagate(modifyStorageErr, "Failed to modify members storage")
 	}
 
 	return nil
--- a/server/pkg/controller/family/admin.go
+++ b/server/pkg/controller/family/admin.go
@@ -190,6 +190,37 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
 	return nil
 }
 
+// ModifyMemberStorage allows admin user to update the storageLimit for a member in the family
+func (c *Controller) ModifyMemberStorage(ctx context.Context, adminID int64, id uuid.UUID, storageLimit *int64) error {
+	familyAdminID, err := c.UserRepo.GetFamilyAdminID(adminID)
+	if err != nil {
+		stacktrace.Propagate(err, "Could not get family admin ID")
+	}
+
+	member, err := c.FamilyRepo.GetMemberById(ctx, id)
+	if err != nil {
+		stacktrace.Propagate(err, "Couldn't fetch Family Member")
+	}
+
+	// get admin subscription in order to get the size of total storage quota
+	// available in that family. and let the user chose storage of outw of the whole
+	// storage quota or maximum limit.
+	activeSub, err := c.BillingCtrl.GetActiveSubscription(adminID)
+	if err != nil {
+		stacktrace.Propagate(err, "couldn't get active subscription")
+	}
+
+	if adminID == *familyAdminID && id == member.ID {
+		if storageLimit != nil && *storageLimit > activeSub.Storage {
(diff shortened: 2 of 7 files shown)


server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,54 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
+
+	// gets admin subscription in order to get the size of total storage quota
+	activeSub, err := c.BillingCtrl.GetActiveSubscription(adminID)
+	if err != nil {
+		return stacktrace.Propagate(ente.ErrNoActiveSubscription, "couldn't get active subscription")
+	}
+
+	if storageLimit != nil && *storageLimit > activeSub.Storage {
+		return stacktrace.Propagate(ente.ErrStorageLimitExceeded, "potential storage limit is more than subscription storage")
+	}
```
ua741: We also have storage bonus. For an admin, you can probably just call FetchMembersForAdminID, this will return both members and adminStorage and storageBonus.

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,54 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
 	return nil
 }
 
+// ModifyMemberStorage allows admin user to update the storageLimit for a member in the family
+func (c *Controller) ModifyMemberStorage(ctx context.Context, adminID int64, id uuid.UUID, storageLimit *int64) error {
+	familyAdminID, err := c.UserRepo.GetFamilyAdminID(adminID)
+	if err != nil {
+		return stacktrace.Propagate(ente.ErrBadRequest, "Could not get family admin ID")
+	}
+
```
ua741: We don't need to make this api call.
Once the member is fetched, then you can

check member.isAdmin -> return error that admin storage can not be modifeid
member.AdminID != actorUserID (here actorUserID is same as adminID in the request), then reject saying you don't have sufficient perm

server/pkg/controller/family/admin.go
ua741: Also, may be there's no need to check for adminStorage. Just member storage is enough. Anyways by default users has unbounded storage.

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,54 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
+	if adminID == *familyAdminID && member.IsAdmin {
+		return stacktrace.Propagate(ente.ErrCannotModifyAdminStoragLimit, "cannot modify admin storage limit")
+	}
+
+	if memberUsage > *storageLimit {
+		return stacktrace.Propagate(ente.ErrFailedReducingStorageLimit, "Cannot reduce storage, current usage is more.")
+	}
+
+	if id == member.ID && member.Status == "ACCEPTED" {
```
ua741: id == member.ID is redundant check because we anyways checked it via id.
Also, we should check for status early before fetching other things.. Can allow them to modify if user status is invited or accepted.

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,54 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
+	memberUsage, memUsageErr := c.UsageRepo.GetUsage(member.MemberUserID)
+	if memUsageErr != nil {
+		return stacktrace.Propagate(memUsageErr, "Couldn't find members storage usage")
+	}
+
+	if adminID == *familyAdminID && member.IsAdmin {
+		return stacktrace.Propagate(ente.ErrCannotModifyAdminStoragLimit, "cannot modify admin storage limit")
+	}
+
+	if memberUsage > *storageLimit {
```
ua741: missing nil check for `storageLimit`

mngshm: This was done from L211-L213, hence skipped it. Also, while thoroughly testing, I did not see any panics or dereference errors.

Tested this again before pushing changes

https://github.com/ente-io/ente/blob/15d58e3446deb707bc4e7e1dd2e7b04778051f38/server/pkg/controller/family/admin.go#L211-L213

server/pkg/controller/family/admin.go
mngshm: Ommited `id == member.ID` and moved it just after the call to `GetMemberByID`.

server/pkg/controller/family/admin.go
mngshm: Removed the API call. made use of `member.IsAdmin` and the actorUserID as well

server/pkg/controller/family/admin.go
mngshm: Modified the function call for `FetchMembersForAdminID`. I added this check so that the admin does not enter a too large (than the subscription+bonus) storage Limit.

mngshm:
@ua741 resolved the comments

server/ente/errors.go
```diff
@@ -94,6 +94,13 @@ var ErrLockUnavailable = errors.New("could not acquire lock")
 // ErrActiveLinkAlreadyExists is thrown when the collection already has active public link
 var ErrActiveLinkAlreadyExists = errors.New("Collection already has active public link")
 
+// ErrCannotModifyAdminStoragLimit is thrown when the admin is modifying their own storage limit
+var ErrCannotModifyAdminStoragLimit = errors.New("cannot modify admin storage limit")
+
+// ErrFailedReducingStorageLimit is thrown when the admin is setting storage limit lesser than
+// members pre-consumed storage
+var ErrFailedReducingStorageLimit = errors.New("members usage is more than potential storage limit")
```
ua741: No need to define new errors. These unknown errors will result in 5xx error code.
We can just reuse NewBadRequestWithMessage func to get new bad request error

server/pkg/repo/family.go
```diff
@@ -196,6 +196,16 @@ func (repo *FamilyRepository) RemoveMember(ctx context.Context, adminID int64, m
 	return stacktrace.Propagate(tx.Commit(), "failed to commit")
 }
 
+// UpdateStorage is used to set Pre-existing Members Storage Limit.
+func (repo *FamilyRepository) ModifyMemberStorage(ctx context.Context, adminID int64, id uuid.UUID, storageLimit *int64) error {
```
ua741, line 200: nit: adminID is not required.

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,53 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
+		return stacktrace.Propagate(memUsageErr, "Couldn't find members storage usage")
+	}
+
+	if memberUsage > *storageLimit {
+		return stacktrace.Propagate(ente.ErrFailedReducingStorageLimit, "Cannot reduce storage, current usage is more.")
+	}
+
+	modifyStorageErr := c.FamilyRepo.ModifyMemberStorage(ctx, adminID, member.ID, storageLimit)
+	if modifyStorageErr != nil {
+		return stacktrace.Propagate(modifyStorageErr, "Cannot Modify Members Storage")
```
ua741: nit: failed to modify member's storage

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,53 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
+	// Handle if the admin user tries reducing the storage Limit
+	// and the members Usage is more than the potential storage Limit
+	memberUsage, memUsageErr := c.UsageRepo.GetUsage(member.MemberUserID)
+	if memUsageErr != nil {
+		return stacktrace.Propagate(memUsageErr, "Couldn't find members storage usage")
+	}
+
+	if memberUsage > *storageLimit {
+		return stacktrace.Propagate(ente.ErrFailedReducingStorageLimit, "Cannot reduce storage, current usage is more.")
+	}
```
ua741: This should be inside a if block.
`if storageLimit != nil`
Otherwise we are redundantly fetching storage, and also will result in panic as we are referencing storageLimit without checking for nil.

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,53 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
+		return stacktrace.Propagate(ente.ErrBadRequest, "user is not a part of family")
+	}
+
+	// gets admin subscription in order to get the size of total storage quota (including bonus)
+	activeSub, err := c.FetchMembersForAdminID(ctx, member.AdminUserID)
+	if err != nil {
+		return stacktrace.Propagate(ente.ErrNoActiveSubscription, "couldn't get active subscription")
+	}
+
+	if storageLimit != nil && *storageLimit > activeSub.Storage {
```
ua741: activeSub.Storage + activeSub.AdminBonus?
Also, we  may be we can rename activeSub to `familyMembersData`

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,53 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
 	return nil
 }
 
+// ModifyMemberStorage allows admin user to update the storageLimit for a member in the family
+func (c *Controller) ModifyMemberStorage(ctx context.Context, adminID int64, id uuid.UUID, storageLimit *int64) error {
```
ua741: adminID int64
can we rename this to `actorUserID` or just `userID`

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,55 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
 
+// ModifyMemberStorage allows admin user to update the storageLimit for a member in the family
+func (c *Controller) ModifyMemberStorage(ctx context.Context, actorUserID int64, id uuid.UUID, storageLimit *int64) error {
+	member, err := c.FamilyRepo.GetMemberById(ctx, id)
+	if err != nil {
+		return stacktrace.Propagate(err, "Couldn't fetch Family Member")
+	}
+
+	if member.AdminUserID != actorUserID {
+		return stacktrace.Propagate(ente.ErrBadRequest, "you do not have sufficient permission")
```
ua741: This should be ente.ErrPermissionDenied

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,55 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
+	}
+
+	if member.IsAdmin {
+		return stacktrace.Propagate(ente.NewBadRequestWithMessage("failed to modify storage"), "cannot modify admin storage limit")
```
ua741: NewBadRequestWithMessage("can not limit admin storage")

server/pkg/controller/family/admin.go
```diff
@@ -190,6 +190,55 @@ func (c *Controller) CloseFamily(ctx context.Context, adminID int64) error {
...
+	if memberUsage > *storageLimit {
+		return stacktrace.Propagate(ente.NewBadRequestWithMessage("Failed to reduce storage"), "User's current usage is more")
+	}
+
+	modifyStorageErr := c.FamilyRepo.ModifyMemberStorage(ctx, actorUserID, member.ID, storageLimit)
+	if modifyStorageErr != nil {
+		return stacktrace.Propagate(modifyStorageErr, "Failed to modify members storage")
+	}
+	return nil
+}
```
ua741: This also needs to be inside that if block?

PR #5159, opened 2025-02-24
[server] Allow null values for storage_limit (col) & remove storageLimit dep from Invite Member workflow

mngshm:


server/migrations/96_storage_table_families.up.sql
```diff
@@ -1,2 +1,2 @@
 ALTER TABLE families
-ADD COLUMN storage_limit BIGINT;
\ No newline at end of file
+ADD COLUMN storage_limit BIGINT NULL;
```
ua741, line 2: This is a no-op as by default the value will be already NULL.

Also, any modification post main merge should go as new version otherwise the changes won't be applied.

server/pkg/controller/family/admin.go
```diff
@@ -57,7 +57,7 @@ func (c *Controller) CreateFamily(ctx context.Context, adminUserID int64) error
 }
 
 // InviteMember invites a user to join the family plan of admin User
-func (c *Controller) InviteMember(ctx *gin.Context, adminUserID int64, email string, storageLimit *int64) error {
+func (c *Controller) InviteMember(ctx *gin.Context, adminUserID int64, email string) error {
```
ua741, line 60: This change is not really required.
During invite we anyways don't check for usage. May be it's fine to have the existing change as it is.. Where user can just invite someone with X storage limit.
Once they accept the invite, they won't be able to upload any photos.


We do need to keep the check when admin is trying to modify members storage.
In that, We can just fail the request early if the member status is just invited.

Essentially, you just want to modify this check
      ```go
      if member.Status != ente.ACCEPTED && member.Status != ente.INVITED {
		return stacktrace.Propagate(ente.ErrBadRequest, "user is not a part of family")
	}
	```
	
	to
	
	```go
	if member.Status != ente.ACCEPTED  {
		return stacktrace.Propagate(ente.ErrBadRequest, "user is currently not a part of family")
	}```

server/migrations/96_storage_table_families.up.sql
mngshm: Thanks, I'll most probably revert the changes in this file. Don't think this is necessary if NULL values are allowed (also tested locally with a new column)

PR #5168, opened 2025-02-25
[fix] steaming fixes

prateekmedia:
## Description

- Update queuing logic
- Use native_video_player + exoplayer branch to fix artifacts for android
- Support newer namespace and sdk 35
- Store failures and don't retry them
- Modify how sync file data status is happening i.e. don't remote sync everytime we need to update previewIds, instead update it locally
- Add bandwidth check before queuing
- Remove rotate metadata parameter

Now the player decision logic is:
- Android: Native video Player for both Stream and original source
- iOS: Media Kit for stream and Native Video Player for Original

The UseMediaKitEvent is still there in case of any error or if user long presses on that button (android) for original source.


mobile/lib/services/preview_video_store.dart
```diff
@@ -58,9 +60,9 @@ class PreviewVideoStore {
   void init(SharedPreferences prefs) {
     _prefs = prefs;
 
-    FileDataService.instance.syncFDStatus().then(
-          (_) => _putFilesForPreviewCreation(),
-        );
+    if (!initSuccess) {
+      _putFilesForPreviewCreation();
+    }
```
ua741: This will always run on app init as initSuccess will be false?

mobile/lib/db/upload_locks_db.dart
```diff
@@ -118,6 +127,15 @@ class UploadLocksDB {
                   PRIMARY KEY (${_partsTable.columnObjectKey}, ${_partsTable.columnPartNumber})
                 )
                 ''',
+      '''
+                CREATE TABLE IF NOT EXISTS ${_streamUploadErrorTable.table} (
+                  ${_streamUploadErrorTable.columnID} INTEGER PRIMARY KEY,
```
ua741: Is there any reason for adding separate id? Why can't we just use uploadedFileID as the primary key?

mobile/lib/services/filedata/filedata_service.dart
```diff
@@ -29,6 +29,12 @@ class FileDataService {
     _prefs = prefs;
   }
 
+  void appendPreview(int id) {
```
ua741: Can you document what this method does and why do we need it? Unable to infer quickly while looking that the code

mobile/lib/services/preview_video_store.dart
prateekmedia: It will basically not trigger again in sync is ran again by any other function. I have updated the logic how it will be set and reset based on the state of the logic.

mobile/lib/db/upload_locks_db.dart
prateekmedia: Ok It's fine to remove that.

mobile/lib/services/filedata/filedata_service.dart
prateekmedia: I have documented, let me know if it seems fine?

mobile/android/gradle.properties
```diff
@@ -1,3 +1,4 @@
 org.gradle.jvmargs=-Xmx4608m
 android.useAndroidX=true
 android.enableJetifier=true
+android.enableR8.fullMode=false
```
ua741: Let's add an comment here why this is disable, with may be some references for relevant Github Issues. So that in future, if someone revisits this, they will have some context.

mobile/lib/services/preview_video_store.dart
```diff
@@ -34,12 +35,18 @@ import "package:photos/utils/exif_util.dart";
...
+import "package:photos/utils/network_util.dart";
 import "package:photos/utils/toast_util.dart";
 import "package:shared_preferences/shared_preferences.dart";
 
+const _maxRetryCount = 3;
+
 class PreviewVideoStore {
   final LinkedHashMap<int, PreviewItem> _items = LinkedHashMap();
   LinkedHashMap<int, PreviewItem> get previews => _items;
+  late Set<int> _failureFiles;
```
ua741: It might be better to either define this as emptySet or nullable.
Ideally, late variables should be initialized in the construction phase. Otherwise there's a risk of running into late variable not initialized error

mobile/lib/db/upload_locks_db.dart
```diff
@@ -291,6 +307,67 @@ class UploadLocksDB {
...
+
+    await db.insert(
+      _streamUploadErrorTable.table,
+      {
+        _streamUploadErrorTable.columnUploadedFileID: uploadedFileID,
+        _streamUploadErrorTable.columnErrorMessage: errorMessage,
+        _streamUploadErrorTable.columnLastAttemptedAt:
+            DateTime.now().millisecondsSinceEpoch,
+      },
+    );
```
ua741: Let's specify the conflict resolution here. insertOrReplace?

mobile/lib/services/filedata/filedata_service.dart
ua741: Understood. Don't we have objectID and size information avoid at this point? If yes, why not just append that?

mobile/lib/db/upload_locks_db.dart
prateekmedia: insertOrReplace is not a method.

mobile/lib/services/filedata/filedata_service.dart
prateekmedia: Got it, we do have that info as it is just uploaded.

ua741 (approved):

PR #5190, opened 2025-02-26
[auth] Add support for editing number of digits & algorithm

AmanRajSinghMourya:
## Description
This PR add support to edit the number of digits (between 1 to 10) for the 2FA codes and also give an option to select algorithms 
image

### DIFF
--- a/auth/lib/models/code.dart
+++ b/auth/lib/models/code.dart
@@ -116,16 +116,17 @@ class Code {
     Algorithm? algorithm,
   ) {
     final String encodedIssuer = Uri.encodeQueryComponent(issuer);
+    final Algorithm algo = algorithm ?? Algorithm.sha1;
     return Code(
       account,
       issuer,
       digits,
       defaultPeriod,
       secret,
-      algorithm ?? Algorithm.sha1,
+      algo,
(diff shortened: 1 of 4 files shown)


auth/lib/models/code.dart
```diff
@@ -121,7 +122,7 @@ class Code {
       digits,
       defaultPeriod,
       secret,
-      Algorithm.sha1,
+      algorithm ?? Algorithm.sha1,
       type,
       0,
       "otpauth://${type.name}/$issuer:$account?algorithm=SHA1&digits=$digits&issuer=$encodedIssuer&period=30&secret=$secret",
```
ua741: The algorithm needs to change here

auth/lib/models/code.dart
```diff
@@ -113,18 +113,20 @@ class Code {
     String secret,
     CodeDisplay? display,
     int digits,
+    Algorithm? algorithm,
```
ua741: These can be named parameters, we don't have to modify existing usage for these methods.

ua741 (approved):

ua741 (approved):

PR #5232, opened 2025-03-05
[auth] Workflow fixes

prateekmedia:
## Description

Remove deprecated ubuntu-20.04 workflow and merge it with the latest.

### DIFF
--- a/.github/workflows/auth-release.yml
+++ b/.github/workflows/auth-release.yml
@@ -32,8 +32,8 @@ env:
     FLUTTER_VERSION: "3.24.3"
 
 jobs:
-    build-ubuntu:
-        runs-on: ubuntu-20.04
+    build-linux-latest:
+        runs-on: ubuntu-latest
 
         defaults:
             run:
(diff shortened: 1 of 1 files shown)


.github/workflows/auth-release.yml
```diff
@@ -163,17 +102,16 @@ jobs:
             - name: Build desktop app
               run: |
                   flutter config --enable-linux-desktop
-                  # dart pub global activate flutter_distributor
-                  dart pub global activate --source git https://github.com/prateekmedia/flutter_distributor --git-ref develop --git-path packages/flutter_distributor
-                  # Run below command if it is a beta or nightly
-                  if [[ ${{ github.ref }} =~ beta|nightly ]]; then
-                    flutter_distributor package --platform=linux --targets=pacman --skip-clean
-                    mv dist/**/*-*-linux.pacman artifacts/ente-${{ github.ref_name }}-x86_64.pacman
```
ua741, line 171: We don't need this anymore?

ua741 (approved):

.github/workflows/auth-release.yml
prateekmedia: Yes, pacman is not stable yet. It doesn't work simply put as of its current implementation.

PR #5234, opened 2025-03-05
[mob][photos] Properly handle and log error/exception thrown in isolate 

ashilkn:


mobile/lib/services/local/local_sync_util.dart
```diff
@@ -34,17 +34,26 @@ Future<Tuple2<List<LocalPathAsset>, List<EnteFile>>> getLocalPathAssetsAndFiles(
...
+          },
+          taskName:
+              "getLocalPathAssetsAndFiles-${pathEntity.name}-count-${assetsInPath.length}",
+        );
+      } catch (e) {
+        _logger.severe("_getLocalIDsAndFilesFromAssets failed", e);
+        _logger.info(
+          "Failed for pathEntity: ${pathEntity.name}",
+        );
+        result = const Tuple2({}, []);
```
ua741: Won't this result in nothing getting imported? Let's discuss about this in person before merging this case.
May we can test the baheviour by throwing exception when entity name is equal to one of the local file on the device..

ua741 (approved):

PR #5415, opened 2025-03-23
Add OTP ente_account_add.md

voodoologic:
Clear up documentation for users who are trying to get set up.

## Description
I spent a lot of time digging through the documentation in order to discover what OTP was and where the password was.  This should help others.  Ideally the CLI would mention this as well.

### DIFF
--- a/cli/docs/generated/ente_account_add.md
+++ b/cli/docs/generated/ente_account_add.md
@@ -15,6 +15,7 @@ ente account add [flags]
 ```
   -h, --help   help for add
 ```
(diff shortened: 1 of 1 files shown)


ua741:
This is an auto-generated docs which will get cleaned up.
We have already documented guide for self-hosting here: https://help.ente.io/self-hosting/faq/otp.
If there's any suggestion or changes which you think will be helpful, request you to please start a discussion and share screenshots of the step/place where additional clarification can be helpful.

PR #5419, opened 2025-03-23
[Auth] Flakey fix for steam codes in web app

### LINKED ISSUE #5325: [Auth] Wrong display of auth codes
### Description

I added Steam authorization through external tool and it correctly displays on desktop and mobile versions, but displays default 6-digit on web version. Screenshots of desktop and web are below.

For context, these codes are working, I tested it just before writing an issue.

That's a web
Image

That's a desktop one

Image

### Version

_No response_

Ente Auth

Web - Firefox

Amanse:
## Description
There is difference in how steam otps are handled in flutter and in web app, in flutter we check both code.type and code.issuer to be steam while in Web app we only rely on code.type, which breaks code generation for steam.

Closes #5325 

I know this fix is not of best quality, if you are open to change this generate codes function to not use switch and instead we call based on conditions to include issuer condition as well, I think that can be a better solution.

### DIFF
--- a/web/apps/auth/src/services/code.ts
+++ b/web/apps/auth/src/services/code.ts
@@ -262,6 +262,9 @@ const parseCodeDisplay = (url: URL): CodeDisplay | undefined => {
 export const generateOTPs = (code: Code): [otp: string, nextOTP: string] => {
     let otp: string;
     let nextOTP: string;
+    if (code.issuer.toLowerCase().includes("steam")) {
+        code.type = "steam";
+    }
     switch (code.type) {
         case "totp": {
             const totp = new TOTP({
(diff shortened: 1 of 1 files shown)


Amanse:
As code.type will only be used in this context, correcting it before this won't really make much of difference and this way let's us keep the switch statement.

prateekmedia:
Can you make it change type to steam when code object is getting constructed i.e. decryption is taking place.

Amanse:
Yes, That sounds good, I'll get to it in a few hours

mnvr:
Apologies for noticing this late, unfortunately I'll have to close this. Instead of propagating the workaround to all (current and future) clients, we'd rather fix it in the source (mobile app).

Amanse:
I can give it a shot if no one else is working on it, but it should include not only setting correct type while adding but also updating type for existing entries with issuer as steam as well... but I am guessing we shouldn't do that without any user action, like updating their entries in db as well when they load?

PR #5432, opened 2025-03-25
[auth] Misc fixes & locale improvements

ua741:


niklaskorz:
Why did the commit bumping the macOS target introduce a dependency on Java for the Linux build?

PR #5466, opened 2025-03-28
[auth] add custom icon

Bl4ckspell7:
## Description

add icon for fanatical

fanatical

### DIFF
--- a/auth/assets/custom-icons/_data/custom-icons.json
+++ b/auth/assets/custom-icons/_data/custom-icons.json
@@ -377,6 +377,13 @@
       "hex": "858585"
     },
     {
+      "title": "Fanatical",
+      "slug": "fanatical",
+      "altNames": [
+        "FANATICAL"
+      ]
+    },
+    {
       "title": "Fastmail"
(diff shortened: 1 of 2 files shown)


auth/android/.gitignore
```diff
@@ -5,6 +5,7 @@ gradle-wrapper.jar
 /gradlew.bat
 /local.properties
 GeneratedPluginRegistrant.java
+/app/.cxx/*
```
ua741: Can we undo this change?

Bl4ckspell7: yes, but `.cxx/` is just build output from the NDK - safe to ignore and shouldn't be in version control.

ua741: Ack, at what steps where you seeing changes in this directory? Asking because we have run into changes in this path.

Bl4ckspell7: the `.cxx` directory got created when running `$ flutter build apk --release --flavor independent`

image

ua741 (approved):

PR #5546, opened 2025-04-05
Sentence fix in translation.json

mekecemre:
Changed the sentence "The folders you add here will monitored to automatically" to "The folders you add here will be monitored automatically"

### DIFF
--- a/web/packages/base/locales/en-US/translation.json
+++ b/web/packages/base/locales/en-US/translation.json
@@ -477,7 +477,7 @@
     "watch_folders": "Watch folders",
     "watched_folders": "Watched folders",
     "no_folders_added": "No folders added yet",
-    "watch_folders_hint_1": "The folders you add here will monitored to automatically",
+    "watch_folders_hint_1": "The folders you add here will be monitored automatically",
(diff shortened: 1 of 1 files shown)


mnvr:
Thank you. But the current sentence is correct, it is the lead up to the subsequent points:

Screenshot 2025-04-07 at 09 55 20

Also, changing the translations for English invalidates all the other language translations, so we'd only like to change it if there is a functional issue, not stylistic preferences.

---

(ps: Appreciate your intent to make Ente better!)

j23n:
I just stumbled across this as well and it is incorrect. The passive voice requires the verb "to be"! It should read: 

> The folders you add here will be monitored to automatically:

The phrase "will monitored" is incomplete - you need "will be monitored" to form the future passive voice.

Alternatively:

> The folders you add here are monitored to automatically:

This is in line with e.g. the German and French translations and not a change in the meaning requiring new translation but rather a fix of a grammatical mistake.

mnvr:
Ah, I see it now, the "be" is missing!

I've updated the English string (in Crowdin and in source code) to match the second alternative you suggested: https://github.com/ente-io/ente/pull/7118

Thank you both!

PR #5550, opened 2025-04-07
[server] Retention emails

mngshm:


server/cmd/museum/main.go
```diff
@@ -992,12 +992,17 @@ func setupAndStartCrons(userAuthRepo *repo.UserAuthRepository, publicCollectionR
...
+		emailNotificationCtrl.SendFamilyNudgeEmail()
+	})
+
 	scheduleAndRun(c, "@every 60m", func() {
 		emergencyCtrl.SendRecoveryReminder()
 		kexCtrl.DeleteOldKeys()
 	})
 
 	c.Start()
+	log.Infoln(c.Entry(17).Next)
```
ua741: remove?

server/pkg/controller/billing.go
```diff
@@ -366,6 +370,13 @@ func (c *BillingController) UpdateSubscription(r ente.UpdateSubscriptionRequest)
 	if err != nil {
 		return stacktrace.Propagate(err, "")
 	}
+
+	// Delete users entry from notification_history table on subscription update
+	err = c.NotificationHistory.DeleteLastNotificationTime(r.UserID, "90_percent_consumed")
+	if err != nil {
+		return stacktrace.Propagate(err, "")
+	}
+
```
ua741: Is this flow also triggered when a normal user changes their subscription or just when support team updates the subscription?

server/pkg/repo/user.go
```diff
@@ -280,17 +310,17 @@ func (repo *UserRepository) GetPublicKey(userID int64) (string, error) {
...
+func (repo *UserRepository) GetUsersWithExceedingStorages(percentageThreshold int64) ([]ente.User, error) {
 	rows, err := repo.DB.Query(`
 		SELECT users.user_id, users.encrypted_email, users.email_decryption_nonce, users.email_hash, usage.storage_consumed, subscriptions.storage
 		FROM users 
 		INNER JOIN usage 
 		ON users.user_id = usage.user_id 
 		INNER JOIN subscriptions 
-		ON users.user_id = subscriptions.user_id AND usage.storage_consumed > subscriptions.storage AND users.encrypted_email IS NOT NULL AND users.family_admin_id IS NULL;
-	`)
+		ON users.user_id = subscriptions.user_id AND usage.storage_consumed >= (subscriptions.storage * $1 / 100.0) AND users.encrypted_email is not null AND users.family_admin_id is NULL;
```
ua741: `* $1 / 100.0` Is this intentional?

server/pkg/controller/email/email_notification.go
```diff
@@ -171,33 +182,113 @@ func (c *EmailNotificationController) SendStorageLimitExceededMails() {
...
+			}
+			if isFamilyAdmin != nil {
+				break
+			}
+
+			thirtyDays := 30 * 24 * t.Hour
+			creationTime := t.Unix(0, user.CreationTime)
+			timeSinceCreation := t.Since(creationTime)
+
+			if timeSinceCreation >= thirtyDays {
```
ua741: We can this filter in `GetSubscribedUsersWithoutFamily`, can pass a param createdBeforeInDays int64 and use that to get only users with account older than 30 days.

mngshm: Hmm. Probably this is a better way to do it then. In one more type of retention-email I have to send an email to list of users who have signed up half hour ago. Maybe i'll try to get that done in this PR itself.

server/pkg/repo/user.go
mngshm: Some AI bad code which missed through my eyes while re-reviewing code.

server/pkg/controller/billing.go
mngshm: Apparently, there are no checks checking if the update was triggered by the user themselves (support is also an admin account if I'm not wrong), so this change will get triggered in both cases. And ideally, I'm guessing this change in the table has to be triggered in any case when the subscription has changed because anyhow to reset the status of 90% consumption.

server/pkg/repo/user.go
mngshm: Yes, this logic is intentional and it filters users based on the storage they have consumed. The queries were ran manually to test if they work while I was writing the code. Here's a proof with 90% and 100% consumptions. 

Capture-Apr-21-Apr-04-1745230664

then the code below this does further logic. Please let me know if there are any inconsistencies you think are still there. 

Also, thank you for raising the question. There was a small bug (which could've had bigger impact) regarding the comparison `usage.storage_consumed >= (subscription.storage * 90 /100)`. The `>=` operator was the bug here leading to get both 90% and 100% consumed list of users when `percentageThreshold == 90`

server/pkg/controller/email/email_notification.go
mngshm: I will give this a try @ua741

server/pkg/controller/billing.go
ua741: If we check the usage of `UpdateSubscription`, we can identify that this method is only called when support updates the subscription.

Assuming if we only want to reset the alerts on subscription change, we should look at where all we are calling `ReplaceSubscription` method in the billing repo.

Also, if a users goes above 90%, deletes some data, and then again goes above 90, we don't want to send email again?

mngshm: >Assuming if we only want to reset the alerts on subscription change, we should look at where all we are calling ReplaceSubscription method in the billing repo. 

Ack, i'll check for these points and make necessary changes. I'm assuming, we will have to make the same change wherever there's any kind of change in the subscription plan?

>Also, if a users goes above 90%, deletes some data, and then again goes above 90, we don't want to send email again?

Ideally, this email notification is added to tell the user that they're going to exceed the storage quota very sooner. So, yes, it will send an email if the user exceeds 90%, deletes some data and then again crosses the threshold.

mngshm: Modified this in [20a26ea](https://github.com/ente-io/ente/pull/5550/commits/20a26eac3bacbe00794b3dcf9a2a8a8d2519f9c2) and [ab5c02d](https://github.com/ente-io/ente/pull/5550/commits/ab5c02d792dba870300f627a88f1ae042c70b842).

server/pkg/controller/email/email_notification.go
mngshm: made requested changes in [ce37e9d](https://github.com/ente-io/ente/pull/5550/commits/ce37e9d573248f78b928bc5c4c7e578b1a3758d7)

server/pkg/controller/email/email_notification.go
```diff
@@ -171,33 +182,113 @@ func (c *EmailNotificationController) SendStorageLimitExceededMails() {
...
+	}
+
+	batchSize := 100
+	totalSubUsers := len(subscribedUsers)
+
+	for i := range totalSubUsers {
+		end := i + batchSize
+		if end > totalSubUsers {
+			end = totalSubUsers
+		}
```
ua741: What's the motivation behind this batching?

server/pkg/repo/user.go
```diff
@@ -280,17 +313,17 @@ func (repo *UserRepository) GetPublicKey(userID int64) (string, error) {
...
+func (repo *UserRepository) GetUsersWithExceedingStorages(percentageThreshold int64) ([]ente.User, error) {
 	rows, err := repo.DB.Query(`
 		SELECT users.user_id, users.encrypted_email, users.email_decryption_nonce, users.email_hash, usage.storage_consumed, subscriptions.storage
 		FROM users 
 		INNER JOIN usage 
 		ON users.user_id = usage.user_id 
 		INNER JOIN subscriptions 
-		ON users.user_id = subscriptions.user_id AND usage.storage_consumed > subscriptions.storage AND users.encrypted_email IS NOT NULL AND users.family_admin_id IS NULL;
-	`)
+		ON users.user_id = subscriptions.user_id AND usage.storage_consumed = (subscriptions.storage * $1 / 100) AND users.encrypted_email is not null AND users.family_admin_id is NULL;
```
ua741: This ($1/100) change is still present.

server/pkg/controller/email/email_notification.go
```diff
@@ -171,33 +182,114 @@ func (c *EmailNotificationController) SendStorageLimitExceededMails() {
...
+			if err != nil {
+				return stacktrace.Propagate(err, "User is a family admin")
+			}
+			if isFamilyAdmin != nil {
+				break
+			}
+
+			creationTime := t.Unix(0, user.CreationTime)
+			timeSinceCreation := t.Since(creationTime)
+			if timeSinceCreation >= thirtyDaysDuration {
```
ua741: Both GetFamilyAdminID and creationTime check are redundant? `GetSubscribedUsersWithoutFamily` function has already checks/filters for it.

server/pkg/repo/user.go
```diff
@@ -128,6 +129,38 @@ func (repo *UserRepository) GetAll(sinceTime int64, tillTime int64) ([]ente.User
 	return users, nil
 }
 
+// GetSubscribedUsers will return the list of all subscribed.
+func (repo *UserRepository) GetSubscribedUsersWithoutFamily(duration string) ([]ente.User, error) {
```
ua741: So, the created_at does not get updated when user switches from free to paid plan.
```go
func (repo *UserRepository) GetSubscribedUsersWithoutFamily(accountAgeInDays int) ([]int64, error) {
	rows, err := repo.DB.Query(`SELECT u.user_id, u.encrypted_email, u.email_decryption_nonce, u.email_hash, s.created_at 
	FROM subscriptions s 
	INNER JOIN users u ON s.user_id = u.user_id 
	WHERE u.creation_time <= $1
	AND u.family_admin_id IS NULL
	AND u.encrypted_email IS NOT NULL
	AND s.product_id != 'free'`, time.MicrosecondBeforeDays(accountAgeInDays))
	
	...
```

server/pkg/repo/user.go
```diff
@@ -128,6 +129,38 @@ func (repo *UserRepository) GetAll(sinceTime int64, tillTime int64) ([]ente.User
...
+	rows, err := repo.DB.Query(`SELECT u.user_id, u.encrypted_email, u.email_decryption_nonce, u.email_hash, s.created_at 
+	FROM subscriptions s 
+	INNER JOIN users u ON s.user_id = u.user_id 
+	WHERE s.created_at >= $1
+	AND u.family_admin_id IS NULL
```
ua741: s.created_at >= $1 should be s.created_at <= $1 (fixed it in the suggested query) and also added check for deleted users.

server/pkg/controller/billing.go
```diff
@@ -53,6 +55,7 @@ func NewBillingController(
 	usageRepo *repo.UsageRepository,
 	storageBonusRepo *storagebonus.Repository,
 	commonBillCtrl *commonbilling.Controller,
+	notificationHistory *repo.NotificationHistoryRepository,
```
ua741: nit: notificationHistoryRepo or just notificaitonRepo. Please rename it at other places too.

server/pkg/repo/user.go
mngshm: https://github.com/ente-io/ente/pull/5550#discussion_r2052227392

server/pkg/repo/user.go
ua741: Understood.
I don't think equal will work as user storage might never reach exact 90%. Even if it reaches, the cron runs at fixed interval, it would increase after that.

server/pkg/controller/email/email_notification.go
mngshm: Nope, there aren't. I followed our practice of not doing any checks on repository (DB) level code. Hence, all the checks are being done inside the controller layer.

https://github.com/ente-io/ente/blob/f609cef79e2e040c4c1e9ec929c4b1be1b5a1a29/server/pkg/repo/user.go#L133

server/pkg/controller/email/email_notification.go
mngshm: We had discussed, we would be batching in case for bulk emails 😅 ?

server/pkg/repo/user.go
mngshm: I was re-iterating over this once again. There's a catch 

if we use just `usage.storage_consumed = (subscription.storage * 90 /100)`, your argument is valid. 

but if we use `>=`, the users between 90% - 100% consumption range will be considered for receiving an email. 

One way I can think of is, handling both the thresholds separately.

server/pkg/repo/user.go
mngshm: In this query, the condition happens between `u.creation_time` <= $1. This is not what was intended. The user might chose to not subscribe and consume the free 10GB storage first. Hence, I went ahead with using `s.created_at`. Ideally, the mail has to be sent to those who have crossed 30 days after subscribing and still do not have a family onboarded.

server/pkg/repo/user.go
mngshm: [0f24ba0](https://github.com/ente-io/ente/pull/5550/commits/0f24ba01f523fbb87cb7bdf2d57edecbf14f76a4)

Refactored

server/pkg/repo/user.go
mngshm: Fixed the requested changes in [abe0ef0](https://github.com/ente-io/ente/pull/5550/commits/abe0ef0a0354c674e6ddf00e7b4ef02f386e6da2)

ua741:
Closing this in favor of upcoming retention email changes.

PR #5556, opened 2025-04-08
Update index.md

cmyplay:
* Added explanation to find hostname
* Added explanation of format for the server endpoint address


cmyplay:
Fair enough.

I'm still learning and it's difficult to do without more details.
Will be keen to learn a better way to do it, but this is the only way I could get it working for the time-being.

PR #5572, opened 2025-04-09
[fix] Database Authentication Error in Quickstart

grittypuffy:
# Description

This PR fixes an error in `quickstart.sh` script used for deploying Ente on self-hosted instances where authentication to the database by museum was failing due to failure in sending password in SCRAM-SHA-256 format, which need to be configured in the container's configuration file (`postgresql.conf`)

It fixes a dirty migration caused upon enabling the configuration and the documentation is updated to reflect the changes, allowing users to troubleshoot issues related to password-based authentication failure.

# Notes:
- This error has been replicated on Artix Linux using OpenRC 0.61 and Linux 6.13.8
- The error has not been present on Ubuntu and Debian based distributions

### DIFF
--- a/server/docs/quickstart.md
+++ b/server/docs/quickstart.md
@@ -46,12 +46,12 @@ Then it creates a Docker compose file (`compose.yaml`) from a template, and a
 Finally, it start the Docker compose cluster running the following services on 3
 open ports:
 
-| Service     | Port     | Notes                   |
-| ----------- | -------- | ----------------------- |
-| museum      | `:8080`  | Ente's API server       |
-| web         | `:3000`  | Ente Photos web app     |
-| postgres    |          | DB                      |
-| minio       | `:3200`  | S3                      |
+| Service  | Port    | Notes               |
+| -------- | ------- | ------------------- |
+| museum   | `:8080` | Ente's API server   |
+| web      | `:3000` | Ente Photos web app |
+| postgres |         | DB                  |
+| minio    | `:3200` | S3                  |
 
 For each of these, it'll use the latest published Docker image.
 
@@ -119,6 +119,15 @@ persistent data is saved in volumes managed by Docker.
 > starting museum, one possibility is that you're recreating the `my-ente`
 > folder but still have leftover volumes (`my-ente_postgres-data` and
 > `my-ente_minio-data` from a previous attempt).
+
+> [!TIP]
+>
+> If you're getting errors like `pq: password authentication failed` when
+> starting museum, another possibility is that the password is not being sent
+> in `scram-sha-256` if there are mentions of using `scram-sha-256` in the
+> logs. Set `password_encryption = scram-sha-256` in `postgresql.conf` 
+> (present in `/var/lib/postgresql/data/`) and restart the cluster after
+> removing it along with volumes.
 
 ### Caveat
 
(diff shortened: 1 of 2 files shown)


mngshm:
1. It's not a real problem if the issue isn't replicable or even occurs on Ubuntu based distributions. Any distro from the Arch Linux family hasn't been able to serve as a good server operating system ever :p (saying from experience of using Arch on server). 
2. No one has ever complained in the last 2 weeks
3. `pq: password authentication failure` might occur for many reasons. Less likely that it is related with `scram-sha-256` which is a one-way hashing method. But, a good pick if you've done enough research and then came to a conclusion.

grittypuffy:
1. An update, I have just tried self-hosting on a remote server using Debian and Ubuntu and ended up with the same error with the quickstart.sh script provided by Ente to validate this.
2. Which is surprising, will be helpful if someone is willing to replicate on their system
3. The `postgres` image has a default connection in `pg_hba.conf` with `scram-sha-256`, however the authentication process fails since `password_encryption = scram-sha-256` is not set in the configuration with this error despite running it on a fresh machine without previous volumes:
``` shell
museum-1    | INFO[0000]main.go:104 main Booting up local server with commit #969b0c165518e101f9aaef1354d27715fc31dd96 
museum-1    | INFO[0000]main.go:836 setupDatabase Setting up db                                
museum-1    | INFO[0000]main.go:843 setupDatabase Connected to DB                              
postgres-1  | 2025-04-09 22:04:26.564 UTC [41] FATAL:  password authentication failed for user "pguser"
postgres-1  | 2025-04-09 22:04:26.564 UTC [41] DETAIL:  Connection matched pg_hba.conf line 100: "host all all all scram-sha-256"
museum-1    | panic: pq: password authentication failed for user "pguser"
museum-1    | 
museum-1    | goroutine 1 [running]:
museum-1    | main.setupDatabase()
museum-1    |   /etc/ente/cmd/museum/main.go:846 +0x338
museum-1    | main.main()
museum-1    |   /etc/ente/cmd/museum/main.go:124 +0x44c
museum-1 exited with code 2
```
This was something that was hindering my attempts to deploy 2 days ago and I was trying to fix it so that others don't face similar issues.

mnvr:
I think this might be a case of stale credentials, not the hashing method.

1. The source OS doesn't matter - at this point in the quickstart script, everything is happening inside the docker container.
2. As you mention, no one else has been able to replicate it.

I think what might be happening is that you have stale docker volumes.

e.g. 

1. On a machine, you create a new cluster with quickstart.sh
2. Later you delete that folder, but then create another new cluster with quickstart.sh
3. This cluster will then have new credentials, but will be connecting to a existing docker volume named `my-ente_postgres-data` so postgres will be expecting old credentials.

I've written down a note about this to assist other people who are might also run into the same scenario - https://help.ente.io/self-hosting/troubleshooting/docker#postgres-authentication-failed

Try out those things, I'll close this for now because I think that should solve the issue you're facing, but if it is indeed a different problem, we can dig further.

grittypuffy:
I checked the help documentation before raising the PR to see if it was an error with credentials or stale volumes, but this happens when I try to use `compose` to start the services for the first time using `quickstart.sh` with no pre-existing volumes and I was able to get the services running only by modifying `postgresql.conf`

That's why I intended to bring this to notice, since no one else is able to replicate it, we can leave it.

PR #5586, opened 2025-04-10
[mob] switch to custom ffmpeg kit

prateekmedia:
## Description

Switch to custom ffmpeg kit.

Future actionables on this:
- Creating minimal ffmpeg configuration and using it
- Replace and use .aar for android plugin, currently it is fine and using the published maven package


mobile/ios/Podfile
```diff
@@ -1,6 +1,8 @@
 # Uncomment this line to define a global platform for your project
 platform :ios, '13.0'
 
+source 'https://github.com/ente-io/ffmpeg-kit-custom-repo-ios.git'
```
vishnukvmd, line 4: Let's just call it `ffmpeg-kit`?

mobile/android/build.gradle
```diff
@@ -1,5 +1,6 @@
 ext {
-    appCompatVersion = '1.1.0' // for background_fetch
```
vishnukvmd: Will this change break `background_fetch`?

mobile/pubspec.lock
```diff
@@ -638,13 +638,14 @@ packages:
...
     dependency: "direct main"
     description:
-      name: ffmpeg_kit_flutter_full_gpl
-      sha256: "4f269bcb636bfcb544e5b4d65c706a3d311839970cb42638e72406410c1b5b7b"
-      url: "https://pub.dev"
-    source: hosted
+      path: "flutter/flutter"
+      ref: flurrify
+      resolved-ref: f0f49e4238c1d59fdce06c51f94431831c191c82
+      url: "https://github.com/ente-io/ffmpeg-kit-forked"
```
vishnukvmd: Just `ffmpeg-kit`?

mobile/pubspec.yaml
```diff
@@ -173,8 +177,8 @@ dependencies:
       ref: 2cea396
   screenshot: ^3.0.0
   scrollable_positioned_list: ^0.3.5
-  sentry: ^8.13.2
-  sentry_flutter: ^8.13.2
+  sentry: ^8.14.1
+  sentry_flutter: ^8.14.1
```
vishnukvmd: Best to not modify this dependency within this PR?

mobile/android/build.gradle
prateekmedia: This was made to fix the compilation AFAIK, so it is compiling without it as well. we can test if it was made otherwise.

mobile/ios/Podfile
prateekmedia: This is repository for our pod repo, I think we can call it "pod-repo" or similar so we can host other pods as well in future.

mobile/pubspec.lock
prateekmedia: Yes, we can call that.

mobile/pubspec.yaml
prateekmedia: Without this change the main branch doesn't build, but yes I can create a different PR for this.

prateekmedia: Another PR created.

mobile/android/build.gradle
prateekmedia: I have separated out this code in new PR, let's have discussion there.

vishnukvmd: Will let @ua741 take over 🙏

mobile/ios/Podfile
ua741: How was the OG package handling it?

prateekmedia: They have published it to the official cocoapods repo but I think if we can maintain a custom repo then it would be good cause we are not giving a general package but a stripped out version with our requirements.

ua741 (approved):

PR #5601, opened 2025-04-12
[server] Preview public albums

ua741:
## Tests
[ ] Local testing pending


mnvr (approved):
Tested using https://github.com/ente-io/ente/pull/5625

PR #5604, opened 2025-04-12
[mob] Upgrade targetSdkVersion : 35

prateekmedia:
## Description

This PR fixes the build on main branch for photos mobile.


mobile/android/build.gradle
```diff
@@ -1,5 +1,6 @@
 ext {
-    appCompatVersion = '1.1.0' // for background_fetch
+    compileSdkVersion   = 35
+    targetSdkVersion    = 35
```
ua741, line 3: Can you confirm if targetSDK change really required?
From the background_fetch project docs, it looks like we don't need to add appCompactVersion pinning anymore.

But changing the targetSDK will require us to test the app for any change in behaviour.

https://developer.android.com/about/versions/15/behavior-changes-15

What was the previous targetSDK version which was coming via flutter?

ua741 (approved):

PR #5661, opened 2025-04-19
[mob][photos] Album UI Revamp

AmanRajSinghMourya:
## Description
This update introduces a redesigned user interface for album management, featuring a new layout for albums. Improve user experience by providing intuitive navigation and efficient batch operations.

## Tests
1. A new grid & list layout for the album screen. The first cell is dedicated to quickly adding new albums.
2. App Bar Enhancements
    - Sort Options: Organize albums by name, latest, or time.
    - View Toggle: Switch between grid and list views with ease.
    - Search Functionality: Quickly find albums using the search icon.
3. Long-press to select albums for batch actions. A bottom sheet provides options to delete, share, hide, or pin selected albums.


mobile/lib/ui/viewer/actions/album_selection_action_widget.dart
```diff
@@ -0,0 +1,171 @@
...
+      await updateOrder(
+        context,
+        collection,
+        collection.isPinned ? 1 : 1,
+      );
+    }
+    widget.selectedAlbums.clearAll();
+  }
+
+  Future<void> _onHideClick() async {
```
ua741, line 251: Certain actions like Hide or Pinning/Trash are only supported on albums that are owned by the user.
Also, we don't support trashing/hiding/pinning fav collections. Have we tested for these flows?

mobile/lib/ui/collections/album/list_item.dart
```diff
@@ -0,0 +1,171 @@
...
+    this.selectedAlbums,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final textTheme = getEnteTextTheme(context);
+    final colorScheme = getEnteColorScheme(context);
+    const sideOfThumbnail = 60.0;
+
+    return ListenableBuilder(
```
ashilkn: https://api.flutter.dev/flutter/widgets/ListenableBuilder-class.html 
Check out the "Performance optimizations" section here. I think it can be applied here.

mobile/lib/ui/collections/flex_grid_view.dart
```diff
@@ -68,4 +186,91 @@ class CollectionsFlexiGridViewWidget extends StatelessWidget {
...
+            if (result is Exception) {
+              await showGenericErrorDialog(context: context, error: result);
+            }
+          },
+          child: const NewAlbumListItemWidget(),
+        ),
+      );
+    }
+
+    if (widget.collections != null && widget.collections!.isNotEmpty) {
```
ashilkn: Can we not avoid this for loop and instead let the `SliverChildBuilderDelegate` do the iteration?

mobile/lib/ui/tabs/section_title.dart
```diff
@@ -45,25 +45,31 @@ class SectionOptions extends StatelessWidget {
...
-          ],
+      return GestureDetector(
+        behavior: HitTestBehavior.opaque,
+        onTap: onTap,
+        child: Container(
+          padding: padding,
+          child: Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: [
+              Container(alignment: Alignment.centerLeft, child: title),
```
ashilkn: Can we not use [Align](https://api.flutter.dev/flutter/widgets/Align-class.html) widget instead of a `Container` here?

mobile/lib/models/search/search_result.dart
```diff
@@ -15,5 +16,7 @@ abstract class SearchResult {
 
   List<EnteFile> resultFiles();
 
+  List<Collection> resultCollections();
```
ashilkn: Let's not add this abstract method. All sub classes will have to implement this but it's not _fully_ relevant to any of the current subclasses. It's only relevant to `GenericSearchResult` and that too only when `type()` is `ResultType.shared`. Alternate options are:
1.  Store `resultCollections` in `params` of `GenericSearchResult`, or
2. Create a new sub class for better separation called `ContactSearchResult` that extends `SearchResult` and there you can have a `resultCollections` getter.

Easiest would be option 1 and a much cleaner code would be option 2 which would require some refactoring.

mobile/lib/models/selected_albums.dart
```diff
@@ -0,0 +1,46 @@
...
+    }
+    notifyListeners();
+  }
+
+  void selectAll(Set<Collection> albumsToSelect) {
+    albums.addAll(albumsToSelect);
+    notifyListeners();
+  }
+
+  void unSelectAll(
```
ashilkn: I feel just `unSelect` would be a better name. At first glance just looking at the name, I expected that calling `SelectedAlbum.unSelectAll()` would just clear all selection. I understand you took the names from `SelectedFiles` but `unSelect` is definitely better.

mobile/lib/models/selected_albums.dart
```diff
@@ -0,0 +1,46 @@
...
+    );
+    if (alreadySelected != null) {
+      albums.remove(alreadySelected);
+    } else {
+      albums.add(albumToToggle);
+    }
+    notifyListeners();
+  }
+
+  void selectAll(Set<Collection> albumsToSelect) {
```
ashilkn: Same recommendation, better name would be `select`.

mobile/lib/services/search_service.dart
```diff
@@ -1413,10 +1414,30 @@ class SearchService {
...
+      final List<Collection> collections = _collectionService
+          .getCollectionsForUI(includedShared: true, includeCollab: true);
+
+      for (Collection collection in collections) {
+        if (collection.isHidden() ||
+            collection.isArchived() ||
+            collection.isOwner(ownerID)) {
+          continue;
+        }
+        if (collection.owner.id != ownerID) {
```
ashilkn: Isn't this check unnecessary since `collection.isOwner(ownerID)` is already checked in the `if` condition above?

mobile/lib/ui/actions/collection/collection_sharing_actions.dart
```diff
@@ -333,6 +333,91 @@ class CollectionActions {
...
+          buttonType: ButtonType.secondary,
+          buttonSize: ButtonSize.large,
+          buttonAction: ButtonAction.third,
+          shouldStickToDarkTheme: true,
+          isInAlert: true,
+        ),
+      ],
+      bodyWidget: StyledText(
+        text:
+            "Also delete the photos (and videos) present in these ${collections.length} albums from <bold>all</bold> other albums they are part of?",
```
ashilkn: Need to extract this string for translation. Not sure how to do it with bold tags in it.

mobile/lib/ui/collections/album/list_item.dart
```diff
@@ -0,0 +1,179 @@
...
+                    : colorScheme.strokeFainter,
+              ),
+              borderRadius: const BorderRadius.all(
+                Radius.circular(4),
+              ),
+            ),
+            child: Row(
+              mainAxisAlignment: MainAxisAlignment.spaceBetween,
+              children: [
+                Flexible(
```
ashilkn: This subtree (from this `Flexible`) doesn't depend on the listenable. Can improve performance by passing this subtree as `child` of `ListenableBuilder`. See https://api.flutter.dev/flutter/widgets/ListenableBuilder-class.html

mobile/lib/ui/collections/album/new_row_item.dart
```diff
@@ -0,0 +1,96 @@
...
+
+              try {
+                final Collection c =
+                    await CollectionsService.instance.createAlbum(text);
+                // ignore: unawaited_futures
+                await routeToPage(
+                  context,
+                  CollectionPage(CollectionWithThumbnail(c, null)),
+                );
+                Navigator.of(context).pop();
```
ashilkn: Isn't it possible to remove this `pop()` and set `popnavAfterSubmission: true` in `showTextInputDialog`?

mobile/lib/ui/collections/collection_list_page.dart
```diff
@@ -57,46 +74,195 @@ class _CollectionListPageState extends State<CollectionListPage> {
...
+      String text = key.toString();
+      switch (key) {
+        case AlbumSortKey.albumName:
+          text = S.of(context).name;
+          break;
+        case AlbumSortKey.newestPhoto:
+          text = "Date Created";
+          break;
+        case AlbumSortKey.lastUpdated:
+          text = "Date Modified";
```
ashilkn: Need to extract these strings for translation.

mobile/lib/ui/collections/collection_list_page.dart
```diff
@@ -57,46 +74,195 @@ class _CollectionListPageState extends State<CollectionListPage> {
...
+              if (selectedValue != null) {
+                sortKey = AlbumSortKey.values[selectedValue];
+                await localSettings.setAlbumSortKey(sortKey!);
+                albumSortDirection =
+                    albumSortDirection == AlbumSortDirection.ascending
+                        ? AlbumSortDirection.descending
+                        : AlbumSortDirection.ascending;
+                await localSettings.setAlbumSortDirection(albumSortDirection!);
+                await refreshCollections();
+                setState(() {});
```
ashilkn: If this `setState()` was intended only to rebuild widgets in `_sortMenu`, use a widget instead of a helper method.

mobile/lib/ui/collections/device/device_folders_vertical_grid_view.dart
```diff
@@ -66,8 +66,8 @@ class _DeviceFolderVerticalGridViewBodyState
   Width changes dynamically with screen width such that we can fit 2 in one row.
```
ashilkn, line 66: Remove/update this comment

mobile/lib/ui/collections/flex_grid_view.dart
```diff
@@ -1,17 +1,33 @@
...
 import "package:photos/ui/collections/album/row_item.dart";
+import "package:photos/ui/viewer/gallery/collection_page.dart";
+import "package:photos/utils/dialog_util.dart";
+import "package:photos/utils/local_settings.dart";
+import "package:photos/utils/navigation_util.dart";
 
-class CollectionsFlexiGridViewWidget extends StatelessWidget {
+class CollectionsFlexiGridViewWidget extends StatefulWidget {
   /*
   Aspect ratio 1:1 Max width 224 Fixed gap 8
```
ashilkn: Remove/update this comment

mobile/lib/ui/collections/flex_grid_view.dart
```diff
@@ -23,49 +39,255 @@ class CollectionsFlexiGridViewWidget extends StatelessWidget {
...
+
   @override
   Widget build(BuildContext context) {
+    return widget.albumViewType == AlbumViewType.grid
+        ? _buildGridView(context, const ValueKey("grid_view"))
+        : _buildListView(context, const ValueKey("list_view"));
+  }
+
+  Widget _buildGridView(BuildContext context, Key key) {
     final double screenWidth = MediaQuery.of(context).size.width;
```
ashilkn: Better to use more specific alternative `MediaQuery.sizeOf(context)`. See why [here](https://stackoverflow.com/questions/76474858/with-flutter-3-10-which-is-the-better-way-to-get-the-size-with-mediaquery).

FYI it doesn't make any performance differences here from a quick glance.

mobile/lib/ui/components/bottom_action_bar/album_action_bar_widget.dart
```diff
@@ -0,0 +1,84 @@
...
+        child: Row(
+          mainAxisAlignment: MainAxisAlignment.spaceBetween,
+          children: [
+            Flexible(
+              flex: 1,
+              child: ValueListenableBuilder(
+                valueListenable: _selectedAlbumNotifier,
+                builder: (context, value, child) {
+                  return Text(
+                    "${widget.selectedAlbums?.albums.length ?? 0} selected",
```
ashilkn: Need to extract string. Check what file selection bar is doing (I haven't checked)

mobile/lib/ui/components/bottom_action_bar/album_bottom_action_bar_widget.dart
```diff
@@ -0,0 +1,59 @@
...
+    this.sectionType, {
+    super.key,
+    this.backgroundColor,
+    this.onCancel,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final bottomPadding = MediaQuery.paddingOf(context).bottom;
+    final widthOfScreen = MediaQuery.of(context).size.width;
```
ashilkn: Use `MediaQuery.sizeOf(context)` have mentioned reason in a previous comment.

mobile/lib/ui/sharing/add_participant_page.dart
```diff
@@ -72,9 +84,11 @@ class _AddParticipantPage extends State<AddParticipantPage> {
       resizeToAvoidBottomInset: isKeypadOpen,
       appBar: AppBar(
         title: Text(
-          widget.isAddingViewer
-              ? S.of(context).addViewer
-              : S.of(context).addCollaborator,
+          widget.isMultipleCollections
+              ? "Add participants"
```
ashilkn: Extract string

mobile/lib/ui/sharing/add_participant_page.dart
```diff
@@ -20,10 +20,22 @@ import 'package:photos/ui/sharing/user_avator_widget.dart';
...
+  final Collection? collection;
   final bool isAddingViewer;
+  final List<Collection>? collections;
 
-  const AddParticipantPage(this.collection, this.isAddingViewer, {super.key});
+  const AddParticipantPage(
+    this.collection,
+    this.isAddingViewer, {
+    super.key,
+    this.collections,
```
ashilkn, line 34: Two similar instance variables `collection` and `collections`. Better to keep just `List<Collection>? collections`

mobile/lib/ui/sharing/add_participant_page.dart
```diff
@@ -20,10 +20,22 @@ import 'package:photos/ui/sharing/user_avator_widget.dart';
...
+  }) : assert(
+          collection != null || (collections != null),
+          'Either collection or non-empty collections must be provided',
+        );
+
+  bool get isMultipleCollections =>
+      collections != null && collections!.isNotEmpty;
```
ashilkn: This isn't making sense. If I'm not wrong, after making the change mentioned in the pervious comment, it should be 
`bool get isToMultipleCollections => (collections?.length ?? 0) > 1`

mobile/lib/ui/sharing/add_participant_page.dart
```diff
@@ -20,10 +20,22 @@ import 'package:photos/ui/sharing/user_avator_widget.dart';
 import "package:photos/ui/sharing/verify_identity_dialog.dart";
 
 class AddParticipantPage extends StatefulWidget {
```
ashilkn, line 28: Please go through this class again to make sure everything is working as expected and that there's no unnecessary code. Let me know once you're done, I'll review it again.

mobile/lib/ui/tabs/user_collections_tab.dart
```diff
@@ -121,175 +128,109 @@ class _UserCollectionsTabState extends State<UserCollectionsTab>
...
-              }
-            },
-            child: const IconButtonWidget(
-              icon: Icons.sort_outlined,
-              iconButtonType: IconButtonType.secondary,
+              ),
+            ),
+            SliverToBoxAdapter(
+              child:
+                  SizedBox(height: 64 + MediaQuery.of(context).padding.bottom),
```
ashilkn: Please change this to use `MediaQuery.paddingOf(context).bottom`

mobile/lib/ui/viewer/actions/album_selection_action_widget.dart
```diff
@@ -0,0 +1,372 @@
...
+  @override
+  void dispose() {
+    widget.selectedAlbums.removeListener(_selectionChangedListener);
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    if (widget.selectedAlbums.albums.isEmpty) {
+      return const SizedBox();
```
ashilkn: `SizedBox.shrink()` is recommended in such cases.

mobile/lib/ui/viewer/actions/album_selection_action_widget.dart
```diff
@@ -0,0 +1,372 @@
...
+
+    for (final collection in widget.selectedAlbums.albums) {
+      count = await FilesDB.instance.collectionFileCount(collection.id);
+      final bool isEmptyCollection = count == 0;
+      if (isEmptyCollection) {
+        try {
+          await CollectionsService.instance.trashEmptyCollection(collection);
+        } catch (e, s) {
+          _logger.warning("failed to trash collection", e, s);
+          await showGenericErrorDialog(context: context, error: e);
```
ashilkn: If trashing fails for multiple empty collections, this will result in multiple back to back dialogs.

mobile/lib/ui/viewer/actions/album_selection_action_widget.dart
```diff
@@ -0,0 +1,372 @@
...
+        }
+      } else if (collection.type == CollectionType.favorites) {
```
ashilkn: What if the favourite collection is in selected list and is empty? Will `trashEmptyCollection` trash the favourites album? @ua741  Do we have some check on the backend that stops this?

Or is this fine since a new favourites collection is created on the client side if favourite collection does not exist?

Regardless, we should move the 'is favourites collection' check to the beginning of the loop since the rule, based on the usage of `_showFavToast` seems to be that we don't allow any actions on the favourites collection."

AmanRajSinghMourya: for this we can show a toast `_showFavToast`. if user try to delete favourite album

mobile/lib/ui/viewer/actions/album_selection_action_widget.dart
```diff
@@ -0,0 +1,372 @@
...
+      hasFavorites = widget.selectedAlbums.albums
+          .any((album) => album.type == CollectionType.favorites);
+      setState(() {});
+    }
+  }
+
+  void _showFavToast() {
+    showShortToast(
+      context,
+      "The Favorites album cannot be modified",
```
ashilkn: Let's change the string to "Action not supported on Favourites album". Plus, extract the string.

mobile/lib/ui/sharing/album_participants_page.dart
```diff
@@ -54,7 +54,7 @@ class _AlbumParticipantsPageState extends State<AlbumParticipantsPage> {
   Future<void> _navigateToAddUser(bool addingViewer) async {
     await routeToPage(
       context,
-      AddParticipantPage(widget.collection, addingViewer),
+      AddParticipantPage(widget.collection as List<Collection>, addingViewer),
```
ashilkn: This will throw an error because it's trying to cast an `int` to a `List<int>`
Just pass `[widget.collection]` instead.

mobile/lib/ui/viewer/actions/album_selection_action_widget.dart
ashilkn: > for this we can show a toast `_showFavToast`. if user try to delete favourite album

Yes, let's do this. Let's move the `collection.type == CollectionType.favorites` check before `isEmptyCollection` check to handle the case where favourite collection is empty. And then call `_showFavToast` in the `collection.type == CollectionType.favorites` if condition.

ashilkn (approved):

ashilkn (approved):

PR #5662, opened 2025-04-19
[cli] Implement multithreaded downloading of files in export

apprehensions:
## Description

Using [Golang Semaphores](golang.org/x/sync/semaphore), i was able to implement multithreaded downloading of files during the export process of files.

## Functional changes

1. A cache is in place for the retrieval of an `albumDiskInfo`, rather than using a pointer. This is made due to the nature of goroutines.
2. `log.Fatal` is called upon failure of a download, where the previous behavior was to simply return the error. This is intentional, since otherwise, there would be two ways to forward the error, such as errgroups or cancelling the context.
3. A new log is printed when a download finishes.

This speeds up the download process **significantly!** The jobs count out of the box is 12, but i feel like it should be increased too.

## Tests

None to be implemented, since this is on the user's side

### DIFF
--- a/cli/cmd/export.go
+++ b/cli/cmd/export.go
@@ -16,12 +16,14 @@ var exportCmd = &cobra.Command{
 		hidden, _ := cmd.Flags().GetBool("hidden")
 		albums, _ := cmd.Flags().GetStringSlice("albums")
 		emails, _ := cmd.Flags().GetStringSlice("emails")
+		jobs, _ := cmd.Flags().GetInt64("jobs")
 		// Create Filters struct with flag values
 		filters := model.Filter{
 			ExcludeShared: !shared,
 			ExcludeHidden: !hidden,
 			Albums:        albums,
 			Emails:        emails,
+			Jobs:          jobs,
 		}
 		// Call the Export function with the filters
 		ctrl.Export(filters)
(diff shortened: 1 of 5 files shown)


ua741:
We will also need to handle concurrency related issues inside `downloadEntry` as we read the existing disk info to identify file names and bunch of other things what would need some changes.

apprehensions:
I don't see how that's a problem.

ua741:
> I don't see how that's a problem.
If multiple files in an album have same name, we can potentially end up with incorrect/corrupted export.
You can try uploading say 12 files with same name but different content in a single album, and try to export that.

apprehensions:
https://github.com/apprehensions/ente/blob/main/cli/pkg/remote_to_disk_file.go#L203

The file duplication check is in here. I assume you mean that there can be a case where two duplicate files will both think there is no duplicate since they're checked at the same time? If so, then the only change that has to be made would be to move the checks before downloading anything, that way, all a routine has to do is download or delete, not rename or check for existing files. This is beyond the PR's scope, and is a design decision. Apologies if I misunderstood.

ua741:
> https://github.com/apprehensions/ente/blob/main/cli/pkg/remote_to_disk_file.go#L203
> 
> The file duplication check is in here. I assume you mean that there can be a case where two duplicate files will both think there is no duplicate since they're checked at the same time? If so, then the only change that has to be made would be to move the checks before downloading anything, that way, all a routine has to do is download or delete, not rename or check for existing files. This is beyond the PR's scope, and is a design decision. Apologies if I misunderstood.


Yes, the current design was to incorporate nuances related to naming of Live photos split, where we don't know the full name of live photo video before we decrypt the zip file that contains more information.

Although, it's not in scope of this PR, unfortunately, we can't merge these changes without handling those nuaunces as it will break existing functionality.

sewnie:
Have these nuances been fixed yet? Exporting quickly and efficiently is quite important to me.

PR #5743, opened 2025-04-28
[mob] isolated ffmpeg run

prateekmedia:
## Description

- [x] Create a new service to use for FFmpeg in an isolate'ed environment
- [ ] Apply this service wherever we use FFmpeg Kit

## Tests

- [ ] Test FFmpeg intergrations


mobile/lib/services/isolated_ffmpeg_service.dart
```diff
@@ -0,0 +1,12 @@
...
+
+class IsolatedFfmpegService {
+  static Future<FFmpegSession> ffmpegRun(Map args) async {
+    final command = args['command'] as String;
+
+    final session = await FFmpegKit.execute(command);
+
+    return session;
+  }
+}
```
ua741: Instead of using shared computer/isolate, let's use a dedicated computer instance for ffmpeg so that it doesn't block one of the shared computer for long time.
Also, we can simplify the call by adding a helper method.
```dart
class IsolatedFfmpegService {
  final Computer _computer = Computer.create();
  void init() {
    _computer.turnOn(workersCount: 1);
  }
  Future<FFmpegSession> runFfmpeg(String command) async {
    return _computer.compute<Map, FFmpegSession>(
      _ffmpegRun,
      param: {'command': command},
    );
  }

  static Future<FFmpegSession> _ffmpegRun(Map args) async {
    final command = args['command'] as String;
    final session = await FFmpegKit.execute(command);
    return session;
  }
}
```

ua741: If you want to init ffmpeg inside that isolate/worker, that can either work with this computer with single worker that will ensure that the ffmpeg is initialized in that isolate or start another isolate and manage it ourself similar to MLComputer

mobile/lib/main.dart
```diff
@@ -212,6 +213,8 @@ Future<void> _init(bool isBackground, {String via = ''}) async {
     Computer.shared().turnOn(workersCount: 4).ignore();
     CryptoUtil.init();
 
+    IsolatedFfmpegService.init();
```
ua741: Let's init it right before processing any video to avoid creating redundant isolate.

mobile/apps/photos/lib/services/video_preview_service.dart
```diff
@@ -303,13 +305,22 @@ class VideoPreviewService {
           objectSize = result.$2;
 
           // Fetch resolution of generated stream by decrypting a single frame
-          final FFmpegSession session2 = await FFmpegKit.execute(
+          final map2 = await IsolatedFfmpegService.runFfmpeg(
```
ua741: should. we rename this?

PR #5749, opened 2025-04-29
Copy --link

adduty:
## Description
`--link` is desirable in almost all situations where `COPY` is used due to caching improvements. See https://docs.docker.com/reference/dockerfile/#benefits-of-using---link and https://docs.docker.com/reference/dockerfile/#incompatibilities-with---linkfalse. The syntax parser directive is used to pin the Dockerfile syntax for BuildKit. See https://docs.docker.com/reference/dockerfile/#syntax.

## Tests
Have confirmed all of the images build successfully, with the exception of cli/Dockerfile-x86, which is broken for unrelated reasons. I assume there are some tests defined in the Actions workflows, but will run tests locally if needed.

### DIFF
--- a/cli/Dockerfile
+++ b/cli/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 FROM golang:1.23-alpine3.21 AS builder
 RUN apk add --no-cache gcc musl-dev git build-base pkgconfig libsodium-dev
 
--- a/cli/Dockerfile
+++ b/cli/Dockerfile
@@ -5,18 +5,18 @@ ENV GOOS=linux
 
 WORKDIR /etc/ente/
 
-COPY go.mod .
-COPY go.sum .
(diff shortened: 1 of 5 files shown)


mnvr:
Thank you. From my cursory understanding, `--link` seems to allow reusing older layers even when the intervening (unrelated to the files that are being copied) layers change. However, (a) there come caveats around symlink, and I cannot exhaustively rule out issues from a cursory glance, but right now I'm unable to revisit all the files to make sure there is no correctness issue, (b) the gains, if any, would be marginal at best for our specific cases, (c) and increases the volume of code (without a clear cut benefit for our specific case).

For now, I'll close this PR. But I'd (or someone else from the team) will be happy to revisit this at a later point, say when reworking on one of these files for some other functional change - at that times when these directives can be introduced piecewise in a more controlled and vetted manner.

Again, thanks for the PR, and it is not great that it had to be closed.

PR #5756, opened 2025-04-29
Add Bitazza icon

eark39:
Add Bitazza icon


ua741:
> File name is not lowercase: assets/custom-icons/icons/Bitazza.svg

PR #5767, opened 2025-04-30
[mob][photos] Faces queue

laurenspriem:
## Description

Use custom task queue instead of pool package for face thumbnail generation

## Tests

### DIFF
--- a/mobile/lib/utils/face/face_box_crop.dart
+++ b/mobile/lib/utils/face/face_box_crop.dart
@@ -1,3 +1,4 @@
+import "dart:async";
 import "dart:io" show File;
 
 import "package:flutter/foundation.dart";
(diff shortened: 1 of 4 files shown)


mobile/lib/utils/face/face_box_crop.dart
```diff
@@ -30,10 +37,16 @@ final _logger = Logger("FaceCropUtils");
 const int _retryLimit = 3;
 final LRUMap<String, Uint8List?> _faceCropCache = LRUMap(1000);
 final LRUMap<String, Uint8List?> _faceCropThumbnailCache = LRUMap(1000);
-Pool _poolFullFileFaceGenerations =
-    Pool(20, timeout: const Duration(seconds: 15));
-Pool _poolThumbnailFaceGenerations =
-    Pool(100, timeout: const Duration(seconds: 15));
+TaskQueue _queueFullFileFaceGenerations = TaskQueue<String>(
+  maxConcurrentTasks: 20,
```
ua741: May be 5 concurrent tasks should be more than enough. 20 feels too much.

laurenspriem: I've lowered it to 5. Hard to say from testing if it made things better or worse but you're right that it's a safer start.

ua741 (approved):

PR #5770, opened 2025-04-30
[staff] show family member storage quota

mngshm:


mngshm:
The linters are failing for a component file. I've tried fixing that locally, but it particularly doesn't fix for that file and the same commands locally cannot find any issues in that file. I'm assuming this should be fixed once #5765 is merged. Will re-run the linter jobs after merge.

infra/staff/src/components/FamilyComponentTable.tsx
```diff
@@ -134,7 +138,11 @@ const FamilyTableComponent: React.FC = () => {
                                 <TableCell>
                                     {formatUsageToGB(member.usage)}
                                 </TableCell>
-                                <TableCell>{member.id}</TableCell>
+                                <TableCell>
+                                    {member.status !== "SELF"
+                                        ? formatUsageToGB(member.usage)
```
ua741: This seems incorrect if we want to show quota/limit?

mngshm: [dd0cc71](https://github.com/ente-io/ente/pull/5770/commits/dd0cc71f365274f357b3a385704a96f219d7eda2)

Oops, fixed it. Its supposed to be storageLimit.

infra/staff/src/components/FamilyComponentTable.tsx
```diff
@@ -134,7 +138,11 @@ const FamilyTableComponent: React.FC = () => {
                                 <TableCell>
                                     {formatUsageToGB(member.usage)}
                                 </TableCell>
-                                <TableCell>{member.id}</TableCell>
+                                <TableCell>
+                                    {member.status !== "SELF"
+                                        ? formatUsageToGB(member.storageLimit)
```
ua741: What happens when there is no storage limit?

mngshm: image

Show's 0 at the moment because formatBytesToGb (or typescript) coalesces null (coming from the request response) to 0. Want me to handle this with some different UX?

ua741: Yes, let's show that as NA. Can mark that field as nullable and handle this.

mngshm: image

fixed in [be0612b](https://github.com/ente-io/ente/pull/5770/commits/be0612bc9184ac5e5ea95ac4d21f2319df55c4fd)

ua741 (approved):

PR #5772, opened 2025-04-30
[server] Render html responses for account recovery

ua741:


server/Dockerfile
```diff
@@ -19,6 +19,7 @@ COPY --from=builder /etc/ente/museum .
 COPY configurations configurations
 COPY migrations migrations
 COPY mail-templates mail-templates
+COPY templates templates
```
vishnukvmd: `web-templates` would be clearer?

vishnukvmd (approved):

PR #5792, opened 2025-05-03
[mob][photos] Deletion fixes

laurenspriem:
## Description

- Fix issue where user is not able to delete own files in a shared album
- Fix issue where deletion of not yet uploaded files leads to showing grey boxes in gallery

## Tests

### DIFF
--- a/mobile/lib/models/gallery_type.dart
+++ b/mobile/lib/models/gallery_type.dart
@@ -105,7 +105,6 @@ extension GalleyTypeExtension on GalleryType {
     switch (this) {
       case GalleryType.ownedCollection:
       case GalleryType.sharedCollection:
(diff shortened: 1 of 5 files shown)


mobile/lib/models/gallery_type.dart
```diff
@@ -104,6 +104,8 @@ extension GalleyTypeExtension on GalleryType {
   bool showDeleteOption() {
     switch (this) {
       case GalleryType.ownedCollection:
+      case GalleryType.sharedCollection:
+      case GalleryType.sharedPublicCollection:
```
ua741: We need to hide delete option from `sharedPublicCollection`:

ua741 (approved):

PR #5796, opened 2025-05-04
Check is special choose in added

Zamanhuseyinli:
## Description
The identification mechanism based on prediction can now predict what kind of emotion the user is expressing in the photo or whether it is an animal, an object or a human according to any photo or x-y-y position value, but with considerable development, this prototype can be moved to the active-learning level with user feedback and the algorithm structure can be expanded or multiplied, or with the check here, the prediction can be adjusted according to the user's photo number value.
## Tests
I am currently testing it on my own device, Redmi Note 8 Pro. It may be more functional soon.

### DIFF
--- a/mobile/lib/models/ml/face/box.dart
+++ b/mobile/lib/models/ml/face/box.dart
@@ -2,38 +2,43 @@ import "package:photos/utils/standalone/parse.dart";
 
 /// Bounding box of a face.
 ///
-/// [ x] and [y] are the minimum coordinates, so the top left corner of the box.
-/// [width] and [height] are the width and height of the box.
-///
-/// WARNING: All values are relative to the original image size, so in the range [0, 1].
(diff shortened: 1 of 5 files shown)


ua741:
As this a change related to ML pipeline, I would request you to please start a discussion with proposed change & intension behind this change. We will need to sync any changes across devices & various clients, ensure it doesn't break existing clients.

Zamanhuseyinli:
I will develop it in a way that will be synchronized, by changing or modifying the necessary solution and pipeline structure between other devices, the process of making more predictions will be more understandable with check and by determining more self-training, user situations or any object can be determined directly in the process, I will add a feedback process for AI definition, and for users who do not want to use it, it will be possible to segment or detect photos or videos directly based on AI results and divide them into layers accordingly.

PR #5804, opened 2025-05-05
[quickstart] Gracefully handle case when docker compose is not present

mnvr:
When docker is present but docker compose is not present, the `docker compose` invocation would fail. We want the early exit (`set -e`), so instead do a fallback to set dcv to an empty string so that it later fails in the `test -z dcv` case below and prints the intended error message.


mngshm (approved):
LGTM, tested it locally

PR #5828, opened 2025-05-06
[quickstart] add guidelines on HTTPS museum usage

wrobelda:
## Description
Setting up Museum over the HTTPS for self-hosting is actually very easy, but requires 3 extra steps. This adds necessary but commented out parameters, with extra instructions, as well some extra docs content to the `quickstart.md` document — although I do not see it linked anywhere on https://help.ente.io ?

### DIFF
--- a/server/docs/quickstart.md
+++ b/server/docs/quickstart.md
@@ -48,7 +48,7 @@ open ports:
 
 | Service     | Port     | Notes                   |
 | ----------- | -------- | ----------------------- |
-| museum      | `:8080`  | Ente's API server       |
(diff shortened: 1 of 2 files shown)


mnvr:
There are two different problems:
1. Mentioning somewhere in the docs the steps that make it easier for someone to setup TLS with lets's encrypt (the original intent of the PR)
2. Refactoring the docs 

I don't disagree that 2 can be done better, but it is a more involved change that also needs input from the other people who are currently maintaining the docs, and I don't want to go about unilaterally changing the doc structure. 

Just 1 is a good thing meanwhile - if we want those to get there in isolation, we can do what was suggested in https://github.com/ente-io/ente/pull/5828#pullrequestreview-2818312238.

PR #5892, opened 2025-05-13
[mob] Cache video streams for shared files

ua741:
## Description
Title

## Tests
- Verified from logs that shared stream videos are getting cached correctly.

### DIFF
--- a/mobile/lib/services/preview_video_store.dart
+++ b/mobile/lib/services/preview_video_store.dart
@@ -107,7 +107,7 @@ class PreviewVideoStore {
 
   Future<bool> isSharedFileStreamble(EnteFile file) async {
     try {
-      await getPreviewUrl(file);
+      await _getPreviewUrl(file);
       return true;
     } catch (_) {
       return false;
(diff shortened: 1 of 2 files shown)


mobile/lib/services/preview_video_store.dart
```diff
@@ -529,16 +530,23 @@ class PreviewVideoStore {
...
-          ? null
-          : await cacheManager.getFileFromCache(
-              _getDetailsCacheKey(objectKey),
-            );
+      late final String objectKey;
+      final PreviewInfo? previewInfo =
+          FileDataService.instance.previewIds?[file.uploadedFileID!];
+      (String, String)? previewURLResult;
+      if (previewInfo == null) {
+        previewURLResult = await _getPreviewUrl(file);
```
prateekmedia, line 538: Can we do an additional check to only run this condition when file is not owned?

ua741: I didn't see any particular reason for adding that check. Previously, if the file data information was missing, we were fetching and previewing it without caching anything as cacheID was null.

Let me know if I am missing some context.

prateekmedia (approved):
LGTM!

PR #5905, opened 2025-05-14
[mob] Show person's name for shared fav collection

ua741:


laurenspriem (commented):
Can we get rid of the 'ignore: deprecated...' comment by using collectionName instead of name?

ua741:
> Can we get rid of the 'ignore: deprecated...' comment by using collectionName instead of name?

That's still required because we are using `name` field that's marked as deprecated.

laurenspriem:
The deprecated note says to use `collectionName` instead, can't we use that?

laurenspriem (approved):

PR #5908, opened 2025-05-14
[mob] Fix: Show switch stream options for shared video files

ua741:
## Description
- This change also fixes the issue where switch stream option wasn't available when local video streaming was disabled.
- Also cleaned up the widget by removing queue/processing relate state from the widget.


prateekmedia (approved):
LGTM!

PR #5999, opened 2025-05-22
[mob] Trust the user trust store (Android)

denysvitali:
## Description

It's currently not possible to use a server that has a CA that is trusted by the _user_ trust store in Android.  
This is generally due  to a missing configuration of the `networkSecurityConfig` (by default only system-level CAs are accepted).  
  
With this change, the app should be able to reach out to servers that do not use a publicly trusted CA (but that are still trusted by the user).

## Tests

None

## Remarks

I wasn't able to test my changes as the Flutter build fails with:

```
warning: [options] source value 8 is obsolete and will be removed in a future release
warning: [options] target value 8 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
$HOME/.pub-cache/git/ffmpeg-kit-6d5d27a8c259eda6292f204a27fba53da70af20e/flutter/flutter/android/src/main/java/com/arthenica/ffmpegkit/flutter/FFmpegKitFlutterPlugin.java:157: error: cannot find symbol
    public static void registerWith(final io.flutter.plugin.common.PluginRegistry.Registrar registrar) {
                                                                                 ^
  symbol:   class Registrar
  location: interface PluginRegistry
$HOME/.pub-cache/git/ffmpeg-kit-6d5d27a8c259eda6292f204a27fba53da70af20e/flutter/flutter/android/src/main/java/com/arthenica/ffmpegkit/flutter/FFmpegKitFlutterPlugin.java:651: error: cannot find symbol
    protected void init(final BinaryMessenger messenger, final Context context, final Activity activity, final io.flutter.plugin.common.PluginRegistry.Registrar registrar, final ActivityPluginBinding activityBinding) {
                                                                                                                ^
  symbol:   class Registrar
  location: interface PluginRegistry
2 errors
3 warnings

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':ffmpeg_kit_flutter:compileDebugJavaWithJavac'.
> Compilation failed; see the compiler error output for details.

* Try:
> Run with --info option to get more log output.
> Run with --scan to get full insights.

BUILD FAILED in 59s
```


mobile/android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,8 @@
+<network-security-config>
+    <base-config>
+        <trust-anchors>
+            <certificates src="system" />
+            <certificates src="user" />
+        </trust-anchors>
+    </base-config>
```
ua741, line 7: Can you also add domain-config for ente, to ensure that only system certificates are trusted for our production instance.
```xml
<domain-config cleartextTrafficPermitted="false">
        <domain includeSubdomains="true">ente.io</domain>
        <trust-anchors>
            <!-- Only trust pre-installed CAs for ente production domains-->
            <certificates src="system" />
        </trust-anchors>
</domain-config>
```

ua741 (approved):

mobile/android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,14 @@
...
+    <domain-config cleartextTrafficPermitted="false">
```
prestonvanloon, line 8: Seems like this might have broken self hosted without SSL. https://github.com/ente-io/ente/issues/6186

denysvitali: This might be because the default is `cleartextTrafficPermitted="false"` (not on this specific line). I'll create another PR to fix the regression

denysvitali: See #6201 - it _should_ fix the issue. Sorry about that!

Lense:
Because the underlying HTTP library doesn't support the user cerficate store (https://github.com/dart-lang/sdk/issues/50435), I don't believe this PR fixes anything.

The following screenshot shows photos 1.1.0 raising an error with a trusted certificate (which works in web browsers):

entecert

The same error was raised before and after this PR. Thank you for the change of printing the error message; I had to build a patched APK to see it before.

Because Ente/Flutter/Dio/Dart don't support user certificates, I've been using a plaintext HTTP connection, but this PR broke that too. I would really appreciate a Play Store release with #6201.

It would be great to actually make user certificates work in Ente, but that seems infeasible with the tech stack.

jfly:
> It would be great to actually make user certificates work in Ente, but that seems infeasible with the tech stack.

@Lense, this is possible, it's just annoying. See my PR to Finamp (another Dart/Flutter app) showing how to make it work: https://github.com/jmshrv/finamp/pull/1276.

PR #6104, opened 2025-05-31
Add 5 new icons and rename existing icon

alewi7:
## Description

Added 5 new custom SVG icons:

- **cronometer.svg**
- **xai.svg**
- **toshl_finance.svg**
- **aternos.svg**
- **atomic_mail.svg**

Also renamed an existing icon **'mercado_livre.svg'** to **'mercado_libre.svg'** to match the brand's official name.

## Tests

No formal tests added. Verified manually that all SVGs open correctly in the browser and display as expected.


ua741:
@alewi7 Thank you for your contribution. It seems like the json is invalid which is causing lint failure
> jq: parse error: Unmatched ']' at line 1406, column 7

PR #6106, opened 2025-05-31
[server] Check in with paid customers

vishnukvmd:
## Description
Update the `subscriptions` table to store the time at which subscriptions were upgraded (`upgraded_at`).

Reach out to customers who upgraded 7 days ago to make sure all is well.
Screenshot 2025-05-31 at 1 41 09 PM

Store this information within `notification_history` so they are not contacted again (in case of admin interventions).

> Note: We will not be back-filling data for existing subscriptions.

## Tests
- [x] Tested locally to make sure only customers who upgraded 7 days ago were pinged.

### DIFF
--- a/server/pkg/repo/user.go
+++ b/server/pkg/repo/user.go
@@ -332,6 +332,35 @@ func (repo *UserRepository) GetUsersWithIndividualPlanWhoHaveExceededStorageQuot
 	return users, nil
 }
 
+func (repo *UserRepository) GetUsersWhoUpgradedNDaysAgo(days int) ([]ente.User, error) {
+	rows, err := repo.DB.Query(`
+        SELECT u.user_id, u.encrypted_email, u.email_decryption_nonce, u.email_hash, u.creation_time 
+        FROM users u
+        INNER JOIN subscriptions s ON u.user_id = s.user_id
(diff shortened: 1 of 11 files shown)


grittypuffy (approved):
Looks good to me since it has a chance of improving retention.

Self-hosters are less likely to use subscriptions, so there should be no risk of sending mails to them.

server/pkg/repo/user.go
```diff
@@ -332,6 +332,35 @@ func (repo *UserRepository) GetUsersWithIndividualPlanWhoHaveExceededStorageQuot
 	return users, nil
 }
 
+func (repo *UserRepository) GetUsersWhoUpgradedNDaysAgo(days int) ([]ente.User, error) {
```
ua741, line 335: We we also pass the template detail here and use it to filter out users for whom we have already sent email. This way, we will avoid redundant decryption and look up for users for whom we have already sent the email.

vishnukvmd: Discussed offline – the compute involved will not be much since we are only looking at the customers who upgraded exactly 7 days ago (not 7+ days ago). So will let this be to keep things simple.

PR #6145, opened 2025-06-02
[web][desktop] Routine dependency updates

mnvr:


mnvr:
ONNX runtime update failed because of this - https://github.com/microsoft/onnxruntime/issues/24918

PR #6152, opened 2025-06-03
[mob][photos] Memories improvement

AmanRajSinghMourya:
## Description
1. A subtle zoom-in/out effect for photos and replaced the black area around landscape photos with a blurred background.  
2. Auto play each image for 5 second and video for its duration with a step progress animation. 
3. Long-press to pause animation. Releasing will resume the playback.


mobile/lib/ui/home/memories/full_screen_memory.dart
```diff
@@ -450,3 +556,150 @@ class BottomGradient extends StatelessWidget {
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return widget.isVideo
+        ? widget.child
+        : AnimatedBuilder(
+            animation: _controller,
+            builder: (context, child) {
+              return ClipRect(
```
ashilkn: Is this `ClipRect` required?

mobile/lib/ui/home/memories/full_screen_memory.dart
```diff
@@ -131,6 +136,10 @@ class FullScreenMemory extends StatefulWidget {
 class _FullScreenMemoryState extends State<FullScreenMemory> {
   PageController? _pageController;
   final _showTitle = ValueNotifier<bool>(true);
+  AnimationController? _progressAnimationController;
+  AnimationController? _kenBurnsAnimationController;
```
ashilkn: Dispose these controllers if possible

AmanRajSinghMourya: These are disposed in there respective child widgets

mobile/lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -426,6 +428,9 @@ class _ZoomableImageState extends State<ZoomableImage> {
       _loadedFinalImage = true;
       _logger.info("Final image loaded");
     });
+    if (_imageProvider != null) {
```
ashilkn: Is this check necessary?

mobile/lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -426,6 +428,9 @@ class _ZoomableImageState extends State<ZoomableImage> {
...
+      widget.onFinalFileLoad?.call(5);
```
ashilkn: Use a named parameter for duration here. From my understanding of what '5' is used for, something like `memoryDuration` works. Using a named parameter will make it clear what this '5' is used for.

mobile/lib/ui/viewer/file/zoomable_image.dart
AmanRajSinghMourya: Just to make sure that we don't play the animation until we have the imageProvider

PR #6176, opened 2025-06-05
[mob][photos] Album UI improvements

AmanRajSinghMourya:
### Changes

1. Album corner radius  (114 size - 12px corner radius) (100 size - 8px corner radius)
2. Thumbnail view size 60 -  corner radius 4
3. Padding between album thumbnail and Album name is 6
4. Padding between Album name and number of photos is 2
5. Album name (text size) - Small regular , Full black (color)
6. Number of photos (text size) - Mini Regular  , Text muted (color)
7. New design for Add album thumbnail
8. Distance between album vertically & horizontally 8px
9. List view border radius 6px and thumbnail radius 4px


mobile/lib/ui/collections/collection_action_sheet.dart
```diff
@@ -323,15 +340,25 @@ class _CollectionActionSheetState extends State<CollectionActionSheet> {
...
   Future<List<Collection>> _getCollections() async {
     if (_showOnlyHiddenCollections) {
+      final List<Collection> recentlyCreated = [];
+      final List<Collection> hidden = [];
+
       final hiddenCollections = CollectionsService.instance
           .getHiddenCollections(includeDefaultHidden: false);
-      hiddenCollections.sort((first, second) {
+      for (final collection in hiddenCollections) {
+        if (_recentlyCreatedCollections.contains(collection)) {
```
ashilkn, line 349: This check is unnecessary (let me know if I'm wrong). If `_showOnlyHiddenCollections` is true, `CollectionActionType` is either `moveToHiddenCollection` or `addToHiddenAlbum`. And this is the same condition when new hidden albums are created in `AlbumVerticalListWidget`. So it's safe to assume that all collections in `_recentlyCreatedCollections` are hidden collections here.

ashilkn: Unrelated to your changes but a minor refactor that helps:
Create a method or an extension method for the enum `CollectionActionType` called `isHiddenAction` and use that instead of checking using:
```
 if (widget.actionType == CollectionActionType.moveToHiddenCollection ||
          widget.actionType == CollectionActionType.addToHiddenAlbum)
```

This is currently used in two places. It's safer to refactor this to a single method and use that instead since it deals with hidden collections.

mobile/lib/ui/collections/collection_action_sheet.dart
```diff
@@ -362,13 +390,20 @@ class _CollectionActionSheetState extends State<CollectionActionSheet> {
           }
           continue;
         }
+        if (_recentlyCreatedCollections.contains(collection)) {
+          recentlyCreated.add(collection);
+          continue;
+        }
```
ashilkn, line 393: Why is this necessary? Will `_recentlyCreatedCollections` contain any collection that are not in list of collections returned by `getCollectionsForUI`?

PR #6182, opened 2025-06-05
[photos][mob] Deselect albums with back navigation in second tab

JacquesLucke:
## Description

Previously, back navigation would always go to the home screen, even when albums are currently selected. This is unexpected because typically back navigation first deselect everything if anything is selected.

This patch implements this expected behavior for albums on the second tab. If at least one album is selected, all will be deselected when going-back. Only when everything is deselect, the app transitions to the first tab like before.


ashilkn (approved):

ashilkn:
Thanks for the contribution @JacquesLucke!

PR #6203, opened 2025-06-07
[photos][mob] Allow renaming uploader

JacquesLucke:
## Description

When collecting photos from friends, they sometimes enter a wrong name. Currently, there does not seem to be a way to correct this after the upload which is unfortunate. This patch adds a way to rename an uploader.

The main tricky aspect is the UI because currently the name is mainly displayed when looking at the info of a single file but when renaming one generally wants to rename the uploader of all files with the same name.

Right now, this patch just makes it possible to tap on the `Added by X` label to rename the uploader (this is where I attempted to change the name first). It's not ideal because (1) there is no indication that it is editable and (2) it's in the file info of a single file while changing the uploader name of potentially many files.

It still seems to be good enough for the most common use-cases when collecting photos and someone types a wrong name. 

## Tests

https://github.com/user-attachments/assets/0a9f2e0c-669e-4f63-ba6f-b5c4776d918d

----

There is an older discussion thread for this topic already which didn't have a response yet: https://github.com/ente-io/ente/discussions/585

I also agree with the second part of the mentioned problem that the dialog where people should enter their name is not very clear. At a quick glance it was not obvious to me what name should be written there (file name, collection name, author name, ...). Maybe I could look into improving that separately if desired.

### DIFF
--- a/mobile/lib/generated/l10n.dart
+++ b/mobile/lib/generated/l10n.dart
@@ -6026,6 +6026,16 @@ class S {
     );
   }
 
+  /// `Rename uploader`
+  String get renameUploader {
+    return Intl.message(
+      'Rename uploader',
+      name: 'renameUploader',
+      desc: '',
+      args: [],
+    );
+  }
+
   /// `Leave shared album?`
   String get leaveSharedAlbum {
     return Intl.message(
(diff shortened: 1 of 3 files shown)


JacquesLucke:
Will create a discussion thread for this first now :)

PR #6208, opened 2025-06-08
[photos][web] Improve placeholder name when requesting uploader name

JacquesLucke:
As mentioned in #585 and #6203, the dialog requesting the uploader name is a bit ambiguous currently, especially at a quick glance when just reading the headline and placeholder (the other sentence is fine but feels a bit verbose and is easily skimmed over in practice).

I had someone fill in a name for the uploaded photo and the author of #585 had people fill in the name of the collection. Both are not what's expected.

This patch implements a simple wording change that hopefully makes these kinds of mistakes less likely. It changes the placeholder from `Name...` to `Your name...`. In theory the title of the dialog could be changed too, but maybe it's enough to change it in just one place.

As part of the patch I changed the English and German translation but not the others.

Before:
image

After:
image


mnvr:
> For anything beyond trivial bug fixes, please use [discussions](https://github.com/ente-io/ente/discussions) instead of performing code changes directly.
> 
> https://github.com/ente-io/ente/blob/main/CONTRIBUTING.md#code-contributions

JacquesLucke:
Added some thoughts to the existing discussion thread: #585.

PR #6257, opened 2025-06-12
[docs][script] update quickstart to provide choice to start cluster

grittypuffy:
## Description
- This PR updates `quickstart.sh` to prompt user to start cluster, which acts as a way to modify default configuration.
- Further, indication of usage of MinIO web console is provided in the `quickstart.sh` in commented form, allowing users to administer the object storage using web based interface.
- Updation of requirements to provide more clarity in self-hosting documentation
- Uncommenting environment variables for modification in web service in quickstart script

## Tests
- [x] I have tested these changes on my system

### DIFF
--- a/server/quickstart.sh
+++ b/server/quickstart.sh
@@ -96,9 +96,10 @@ services:
       - 3002:3002 # Public albums
       # - 3003:3003 # Auth
       # - 3004:3004 # Cast
-    # environment:
-    #   ENTE_API_ORIGIN: http://localhost:8080
-    #   ENTE_ALBUMS_ORIGIN: https://localhost:3002
+    # Modify these values to your custom subdomains, if using any
+    environment:
+      ENTE_API_ORIGIN: http://localhost:8080
+      ENTE_ALBUMS_ORIGIN: https://localhost:3002
 
   postgres:
     image: postgres:15


docs/docs/self-hosting/index.md
```diff
@@ -5,20 +5,32 @@ description: Getting started self hosting Ente Photos and/or Ente Auth
...
+on the client. It performs well on small cloud instances, old laptops, and even
 [low-end embedded devices](https://github.com/ente-io/ente/discussions/594).
 
+### Software
+
+#### Operating System
+
+Any Linux or \*nix operating system, Ubuntu or Debian is recommended to have a
+good Docker experience. Non-Linux operating systems tend to provide poor
+experience with Docker and difficulty with troubleshooting and assistance.
```
vishnukvmd, line 27: I feel we should be crisp about our requirements and keep the home page as short as possible. The idea is that anyone can run `quickstart.sh` and if they run into trouble read more.

We could have a separate page for requirements.

docs/docs/self-hosting/index.md
```diff
@@ -28,7 +40,8 @@ sh -c "$(curl -fsSL https://raw.githubusercontent.com/ente-io/ente/main/server/q
 ```
 
 The above `curl` command pulls the Docker image, creates a directory `my-ente`
-in the current working directory and starts all containers required to run Ente.
+in the current working directory and prompts to start the cluster, which upon
```
vishnukvmd, line 43: We don't need to document the bit about entering `y` here?

vishnukvmd (approved):

docs/docs/self-hosting/index.md
grittypuffy: Agreed, will refactor it into a separate page then.

docs/docs/self-hosting/index.md
grittypuffy: Now that I think of it, prompting is enough, will make the prompt a bit clear too.

PR #6292, opened 2025-06-17
[mob] Streaming minor fixes & refactoring

prateekmedia:
## Description

- [x] Better HDR Detection
- [x] Code side all options are merged to one with more documentation
- [x] Better cutoff date
- [x] Max FPS 30
- [x] Max height 720p
- [x] Better Tonemapping
- [x] Remove redundant hls time and video bitrate


mobile/lib/services/video_preview_service.dart
```diff
@@ -70,21 +70,13 @@ class VideoPreviewService {
 
   late final SharedPreferences _prefs;
   static const String _videoStreamingEnabled = "videoStreamingEnabled";
-  static const String _videoStreamingCutoff = "videoStreamingCutoff";
 
   bool get isVideoStreamingEnabled {
-    return _prefs.getBool(_videoStreamingEnabled) ?? false;
+    return _prefs.getBool(_videoStreamingEnabled) ?? true;
```
ua741: Let’s keep it off till we are sure it’s not crushing the app performance or crashing the app cc @setalp

prateekmedia: Done!

ua741 (approved):

PR #6298, opened 2025-06-18
[mob] Log event source for public link handling

ua741:
## Description
Opening public link within the app opens the collection twice, adding logs to investigate the root cause.


prateekmedia (approved):
LGTM!

PR #6308, opened 2025-06-19
[desktop] Fix macOS CI build 

mnvr:


wrobelda:
Can you please check if this is fixed in latest electron-builder? Upgrading it would also enable building aarch64 versions of RPM and DEB packages, see: https://github.com/electron-userland/electron-builder/issues/9563

It would fix https://github.com/ente-io/ente/issues/8025

Also see comments under this commit for reference: https://github.com/signalapp/Signal-Desktop/commit/560dcb91d98ba1177162372a2837919225e3e2c9

PR #6336, opened 2025-06-23
Add CDN support for URLs in S3 configuration

ImJustToNy:
Hi!
This PR adds a way to specify CDN for s3 buckets. This is especially useful if you're using backblaze where egress is free when using alonside cloudflare.

Wish you a good day


ua741:
Please respect the contribution guidelines, and open a discussion before making any code change.

PR #6339, opened 2025-06-23
[auth] Fix ente auth crash on close

wbonbon:
Flutter Windows applications must explicitly call OnDestroy.
may be a Flutter bug, causing crashes if cleanup isn't handled.

### DIFF
--- a/auth/lib/app/view/app.dart
+++ b/auth/lib/app/view/app.dart
@@ -169,6 +169,11 @@ class _AppState extends State<App>
   }
 
   @override
+  void onWindowClose() {
+    windowManager.destroy();
+  }
+
+  @override
   void onWindowResize() {
     WindowListenerService.instance.onWindowResize().ignore();
   }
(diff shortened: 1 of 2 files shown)


prateekmedia:
Hi @wbonbon , Is this PR Work in progress? Cause the WindowManagerListener is not being utilized anywhere. I think you may want to add it to this since it already utilizes WindowListener?
https://github.com/ente-io/ente/blob/76bff25de10aabb24e01b1611567c0a7fd674d39/auth/lib/app/view/app.dart#L172

wbonbon:
~~Hi, I'm new to flutter so I didn't know the correct way to fix it.
I used a pull request to communicate how to fix it, can someone do it for me?
should I close the pull request and post a discussion about it?~~

fixed it based on the advice in the comments

prateekmedia:
@wbonbon Unfortunately this doesn't fixes the error still. Please test it the next time before you commit.

prateekmedia:
It has been fixed, thanks for you inputs @wbonbon.
#6360

PR #6341, opened 2025-06-23
[mob][photos] add hash check for people widget

prateekmedia:
## Description

- [x] Add updatedAt to PeopleEntity
- [x] Add hash functionality to PeopleWidget using this hashCheck
- [x] Clear hash on albums widget clear

## Tests

- [ ] Test if updating different people trigger it or not

### DIFF
--- a/mobile/lib/models/ml/face/person.dart
+++ b/mobile/lib/models/ml/face/person.dart
@@ -6,19 +6,23 @@ import "package:flutter/foundation.dart";
 class PersonEntity {
   final String remoteID;
   final PersonData data;
+  final DateTime updatedAt;
   PersonEntity(
     this.remoteID,
     this.data,
+    this.updatedAt,
   );
 
   // copyWith
   PersonEntity copyWith({
     String? remoteID,
     PersonData? data,
+    DateTime? updatedAt,
   }) {
     return PersonEntity(
       remoteID ?? this.remoteID,
       data ?? this.data,
+      updatedAt ?? this.updatedAt,
     );
   }
 }
--- a/mobile/lib/app.dart
+++ b/mobile/lib/app.dart
@@ -75,7 +75,7 @@ class _EnteAppState extends State<EnteApp> with WidgetsBindingObserver {
       (event) async {
         _changeCallbackDebouncer.run(
           () async =>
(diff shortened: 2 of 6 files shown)


mobile/lib/models/ml/face/person.dart
```diff
@@ -6,19 +6,23 @@ import "package:flutter/foundation.dart";
 class PersonEntity {
   final String remoteID;
   final PersonData data;
+  final DateTime updatedAt;
```
ua741: We should not need to add this property. Based on remoteID, you can get updatedAt

mobile/lib/models/ml/face/person.dart
```diff
@@ -15,6 +15,7 @@ class PersonEntity {
   PersonEntity copyWith({
     String? remoteID,
     PersonData? data,
+    DateTime? updatedAt,
```
ua741: Not needed?

mobile/lib/services/entity_service.dart
```diff
@@ -219,4 +220,22 @@ class EntityService {
...
+
+  Future<String> getHashForIds(List<String> personIds) async {
+    String preHash = "";
+
+    for (final id in personIds) {
+      final entity = await getEntity(EntityType.cgroup, id);
+
+      if (entity != null) {
+        preHash = "$id:${entity.updatedAt}_";
+      }
```
ua741: We can just read list of updatedAt or better just max updatedAt for the list of personIDs from db and use that to identify if recomputing is required?

prateekmedia: In current logic we are creating hash not only based on updatedAt, but whose updatedAt it is. Because
- It might be that one person is removed from selection but we still have same hash if we use the max updatedAt concept.
- If we readList of updatedAt (I did that before), then it might be possible two person have same updatedAt and hash remains the same (I assume).
- This is best way as it not only specifies the updatedAt but whose updatedAt it is, so hash is always unique.

ua741: (Discussed offline)

prateekmedia: Done! I have basically moved the concat logic to sql side so there is no performance loss now.

ua741 (approved):

PR #6366, opened 2025-06-25
Add CDN support for URLs in S3 configuration

ImJustToNy:
Hi!
This PR adds a way to specify CDN for s3 buckets.
Here is discussion: https://github.com/ente-io/ente/discussions/6352
This is especially useful if you're using backblaze where egress is free when using alonside cloudflare.

Wish you a good day


ua741:
Hello,
We will keep an eye for this change request via the Github discussions. Currently, we are routing the file signed url via multiple places (and adding more such endpoints).  Post that, we will build this support.

PR #6406, opened 2025-06-30
[mob][auth] Bump up auth version to 4.4.1

ashilkn:
## Description

For test flight release


ua741:
@ashilkn For tags, let's use the script, it will also bump the version for flathub version.

PR #6428, opened 2025-07-01
[mob][photos] additional fixes for widgets

prateekmedia:
## Description

Some more fixes for widgets.

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -123,7 +123,7 @@ Future<void> _homeWidgetSync([bool isBackground = false]) async {
   }
 
   try {
-    await HomeWidgetService.instance.initHomeWidget();
+    await HomeWidgetService.instance.initHomeWidget(isBackground);
   } catch (e, s) {
     _logger.severe("Error in syncing home widget", e, s);
   }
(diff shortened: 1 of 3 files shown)


mobile/lib/main.dart
```diff
@@ -163,6 +169,7 @@ Future<void> _runMinimally(String taskId, TimeLogger tlog) async {
   LocalFileUpdateService.instance.init(prefs);
   await LocalSyncService.instance.init(prefs);
   RemoteSyncService.instance.init(prefs);
+  await FavoritesService.instance.initFav();
```
ua741: This is expensive call. Is there reason for adding this? We can also use FavoritesServiceInitCompleteEvent event if needed.

mobile/lib/services/album_home_widget_service.dart
```diff
@@ -294,9 +295,13 @@ class AlbumHomeWidgetService {
 
     // If no albums selected, use favorites as default
     if (selectedAlbumIds == null || selectedAlbumIds.isEmpty) {
+      if (!AppLifecycleService.instance.isForeground) {
+        await FavoritesService.instance.initFav();
```
ua741, line 298: Can this result in multiple init call when application will go from background to fg and vice version .. asking because we are using AppLifecycleService

ua741 (approved):

PR #6435, opened 2025-07-01
Custom icons: Mailbox.org, AliasVault

Denperidge:
## Description
Added icons for Mailbox.org and AliasVault!

If a reviewer could double check that the JSON entries are correct? Mailbox.org shows up on Ente as `Mailbox`, but entering `Mailbox` instead of `Mailbox.org` seemed like false positives waiting to happen. 

## Tests
Not relevant as far as I know.


AmanRajSinghMourya:
We have done some refactoring of folder structure in our repo which is causing merge conflicts which are hard to resolve. Could you please create a new PR make sure to pull from the latest main.

PR #6443, opened 2025-07-02
[mob][photos] Fix contact result in search section

AmanRajSinghMourya:
## Description
Fix showing contact result in the search section and also show the albums shared by that contact like the contact section in the shared collection tab.

## Tests
[Video](https://wormhole.app/xk6MZ4#qu9aa_r0VcGhkEDypz7l1g)

### DIFF
--- a/mobile/lib/services/search_service.dart
+++ b/mobile/lib/services/search_service.dart
@@ -1374,11 +1374,18 @@ class SearchService {
   Future<List<GenericSearchResult>> getContactSearchResults(
     String query,
   ) async {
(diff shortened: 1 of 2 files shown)


mobile/lib/services/search_service.dart
```diff
@@ -1409,7 +1416,24 @@ class SearchService {
       }
     }
 
+    for (Collection collection in collections) {
+      if (collection.isHidden() ||
+          collection.isArchived() ||
```
ua741: We can may be remove isArchive check here?

PR #6449, opened 2025-07-02
[mob][photos] only share image link when using Send Link

prateekmedia:
## Description

Same as title. Link is sufficient when doing Send Link.


prateekmedia:
We'll revisit this. Since we have to removed unused things as well and also see if sharing is possible other ways.

PR #6482, opened 2025-07-07
[mobile] Fix memory edge cases

AmanRajSinghMourya:
## Description
Handle edge cases in memory tap when tapping -
1. On 1st memory of the 1st memory block - reset the animation
2. On last memory of the last memory block - do nothing 

## Test

https://github.com/user-attachments/assets/3358bf41-548e-403b-83a2-aabac7a94266

### DIFF
--- a/mobile/apps/photos/lib/ui/home/memories/full_screen_memory.dart
+++ b/mobile/apps/photos/lib/ui/home/memories/full_screen_memory.dart
@@ -278,7 +278,7 @@ class _FullScreenMemoryState extends State<FullScreenMemory> {
       widget.onPreviousMemory!();
     } else {
       isAtFirstOrLastFile = true;
-      _onPageChange(inheritedData, currentIndex);
+      _resetAnimation();
       _toggleAnimation(pause: false);
     }
   }


mobile/apps/photos/lib/ui/home/memories/full_screen_memory.dart
```diff
@@ -260,6 +260,9 @@ class _FullScreenMemoryState extends State<FullScreenMemory> {
       _onPageChange(inheritedData, currentIndex + 1);
     } else if (widget.onNextMemory != null) {
       widget.onNextMemory!();
+    } else {
+      hasFinalFileLoaded = true;
```
ashilkn: Let's use a different variable here? If I understood correctly, this is for making `_toggleAnimation` work. Instead of setting `hasFinalFileLoaded = true` here without actually knowing if final file has loaded, create a different state variable and use use `if (hasFinalFileLoaded || newVariable)` in `_toggleAnimation`.

mobile/apps/photos/lib/ui/home/memories/full_screen_memory.dart
```diff
@@ -261,7 +262,7 @@ class _FullScreenMemoryState extends State<FullScreenMemory> {
     } else if (widget.onNextMemory != null) {
       widget.onNextMemory!();
     } else {
-      hasFinalFileLoaded = true;
+      isAtFirstOrLastFile = true;
```
ashilkn, line 266: Doesn't this have to be set to false later?

mobile/apps/photos/lib/ui/home/memories/full_screen_memory.dart
```diff
@@ -271,10 +276,15 @@ class _FullScreenMemoryState extends State<FullScreenMemory> {
       _onPageChange(inheritedData, currentIndex - 1);
     } else if (widget.onPreviousMemory != null) {
       widget.onPreviousMemory!();
+    } else {
+      isAtFirstOrLastFile = true;
```
ashilkn, line 280: This is set to false immediately on `_onPageChange`, which negates the intent of setting it to true in the first place.

ashilkn:
> On last memory of the last memory block - do nothing

@AmanRajSinghMourya Isn't closing the memory a better behaviour in this case? Or was this already discussed and decided to do nothing instead?

AmanRajSinghMourya:
> > On last memory of the last memory block - do nothing
> 
> @AmanRajSinghMourya Isn't closing the memory a better behaviour in this case? Or was this already discussed and decided to do nothing instead?

It was decided to do nothing

ashilkn (approved):

ua741 (approved):

PR #6549, opened 2025-07-15
[mob][photos] Compression settings & Face detection issue logging

laurenspriem:
## Description

- Set a high minimum value for compression when original format can't be decoded. If there's an image with higher resolution we log it and locally fail indexing for that file.
- Check for EXIF orientation of image when starting ML analysis of that file. If there's a non-standard orientation we log that orientation and locally fail indexing for that file.

## Tests

- Confirmed that this doesn't give upscaled images
- Comfirmed that the decoding issue is not there on mobile
- Tested correct surfacing of `file analysis failed` button

### DIFF
--- a/mobile/apps/photos/lib/utils/image_ml_util.dart
+++ b/mobile/apps/photos/lib/utils/image_ml_util.dart
@@ -65,6 +65,8 @@ Future<DecodedImage> decodeImageFromPath(
           await FlutterImageCompress.compressWithFile(
         imagePath,
         format: CompressFormat.jpeg,
+        minWidth: 8000, // High value to ensure image is not scaled down
+        minHeight: 8000, // High value to ensure image is not scaled down
       );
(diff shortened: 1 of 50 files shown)


mobile/apps/photos/lib/utils/image_ml_util.dart
```diff
@@ -65,6 +65,8 @@ Future<DecodedImage> decodeImageFromPath(
           await FlutterImageCompress.compressWithFile(
         imagePath,
         format: CompressFormat.jpeg,
+        minWidth: 8000, // High value to ensure image is not scaled down
+        minHeight: 8000, // High value to ensure image is not scaled down
```
ua741: Should we keep the limit higher?
Also, let's test and validate that this is working as expected?
Ideally, if the decodedImage width or high is matching the higher limit, we should probably fail and log the values.

PR #6616, opened 2025-07-23
[mob][photos] Face thumbnail fix + smooth scroll

laurenspriem:
## Description

- Fix internal issue with face thumbnail generation
- Make all people page scroll more smooth 

## Tests

Tested on internal build.


mobile/apps/photos/lib/db/ml/db.dart
```diff
@@ -360,6 +360,10 @@ class MLDataDB with SqlDbBase implements IMLDataDB<int> {
         }
         return mapRowToFace(faceMaps.first);
       }
+    } else if (clusterID == null) {
+      _logger.severe(
```
ua741, line 364: warning pls?

laurenspriem: Should only happen when there's an actual issue. But should also get reported later on, so I'll remove

mobile/apps/photos/lib/db/ml/db.dart
```diff
@@ -380,11 +384,19 @@ class MLDataDB with SqlDbBase implements IMLDataDB<int> {
...
           }
         }
+      } else {
+        _logger.severe(
+          "Didn't find any faces for clusterID $clusterID in `getCoverFaceForPerson`. faces: $faces",
+        );
       }
     }
     if (personID == null && clusterID == null) {
+      _logger.severe("personID and clusterID cannot be null both");
```
ua741, line 394: there's already an exception that will be logged?

ua741 (approved):

PR #6710, opened 2025-07-31
cd: add ghaction step for app store

codebydant:
## Description

PR to add App Store Github action step to automate the deployment


prateekmedia:
Hi @codebydant , Thank you for your contribution. We tried your workflow and it didn't worked:
https://github.com/ente-io/ente/actions/runs/16693837843

Besides, this workflow is similar to our last approach to build [iOS workflow for photos](https://github.com/ente-io/ente/pull/6650), but sadly that was incomplete and also not working. So instead of merging this PR we are following a different approach towards making use of our Auth MacOS workflow as an inspiration and make it work for iOS here: #6718

Thank you for your effort towards this.

codebydant:
> Hi @codebydant , Thank you for your contribution. We tried your workflow and it didn't worked: https://github.com/ente-io/ente/actions/runs/16693837843
> 
> Besides, this workflow is similar to our last approach to build [iOS workflow for photos](https://github.com/ente-io/ente/pull/6650), but sadly that was incomplete and also not working. So instead of merging this PR we are following a different approach towards making use of our Auth MacOS workflow as an inspiration and make it work for iOS here: #6718
> 
> Thank you for your effort towards this.


Hi @prateekmedia, thank you for testing it out and for the feedback!

I understand the direction you're taking, and it makes sense. Since there's already a new approach in progress with PR #6718, I’ll go ahead and close this one.

Appreciate your time and the thoughtful update. 

Happy to contribute anytime!

PR #6718, opened 2025-08-01
[auth] ios workflow

prateekmedia:
## Description

Adds iOS workflow for auth

Based on our MacOS workflow.

Previous effort for this by @codebydant in #6710 , this takes a different approach to provide provisioning profile and having correct certificates by relying on codemagic tools.

Resolves #6665

## Tests

- [ ] Test a sample workflow run

### DIFF
--- a/.github/workflows/auth-internal-release.yml
+++ b/.github/workflows/auth-internal-release.yml
@@ -93,7 +93,7 @@ jobs:
 
             - name: Install code signing dependencies
               run: |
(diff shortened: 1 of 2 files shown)


codebydant:
the error in the action could be related to this?

image

prateekmedia:
> the error in the action could be related to this?
> 
> 
> 
> image
> 
> 

Yes, we have to update the certificate and provisioning profile secrets to fix that most probably.

PR #6755, opened 2025-08-05
Increase cache extent gallery

ashilkn:


mobile/apps/photos/lib/utils/gallery_util.dart
```diff
@@ -0,0 +1,18 @@
+import "package:photos/service_locator.dart";
+
+double get galleryCacheExtent {
```
ua741: Let's keep it closer to the gallery namespace.. inside `ui/view/gallery`.

We can avoid putting such things inside utils.. Especially items that won't be utilized outside of the specific context (especially when there's single usage).

ashilkn: Got it, makes sense.

Issue #3428, opened 2024-09-23
Manual app lock in Ente Auth not working on Mac with Touch ID


Gobias64:
### Description

The new manual app lock is not working on a MacBook Pro with Touch ID when I have the app lock set to device lock. It just flashes white briefly but doesn't lock.

If I change the app lock setting to PIN or password, the manual app lock works fine. If I switch back to device lock, the manual app lock works once, but then it stops working again.

### Version

v4.0.0

### What product are you using?

Ente Auth

### What platform are you using?

Desktop - macOS

Gobias64:
I just tested, and this is still an issue in v4.0.2.

jaklan:
Confirm the same issue - when opening the app, the initial device lock works, but when unblocked - it's impossible to lock the app manually. The only way is to quit the app completely and reopen

hernandomingo:
Still happening in 4.3.2 (downloaded with brew)

hernandomingo:
Same issue on 4.3.5 also

FrostYTT:
Same issue with v4.4.4, only touch ID does not work. Password and pin both work.

PR #6793, opened 2025-08-08
[auth] Fix manual app lock with macos touch id

eYdr1en:
## Description
This fixes https://github.com/ente-io/ente/issues/3428
This was broken because of https://github.com/eaceto/flutter_local_authentication/issues/8
I've also added that if the app is locked manually, the macOS Touch ID API won't be called until the user either presses the unlock button again or unfocuses the app and then focuses back on it. This behavior also applies when the app window is closed and then reopened.


mobile/apps/auth/lib/ui/tools/lock_screen.dart
```diff
@@ -42,7 +45,16 @@ class _LockScreenState extends State<LockScreen> with WidgetsBindingObserver {
     invalidAttemptCount = _lockscreenSetting.getInvalidAttemptCount();
     WidgetsBinding.instance.addObserver(this);
     WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
-      _showLockScreen(source: "postFrameInit");
+      final Object? args = ModalRoute.of(context)?.settings.arguments;
+      if (args is Map && args['manual'] is bool) {
+        _isManualPresentation = args['manual'] as bool;
+      } else {
+        _isManualPresentation = false;
+      }
```
prateekmedia: You can reduce it to this assuming args is always a Map.
```suggestion
      _isManualPresentation = args?['manual'] as bool? ?? false;
```

prateekmedia (approved):
LGTM!

prateekmedia:
@eYdr1en Please fix the merge conflicts and we'll merge this.

mobile/packages/lock_screen/lib/ui/lock_screen.dart
```diff
@@ -38,6 +38,9 @@ class _LockScreenState extends State<LockScreen> with WidgetsBindingObserver {
   int remainingTimeInSeconds = 0;
   final _lockscreenSetting = LockScreenSettings.instance;
   late Brightness _platformBrightness;
+  bool _isManualPresentation = false;
```
prateekmedia: You may remove this variable I think cause it seems to be used only once.

PR #6810, opened 2025-08-11
[email] retention mails for usage consumption

grittypuffy:
## Description
This PR adds support for sending notification email for users who are about to exceed storage (95% of utilization incl. bonus, etc.) and deletes notification if the subscription has been altered (deletion, upgrade, downgrade)


server/pkg/controller/email/email_notification.go
```diff
@@ -161,17 +168,24 @@ func (c *EmailNotificationController) OnSubscriptionCancelled(userID int64) {
...
 	err = email.SendTemplatedEmail([]string{user.Email}, "vishnu@ente.io", "vishnu@ente.io", SubscriptionCancelledSubject, SubscriptionCancelledTemplate, nil, nil)
 	if err != nil {
 		log.Error("Error sending email", err)
 	}
+
+	// deletion for storage limit exceeding email notification
+	c.NotificationHistoryRepo.DeleteLastNotification(userID, StorageLimitExceedingTemplateID)
+
+	// deletion for storage limit exceeded email notification
+	c.NotificationHistoryRepo.DeleteLastNotification(userID, StorageLimitExceededTemplateID)
```
ua741, line 182: Instead of cancellation, I think we would want to reset it on account upgrade?

server/pkg/repo/notificationhistory.go
```diff
@@ -24,6 +25,26 @@ func (repo *NotificationHistoryRepository) GetLastNotificationTime(userID int64,
...
+		_, err := repo.DB.Exec(`
+			DELETE FROM notification_history 
+			WHERE user_id = $1 AND template_id = $2 AND sent_time = $3
+		`, userID, templateID, lastNotificationTime.Int64)
+
+		if err != nil {
+			return stacktrace.Propagate(err, "failed to delete last notification")
+		}
+	}
+	return nil
```
ua741: Should we just call delete query instead of querying for last instance. It will simplify code.

server/compose.yaml
```diff
@@ -58,7 +58,7 @@ services:
           sh -c '
           #!/bin/sh
 
-          while ! mc config host add h0 http://minio:3200 changeme changeme1234 2>/dev/null
+          while ! mc alias set h0 http://minio:3200 changeme changeme1234 2>/dev/null
```
ua741, line 61: Assuming this change is intended.

grittypuffy: Yes

server/pkg/repo/notificationhistory.go
grittypuffy: Oh wait, intended to remove those lines, will do.

server/pkg/controller/email/email_notification.go
grittypuffy: If a user cancels their subscription and resubscribes, it will be cleaner to handle that edge case, that's why I did this. Will check on this.

ua741: If they got a notification, and then cancel.. will they again get notification?
I am thinking of this flow
- User is exceeding storage, already got an email
- They cancel.. again they will get an email..
- They upgrade/re-sub. will they again get email?

Just not doing anything on cancellation might be enough. I would be missing certain cases.

grittypuffy: Yes, it goes with the flow you have described, for any action on their subscription (incl. cancellation) the notification must be sent again for proactively addressing their storage quota usage. That's why I handled the notification with cancellation too, it's an edge case.

server/pkg/repo/user.go
```diff
@@ -332,6 +332,118 @@ func (repo *UserRepository) GetUsersWithIndividualPlanWhoHaveExceededStorageQuot
...
+		}
+		if refBonusStorage, ok := refBonus[user.ID]; ok {
+			addOnBonusStorage := addOnBonus[user.ID]
+			// cap usable ref bonus to the subscription storage + addOnBonus
+			if refBonusStorage > (subStorage + addOnBonusStorage) {
+				refBonusStorage = subStorage + addOnBonusStorage
+			}
+			if (storageConsumed) <= (95 * (subStorage + refBonusStorage + addOnBonusStorage) / 100) {
+				continue
+			}
```
ua741, line 436: If before cron runs (even for existing users), if storageConsumed is > 100%, will they get two emails.. one for exceeding and another for exceeded?

grittypuffy: Fixed this

ua741:
Cleaning up old stale PRs as we are not planning to incorporate these changes in the near term.

PR #6832, opened 2025-08-12
[auth] Update assets for f-droid

ua741:


prateekmedia (changes requested):
Height of the image is very small (512px), which might make it blurry for people looking at them. Atleast we should be aiming for 1024px, previous one was 1440px for reference.

ua741:
> Height of the image is very small (512px), which might make it blurry for people looking at them. Atleast we should be aiming for 1024px, previous one was 1440px for reference.

Yes, had requested design team for assets after seeing this :)

prateekmedia (commented):
We can update title.txt and update it to "Ente Auth"

PR #6841, opened 2025-08-13
[auth] auto backup feature

a5xwin:
This PR adds a new automatic local backup feature that allows users to enable automatic backups of their 2FA codes to any user specified folder on their device. I've added a new Local Backup at: Settings > Data > Local Backup for toggling this feature ON/OFF. Once the feature is ON, a backup is automatically triggered whenever a 2FA code is added, edited, or deleted by hooking into the CodeStore. Plus, an initial backup is also created the moment a user enables the feature or changes the backup location.

### DIFF
--- a/mobile/apps/auth/lib/services/flagservice.dart
+++ b/mobile/apps/auth/lib/services/flagservice.dart
@@ -1,0 +1,13 @@
+import 'package:ente_auth/core/configuration.dart';
+import 'package:flutter/foundation.dart';
+
+class FeatureFlagService {
+  static bool isInternalUserOrDebugBuild() {
+    final String? email = Configuration.instance.getEmail();
+    final userID = Configuration.instance.getUserID();
+    if (email == null || userID == null) {
+      return kDebugMode;
+    }
+    return (email.endsWith("@ente.io")) || (userID < 1000) || kDebugMode;
+  }
+}
--- a/mobile/apps/auth/lib/store/code_store.dart
+++ b/mobile/apps/auth/lib/store/code_store.dart
@@ -7,7 +7,6 @@ import 'package:ente_auth/events/codes_updated_event.dart';
 import 'package:ente_auth/models/authenticator/entity_result.dart';
 import 'package:ente_auth/models/code.dart';
 import 'package:ente_auth/services/authenticator_service.dart';
-import 'package:ente_auth/services/local_backup_service.dart';
 import 'package:ente_auth/store/offline_authenticator_db.dart';
 import 'package:ente_events/event_bus.dart';
 import 'package:logging/logging.dart';
--- a/mobile/apps/auth/lib/ui/settings/data/data_section_widget.dart
+++ b/mobile/apps/auth/lib/ui/settings/data/data_section_widget.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:ente_auth/l10n/l10n.dart';
 import 'package:ente_auth/services/deduplication_service.dart';
+import 'package:ente_auth/services/flagservice.dart';
 import 'package:ente_auth/theme/ente_theme.dart';
 import 'package:ente_auth/ui/components/captioned_text_widget.dart';
 import 'package:ente_auth/ui/components/expandable_menu_item_widget.dart';
@@ -91,18 +92,19 @@ class DataSectionWidget extends StatelessWidget {
           );
         },
       ),
-      MenuItemWidget(
-        captionedTextWidget: CaptionedTextWidget(
-          title: l10n.localBackupSidebarTitle,
+      if (FeatureFlagService.isInternalUserOrDebugBuild())
+        MenuItemWidget(
+          captionedTextWidget: CaptionedTextWidget(
+            title: l10n.localBackupSidebarTitle + "(i)",
+          ),
+          pressedColor: getEnteColorScheme(context).fillFaint,
+          trailingIcon: Icons.chevron_right_outlined,
+          trailingIconIsMuted: true,
+          onTap: () async {
+            await _handleLocalBackupClick(context);
+          },
         ),
-        pressedColor: getEnteColorScheme(context).fillFaint,
-        trailingIcon: Icons.chevron_right_outlined,
-        trailingIconIsMuted: true,
-        onTap: () async {
-          await _handleLocalBackupClick(context);
-        },
-      ),
-      sectionOptionSpacing,
+      if (FeatureFlagService.isInternalUserOrDebugBuild()) sectionOptionSpacing,
     ]);
     return Column(
       children: children,
--- a/mobile/apps/auth/macos/Podfile.lock
+++ b/mobile/apps/auth/macos/Podfile.lock
@@ -165,34 +165,34 @@ EXTERNAL SOURCES:
     :path: Flutter/ephemeral/.symlinks/plugins/window_manager/macos
 
 SPEC CHECKSUMS:
-  app_links: afe860c55c7ef176cea7fb630a2b7d7736de591d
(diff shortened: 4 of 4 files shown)


mobile/apps/auth/lib/services/local_backup_service.dart
```diff
@@ -0,0 +1,45 @@
...
+import 'package:ente_auth/store/code_store.dart';
+import 'package:flutter/foundation.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+class LocalBackupService {
+  static final LocalBackupService instance =
+      LocalBackupService._privateConstructor();
+  LocalBackupService._privateConstructor();
+
+  Future<void> init() async {}
```
ua741: This can be removed?

mobile/apps/auth/lib/services/local_backup_service.dart
```diff
@@ -0,0 +1,45 @@
...
+
+      final isEnabled = prefs.getBool('isAutoBackupEnabled') ?? false;    //checks if toggle is on
+      if (!isEnabled) return;
+
+      final backupPath = prefs.getString('autoBackupPath');
+      if (backupPath == null) return;
+
+      debugPrint("--- Change detected, triggering automatic backup... ---");
+
+      final allCodes = await CodeStore.instance.getAllCodes(sortCodes: false);
```
ua741: Let's use existing method in current export to get the codes?

mobile/apps/auth/lib/services/local_backup_service.dart
```diff
@@ -0,0 +1,45 @@
...
+      final validCodes = allCodes.where((code) => !code.hasError);
+      String backupContent = "";
+      for (final code in validCodes) {
+        backupContent += "${code.toOTPAuthUrlFormat()}\n";
+      }
+
+      if (backupContent.trim().isEmpty) return;
+
+      const fileName = 'ente-auth-auto-backup.txt';
```
ua741: We should include date in the file name?

Also, should we automatically cleanup exports older than x days?

Can we check how other similar app manages the lifecycle for export files just to get some idea about potential expectation from the users. May be in the initially cut we don't really need to auto delete..

mobile/apps/auth/lib/services/local_backup_service.dart
```diff
@@ -0,0 +1,45 @@
...
+      final filePath = '$backupPath/$fileName';
+      final backupFile = File(filePath);
+      await backupFile.writeAsString(backupContent);
+
+      debugPrint('Automatic backup successful! Saved to: $filePath');
+    } catch (e, s) {
+      debugPrint('Silent error during automatic backup: $e\n$s');
```
ua741: Use logger.severe here

mobile/apps/auth/lib/ui/settings/data/local_backup_settings_page.dart
```diff
@@ -0,0 +1,197 @@
...
+                    Text(
+                      "Select a folder to save backups. Backups run automatically when entries are added, deleted, or edited.",
+                      style: getEnteTextTheme(context).small,
+                    ),
+                    const SizedBox(height: 16),
+                    Column(
+                      crossAxisAlignment: CrossAxisAlignment.start,
+                      children: [
+                        Text(
+                          'Current Location:',
```
ua741: Let's extract the strings

mobile/apps/auth/android/app/build.gradle
```diff
@@ -30,7 +30,7 @@ if (keystorePropertiesFile.exists()) {
 
 android {
     namespace "io.ente.auth"
-    compileSdk 36
+    compileSdk 34
```
ua741: To submit update on playstore, we will need to bump it to 35 ore more.

mobile/apps/auth/lib/ui/settings/data/local_backup_settings_page.dart
```diff
@@ -0,0 +1,244 @@
...
+                        const SizedBox(height: 10),
+    SizedBox(
+      width: double.infinity,
+      child: ElevatedButton(
+        style: ElevatedButton.styleFrom(
+          backgroundColor: Theme.of(context).colorScheme.secondary,
+          foregroundColor: Theme.of(context).colorScheme.onSecondary,
+        ),
+        onPressed: _showSetPasswordDialog,
+        child: const Text('Set Backup Password'),
```
ua741: extract

mobile/apps/auth/lib/ui/settings/data/import/encrypted_ente_import.dart
```diff
@@ -99,27 +106,62 @@ Future<void> _decryptExportData(
...
+              String otpUrl = jsonDecode(line);
+              parsedCodes.add(Code.fromOTPAuthUrl(otpUrl));
             } catch (e) {
               Logger('EncryptedText').severe("Could not parse code", e);
             }
           }
-          for (final code in parsedCodes) {
-            await CodeStore.instance.addCode(code, shouldSync: false);
+
+          final List<Code> codesInApp = await CodeStore.instance.getAllCodes();
```
ua741: Let's undo this? I am not sure what's the context for these changes, but they should be done in separate PR if needed.

mobile/apps/auth/lib/ui/code_widget.dart
```diff
@@ -568,7 +568,7 @@ class _CodeWidgetState extends State<CodeWidget> {
       ),
     );
     if (code != null) {
-      await CodeStore.instance.addCode(code);
+      await CodeStore.instance.updateCode(widget.code, code);
```
ua741: Let's undo these changes. For now, Let's just do daily export.

mobile/apps/auth/lib/ui/settings/data/local_backup_settings_page.dart
a5xwin: Thanks for the feedback! I had already fixed this in a later commit.

mobile/apps/auth/lib/ui/settings/data/import/encrypted_ente_import.dart
a5xwin: I did this because I noticed an issue when I was testing the autobackup, the same code was getting added more than once when we try to import a backup.. (creating duplicates) So what I did was, now while importing a backup, it checks if an account with that same secret already exists and prevents duplicates. If it does, it just updates it with the new info, If it's brand new, it adds it, so no duplicates.. That's why I tied them together to get that autobackup working.. Should I update the pr desc?

mobile/apps/auth/lib/ui/code_widget.dart
a5xwin: This was done to avoid a bug: just tapping on a code leads to creation of unnecessary backups, plus the duplicate entry issue.. so if we revert back to addCode, those might come back. Please let me know if we should keep the bug fix in this PR, or revert back to addCode.

mobile/apps/auth/lib/ui/settings/data/import/encrypted_ente_import.dart
ua741: Understood. Still let's not make this enhancement in the same PR. We already have de-dupe support and a normal user should ideally not run into this issue. 
My intent here is to trim down the PR changes to minimal needed changes to support .

mobile/apps/auth/lib/ui/code_widget.dart
ua741: Let's revert to addCode and only keep minimal change for daily export. We will merge those changes, and then iteratively improve in separately as needed.

Issue #5958, opened 2025-05-18
smtp.SendMail blocks until TCP timeout when using an SMTP server with implicit TLS


oddlama:
### Description

When configuring museum to use a smtp server which does not support STARTTLS (e.g. only supports implicit TLS on port 465), then `smtp.SendMail` will block indefinitely (?), or at least for so long that no error is ever logged by museum. This also causes any requests to `POST /users/ott` to time out after 60 seconds, or whatever timeout the browser has set - which to users will look like the sign up hangs after the client generated keys.

Specifically, I'm talking about this line:

https://github.com/ente-io/ente/blob/b2ab0679b4fe90dc1778047d8c71fe7160b5dc0d/server/pkg/utils/email/email.go#L107

I quickly looked it up and found out that `smtp.SendMail` does not automatically switch to implicit TLS when the port is set to `465`, but instead tries to speak STARTTLS on another port entirely. (related: https://stackoverflow.com/a/61155147)

It took me quite a while to narrow the problem down to smtp since there were no logs by museum, the only thing that happened is that POST /users/ott fails while all other requests succeed. Theoretically i guess the sign up could have proceeded, since the OTT code was logged into the console before any of the smtp stuff got triggered.

Anyhow, I would love to allow ente to send emails for signup, but my provider only supports implicit TLS on 465. Would it be possible to implement a new option that allows us to enable implicit tls for the smtp connections?

And thank you for the great software ❤ 

### Version

v1.0.4

### Last working version

N/A

### What product are you using?

Ente Photos

### What platform are you using?

Server

PR #6863, opened 2025-08-15
[server] Add optional TLS/SSL encryption for SMTP

Nilaik:
## Description

Implement TLS/SSL encryption for sending emails via SMTP. When an SMTP provider explicitly requires TLS/SSL communication the current implementation runs in a timeout and fails. A new configuration parameter for smtp was added to enable TLS/SSL communication.

This would solve #5958 

## Tests

I built a local docker image of my branch. The email provider I was using is mailbox.org and using the tls configuration. Registering a new user then resulted in a sent email containing the verification code.

I did not test a setup without TLS/SSL.

### DIFF
--- a/cli/cmd/admin.go
+++ b/cli/cmd/admin.go
@@ -142,6 +142,22 @@ var _updateFreeUserStorage = &cobra.Command{
 	},
 }
 
+var _sendMail = &cobra.Command{
+	Use:   "send-mail <to-email> <from-email> <from-name>",
+	Args:  cobra.ExactArgs(3),
+	Short: "Sends a test mail via the admin api",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		recoverWithLog()
+		var flags = &model.AdminActionForUser{}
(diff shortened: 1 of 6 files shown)


mnvr:
> Tests

Thank you. Could you specify the tests you did.

Nilaik:
I updated the description. Unfortunately I wasn't able to test a non TLS/SSL setup.

By the way is there an easier way to test the email configuration than registering a new user? For example a command in the CLI tool to just sent a test email?

Nilaik:
I found the admin API POST handler for sending emails (/api/mail) and added a CLI admin command for sending a test email. This makes setting up and testing the email configuration easier:

```
go run github.com/ente-io/cli admin send-mail recipient@gmail.com ente@example.com "Ente Test"
Assuming admin@example.com as the Admin 
------------
Successfully sent test email to recipient@gmail.com
```

ua741:
Thank you for your contribution @Nilaik.
I have tested the non-tls flow and verified that it works as expected. Will merge this change and fix the linter issue in separate PR.

ua741 (approved):

PR #6882, opened 2025-08-18
[mob][photos] fix widget tap on iOS

prateekmedia:
## Description

Widget tap was not opening the specified photo because the app group id was not setting correctly when the method was being called.


mobile/apps/photos/scripts/store_changes.txt
```diff
@@ -1,3 +1,4 @@
+- Fix Widget tap on iOS
```
ua741: We can remove this (anyways this is going to playstore)

PR #6886, opened 2025-08-18
pin to specific release of debian to avoid collation mismatch

swiffer:
## Description

this pins the postgres container to a specific version of debian to prevent issues with updated glibc in newer deb release

https://github.com/docker-library/postgres/issues/1356

### DIFF
--- a/server/config/compose.yaml
+++ b/server/config/compose.yaml
@@ -21,7 +21,7 @@ services:
     command: "TCP-LISTEN:3200,fork,reuseaddr TCP:minio:3200"
 
   postgres:
-    image: postgres:15
+    image: postgres:15-trixie
     env_file:
       - .env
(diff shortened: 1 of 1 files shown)


mnvr:
Copy pasting from the linked thread to improve chances of people finding it if they're searching in this repo

Message 
> DETAIL: The database was created using collation version 2.36, but the operating system provides version 2.41.
HINT: Rebuild all objects in this database that use the default collation and run ALTER DATABASE postgres REFRESH COLLATION VERSION, or build PostgreSQL with the right library version`

tl;dr; solution 
> (after a backup) you could use it to test if the recommended ALTER DATABASE postgres REFRESH COLLATION VERSION

mnvr (approved):

PR #6929, opened 2025-08-21
[mob][photos] refactor Theme.of(context)

prateekmedia:
## Description

Calling Theme.of(context) multiple times can lead to performance issue, this PR tries to take out most of the Theme.of(context) references.

Along with that it also proposes a new method of fetching the ColorScheme and TextTheme of EnteTheme by using EnteTheme.getColorScheme(theme) & EnteTheme.getTextTheme(theme) respectively.

Older methods getEnteColorScheme(context) and getEnteTextTheme(context) are deprecated as they also relied on this Theme.of(context) call.

Next up: Using this pattern for MediaQuery.of(context) and AppLocalizations.of(context) to not repeat them within a widget, we can also create common variables for the color scheme and text theme instead of using them inline.

Ref: https://medium.com/@alaxhenry0121/flutters-hidden-tax-why-99-of-apps-are-wasting-performance-on-buildcontext-lookups-e2906fbeb83f


ua741:
I don't think this will have any noticeable impact. Will let @ashilkn review this and see if it make sense to make such changes.

@prateekmedia can you do any basis benchmarking to see the impact, doesn't have to be within this app.

ashilkn:
> Calling Theme.of(context) multiple times can lead to performance issue, this PR tries to take out most of the Theme.of(context) references.

This PR has removed redundant `Theme.of(context)` calls directly in build methods which will improve performance (non-noticeable in most cases but good to have I guess).
##


> Along with that it also proposes a new method of fetching the ColorScheme and TextTheme of EnteTheme by using EnteTheme.getColorScheme(theme) & EnteTheme.getTextTheme(theme) respectively.

> Older methods getEnteColorScheme(context) and getEnteTextTheme(context) are deprecated as they also relied on this Theme.of(context) call.

The depreciation of these methods are mostly unnecessary because most of the optimisations can be done by making sure to reuse a single `EnteColorScheme` or `EnteTextTheme` returned by `getEnteColorScheme()` and `getEnteTextTheme()` throughout a `build()` of a Widget. 

One place where the new approach is better is when there are multiple helper methods that require `EnteColorScheme` and `EnteTextTheme` used in a `build` method. Here, instead of passing both `EnteColorScheme` and `EnteTextTheme`, using the new approach requires only passing `ThemeData`.

```dart
// old
//...
build(BuildContext context){
   final colorScheme = getEnteColorScheme(context);
   final textTheme = getEnteTextTheme(context);
   return getExampleWidget(colorScheme, textTheme)
}

//...

Widget getExampleWidget(EnteColorScheme colorScheme, EnteTextTheme textTheme){
   return Container(
      color: colorScheme.fillBase,
      child: Text("Example", style: textTheme.small)
   )
}



// new
//...
build(BuildContext context){
   final theme = Theme.of(context);
   return getExampleWidget(theme)
}

//...

Widget getExampleWidget(ThemeData theme){
   return Container(
      color: theme.getColorScheme.fillBase,
      child: Text("Example", style: theme.getTextTheme.small)
   )
}
```

Another place where I found the new approach to do better is when both `EnteTextTheme` and `EnteColorScheme` are required in a `build` function. In the new approach, only one call to `Theme.of(context)` will be made compared to two in the existing approach. 

Although initially I felt deprecation of `getEnteColorScheme()` and `getEnteTextTheme()` are unnecessary, the tiny improvements in the new approach do add some value. Let's land this, but we need to do the same for locker and auth @prateekmedia

ashilkn:
And in general, passing around of `BuildContext` should be avoided from what I know. So this change is going in the right direction I feel.

prateekmedia:
@ashilkn If this gets landed I will replicate in auth and locker as well otherwise me and others also will get lot of merge conflict.

ashilkn:
@prateekmedia  Since this PR touches a lot of screens, should we merge this after the freeze on Monday?

prateekmedia:
I am fine with it @ashilkn , since it might need some testing.

PR #6971, opened 2025-08-26
[auth] Fix missing token

AmanRajSinghMourya:
## Description
When using Auth without backup, it was giving a error `Offline key is missing` 

**Reason**: During the `init` of `BaseConfiguration` if the `tokenKey` is not set, we clear all the keys in the secure storage, and in this process the `offlineAuthSecretKey` was also getting cleared 

**Fix** Fixed by skipping the deletion of `offlineAuthSecretKey`  

## Tests
[Test Video](https://wormhole.app/qz3mol#Dlhr0NRpVQVQsrid2X-quA)

### DIFF
--- a/mobile/packages/configuration/lib/base_configuration.dart
+++ b/mobile/packages/configuration/lib/base_configuration.dart
@@ -4,11 +4,11 @@ import 'dart:convert';
 import 'dart:io' as io;
 
 import 'package:bip39/bip39.dart' as bip39;
+import 'package:ente_configuration/constants.dart';
 import 'package:ente_base/models/database.dart';
 import 'package:ente_base/models/key_attributes.dart';
 import 'package:ente_base/models/key_gen_result.dart';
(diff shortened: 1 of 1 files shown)


mobile/packages/configuration/lib/base_configuration.dart
```diff
@@ -382,7 +395,15 @@ class BaseConfiguration {
...
       if (!_preferences.containsKey(tokenKey)) {
-        await _secureStorage.deleteAll();
+
+        // Delete all keys except preserved ones
+        final allKeys = await _secureStorage.readAll();
+        for (final key in allKeys.keys) {
+          if (!preservedKeys.contains(key)) {
+            await _secureStorage.delete(key: key);
+          }
+        }
```
ua741: Could be in same method resetSecureStorage()

ua741 (approved):

PR #6975, opened 2025-08-26
selection sheet redesign

kartikaeyy:


ua741:
Cleaning up old stale PRs as we are not planning to incorporate these changes in the near term.

PR #6998, opened 2025-08-28
[mob][photos] bypass size/duration limits for manual video stream requests

prateekmedia:
## Summary
- Modified `_checkFileForPreviewCreation` method to accept `isManual` parameter
- Bypass 500MB file size and 60 second duration limits when user manually triggers video stream processing
- Maintains size/duration restrictions for automatic streaming to preserve device performance

## Test plan
- [x] Manual Create/Recreate Stream button bypasses 500MB and 60 second limits
- [x] Automatic streaming still respects size and duration restrictions
- [x] Files larger than 500MB or longer than 60 seconds can be manually processed

### DIFF
--- a/mobile/apps/photos/lib/services/video_preview_service.dart
+++ b/mobile/apps/photos/lib/services/video_preview_service.dart
@@ -299,7 +299,7 @@ class VideoPreviewService {
         "Starting video preview generation for ${enteFile.displayName}",
       );
       // elimination case for <=10 MB with H.264
-      var (props, result, file) = await _checkFileForPreviewCreation(enteFile);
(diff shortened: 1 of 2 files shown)


mobile/apps/photos/lib/services/video_preview_service.dart
```diff
@@ -936,7 +937,7 @@ class VideoPreviewService {
     }
     final int size = enteFile.fileSize!;
```
ua741: Previosuly, the db query was written in such a way that these values will be always non-null.

prateekmedia: Resolved.

PR #7065, opened 2025-09-05
[server] Support for coupons

ua741:
## Tests
Tested locally for happy cases

### DIFF
--- a/server/pkg/controller/discountcoupon/controller.go
+++ b/server/pkg/controller/discountcoupon/controller.go
@@ -44,42 +44,40 @@ type AddCouponsRequest struct {
 	Codes        []string `json:"codes" binding:"required"`
 }
 
-func (c *Controller) ClaimCoupon(ctx *gin.Context, req ClaimCouponRequest) {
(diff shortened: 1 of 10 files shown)


mnvr (approved):

server/pkg/controller/discountcoupon/controller.go
```diff
@@ -0,0 +1,187 @@
...
+	Codes        []string `json:"codes" binding:"required"`
+}
+
+func (c *Controller) ClaimCoupon(ctx *gin.Context, req ClaimCouponRequest) error {
+	if !AllowedProviders[req.ProviderName] {
+		return stacktrace.Propagate(ente.NewBadRequestWithMessage("Invalid provider name"), "")
+	}
+	go c.processClaimRequest(ctx, req)
+	return nil
+}
```
yuseferi, line 53: really?
Is this the way the backend is written?
Having a function that expects an error, then  run the `processClaimRequest` as a goroutine that you don't wait for it, and whatever happens inside of it does not matter, and you just log it?

I can't believe this 🫨 🫨 🫨

PR #7071, opened 2025-09-06
[locker] Update sharing package 

AmanRajSinghMourya:
## Description
Move `notification_widget`, `user_avatar_widget` & `user_extension` to `sharing` package.
Update imports, extract strings and update configuration.


mobile/packages/accounts/pubspec_overrides.yaml
```diff
@@ -12,6 +12,8 @@ dependency_overrides:
     path: ../logging
   ente_network:
     path: ../network
+  ente_sharing:
```
vishnukvmd: Why does `accounts` have a dependency on `sharing`?

mobile/packages/lock_screen/pubspec_overrides.yaml
```diff
@@ -12,6 +12,8 @@ dependency_overrides:
     path: ../logging
   ente_network:
     path: ../network
+  ente_sharing:
```
vishnukvmd: Same as above.

mobile/packages/utils/pubspec.yaml
```diff
@@ -14,6 +14,8 @@ dependencies:
     path: ../../packages/configuration
   ente_logging:
     path: ../../packages/logging
+  ente_sharing:
```
vishnukvmd: Same, why does `utils` care about `sharing`?

vishnukvmd (approved):

PR #7090, opened 2025-09-07
[auth] multi select feature

a5xwin:
This PR introduces a new multi-select feature that allows users to perform bulk actions like pinning, tagging, and deleting multiple 2FA codes at once. Users can enter the multi-select mode by long-pressing on any code in the main list. This brings a new contextual action bar at the bottom of the screen, which shows a set of six actions (edit, share, qr, pin, add tag, trash) when a single code is selected. For multiple codes, it displays set of three actions (pin, add tag, trash) The bar also displays the icons of selected items..

A new “Add tag” bottom sheet was created, allowing users to add or remove tags from multiple codes simultaneously. It also includes the ability to create new tags as well.

Also, the AppBar background color is now consistent and no longer changes on scroll.


ua741:
@codex review

ua741 (approved):

PR #7130, opened 2025-09-10
[mob][photos] Add text embeddings cache service

laurenspriem:
## Description

Add text embeddings cache service to prevent recomputes for:
- Memories
- Magic cache

## Tests


mobile/apps/photos/lib/db/ml/db.dart
```diff
@@ -1429,6 +1430,50 @@ class MLDataDB with SqlDbBase implements IMLDataDB<int> {
...
+
+    if (results.isNotEmpty) {
+      final blob = results.first[embeddingColumn] as Uint8List;
+      return Float32List.view(blob.buffer).toList();
+    } else {
+      // Clean up old/invalid entries if they exist
+      await db.execute(
+        'DELETE FROM $textEmbeddingsCacheTable WHERE $textQueryColumn = ?',
+        [query],
+      );
```
ua741: Is this needed?

ua741 (approved):

mobile/apps/photos/lib/db/ml/db.dart
laurenspriem: It's to remove old embeddings (i.e. older ml version or 3+ months old). I'm indifferent about removing this or not. We can remove since anyways we don't expect to have a lot of embeddings in this table. On the other hand keeping it is also fine since this will be rarely run (once every 3 months + whenever we update the ml version).

ua741 (approved):

mobile/apps/photos/lib/db/ml/db.dart
laurenspriem: I've changed it to be more readable

PR #7158, opened 2025-09-12
Add custom icons

bhaswardg08:
Added custom icons for two services: Leibniz-IPHT and UOA.


AmanRajSinghMourya:
Hey, the size of icons seems to be large, can you compress them and upload them again

bhaswardg08:
Hi,

I tried to compress the images further but only managed it slightly. These as well as the originals are way below 50 kB outlined in the docs. 
leibniz-ipht-compressed
uoa-compressed

PR #7260, opened 2025-09-24
Show/hide window when tray icon is clicked

keysmashes:
## Description	

Right now, clicking the tray icon (at least on Linux) does nothing. This is contrary to platform expectations: clicking the tray icon should usually do something.

The most useful action to take is probably to open the app (or close it if it's already open), and that's what this PR does.

Specifically, this opens the window if it's not "visible", else closes it. On my system, this produces the following behaviours:

- if Ente is focused, clicking the icon hides it
- if Ente is not focused but it's visible on another monitor, the icon hides it
- if Ente is not the visible window in a tab, the icon focuses the existing window
- if Ente is on another workspace, the icon hides it (that seems suboptimal...)
- if Ente is hidden, the icon opens a new Ente window

I'm open to suggestions on the exact behaviours here (e.g. it would seem reasonable to only close the app if it's _focused_, not just if it's visible), but there's such a variety of behaviours in existing apps that I don't think there's necessarily a single correct set of behaviours to aim for.

This resolves https://github.com/ente-io/ente/discussions/6777.

## Tests

I've manually tested this on Linux. I'd appreciate testers on Windows and/or macOS.

### DIFF
--- a/desktop/src/main.ts
+++ b/desktop/src/main.ts
@@ -680,6 +680,15 @@ const setupTrayItem = (mainWindow: BrowserWindow) => {
     const tray = new Tray(trayIcon);
     tray.setToolTip("Ente Photos");
     tray.setContextMenu(createTrayContextMenu(mainWindow));
+    if (process.platform === "linux") {
+        tray.on("click", () => {
+            if (mainWindow.isFocused()) {
+                mainWindow.hide();
+            } else {
+                mainWindow.show();
+            }
+        });
+    }
 };
 
 /**


mnvr:
> clicking the tray icon ... does nothing

But it does do something. It might not be doing something under some particular Linux DE, but on other OSes it does open a context menu, that functionality is triggered by the line above the new code that the PR added:

```
    tray.setContextMenu(createTrayContextMenu(mainWindow));
```

and the new code will disable the context menu.

Even in the discussion you linked, the OP was not saying that the left click didn't do anything, they said that the left click did something (open a menu) which they didn't want and as per them should be done as a right click instead of a left click. I don't agree with them, but I don't disagree with them either - I just don't have enough signal from the community to change things from the current behavior (context menu on click).

mnvr:
From the linked discussion:

> That is, a left-click should toggle the app window display (show the app window if it is currently hidden, or hide it if it is currently shown), while a right-click should show the menu (as it is).

We can do this, but only for Linux to avoid breaking the expectations of users on other OSes. If you do wish to do this, please test both the left click (new behaviour) and right click (existing behaviour)

keysmashes:
> Even in the discussion you linked, the OP was not saying that the left click didn't do anything, they said that the left click did something (open a menu)

Oops, I missed that, sorry.

> the new code will disable the context menu

If this is the case, that sounds suboptimal! I hope there is a way to achieve both "on platforms where left-click and right-click are distinguished, left-click opens Ente and right-click opens the context menu" and "on platforms where left-click and right-click are not distinguished, any click opens the context menu" simultaneously.

I've updated the PR to add the click listener exclusively on Linux. I've also changed it so that the window is only closed/hidden if Ente currently has focus. Left-click and right-click both work as intended for me, on Linux using the i3 window manager and i3bar/snixembed for the tray.

PR #7261, opened 2025-09-24
docs: Clarify which creds to unhide album

cpeel:
## Description

After accessing the hidden album on my Android using my device's passcode, I was confused with which creds to use on the desktop as I assumed they would be the same -- but they're clearly not.

Small update to the docs to clarify which password to use to access hidden album on desktop.


mnvr (approved):
Thank you!

PR #7266, opened 2025-09-26
[locker] Handle errors in resetSecureStorage method

AmanRajSinghMourya:
## Description
The issue is in the resetSecureStorage() method. It's trying to call readAll(), which causes the BadPaddingException.

`PlatformException(Exception encountered, readAll, javax.crypto.BadPaddingException: error:1e000065:Cipher functions:OPENSSL_internal:BAD_DECRYPT`

### DIFF
--- a/mobile/packages/configuration/lib/base_configuration.dart
+++ b/mobile/packages/configuration/lib/base_configuration.dart
@@ -85,12 +85,18 @@ class BaseConfiguration {
   }
 
   Future<void> resetSecureStorage() async {
-    // Delete all keys except preserved ones
-    final allKeys = await _secureStorage.readAll();
-    for (final key in allKeys.keys) {
-      if (!preservedKeys.contains(key)) {
-        await _secureStorage.delete(key: key);
+    try {
+      // Delete all keys except preserved ones
+      final allKeys = await _secureStorage.readAll();
+      for (final key in allKeys.keys) {
+        if (!preservedKeys.contains(key)) {
+          await _secureStorage.delete(key: key);
+        }
       }
+    } catch (e) {
+      // On error, clear all keys
+      _logger.warning('Error resetting secure storage: $e');
+      await _secureStorage.deleteAll();
     }
   }
 


mobile/packages/configuration/lib/base_configuration.dart
```diff
@@ -85,12 +85,18 @@ class BaseConfiguration {
...
+      final allKeys = await _secureStorage.readAll();
+      for (final key in allKeys.keys) {
+        if (!preservedKeys.contains(key)) {
+          await _secureStorage.delete(key: key);
+        }
       }
+    } catch (e) {
+      // On error, clear all keys
+      _logger.warning('Error resetting secure storage: $e');
+      await _secureStorage.deleteAll();
```
ua741, line 99: Can this again cause regression in the offline mode?

AmanRajSinghMourya: No I dont think so, coz auth was working fine even when this try catch block is not there.
Also locker installed from playstore is working fine without this change merged to main.

ua741: Let's only clear if offlineKey is missing?  write might also have chance of failing. 
Ideally, we should never run into this catch block in the offline mode.

mobile/packages/configuration/lib/base_configuration.dart
```diff
@@ -85,11 +85,50 @@ class BaseConfiguration {
...
+      // If readAll or delete operations fail (e.g., BadPaddingException),
+      // fall back to deleting everything
+      _logger.warning(
+        'Error during selective deletion, falling back to deleteAll: $e',
+      );
+      try {
+        await _secureStorage.deleteAll();
+        // Try to restore the offline key if we had it
+        if (offlineKey != null) {
+          try {
```
ua741: Let's only clear if offlineKey is missing? write might also have chance of failing.
Ideally, we should never run into this catch block in the offline mode.

mobile/packages/configuration/lib/base_configuration.dart
```diff
@@ -85,11 +85,60 @@ class BaseConfiguration {
...
+      );
+      if (offlineKeyValue != null && offlineKeyValue.isNotEmpty) {
+        _logger.info('Offline mode detected, skipping secure storage reset');
+        return;
+      }
+    } catch (error, stackTrace) {
+      _logger.warning(
+        'Failed to read offline key prior to secure storage reset: $error',
+      );
+      _logger.fine(stackTrace);
```
ua741, line 95: why not log this like
```dart
.warning("your message", error, stackTrace); ?
```

mobile/packages/configuration/lib/base_configuration.dart
```diff
@@ -85,11 +85,60 @@ class BaseConfiguration {
...
+        if (value != null) {
+          preservedEntries[key] = value;
+        }
+      } catch (error, stackTrace) {
+        _logger.warning(
+          'Failed to read preserved key $key before secure storage reset: '
+          '$error',
+        );
+        _logger.fine(stackTrace);
+      }
```
ua741: Same

PR #7292, opened 2025-10-01
[mob][photos] skip permission request in background mode on Android

prateekmedia:
## Description

- Android 15 blocks permission dialogs in background tasks. Skip the ACCESS_MEDIA_LOCATION permission request when running in background mode to prevent silent hangs during background uploads.
- Also Cronet is not working fine in background and fails many a times. It's better to disable it for android until we know it fully works.


laurenspriem:
@codex review

mobile/apps/photos/lib/utils/file_uploader.dart
```diff
@@ -434,6 +434,15 @@ class FileUploader {
     if (Platform.isAndroid) {
       final bool hasPermission = await Permission.accessMediaLocation.isGranted;
       if (!hasPermission) {
+        // In background mode, we can't request permissions (no UI available)
+        // Skip the permission request and continue with upload
+        if (_processType == ProcessType.background) {
+          _logger.info(
+            "Media location access not granted, but continuing in background mode",
+          );
+          return;
```
ashilkn: We shouldn't continue upload if media location access is not granted. Instead of returning, throw `NoMediaLocationAccessError()`

prateekmedia: Done!

PR #7309, opened 2025-10-03
[mob][photos] Fix issue with flipped thumbnails

laurenspriem:
## Description

Use of the `ColorFiltered` flutter widget appears to give issues for HEIC files viewed on Android, where their meta data seems to make the widget flip the child. In this case, it means image thumbnails were getting flipped on selection. Fixed the issue by no longer using the `ColorFiltered` widget and instead relying on a simple combo of Stack and Container with background color. 

## Tests

Tested in debug mode on my pixel phone, for all UI changes involved.


laurenspriem:
@codex review

PR #7312, opened 2025-10-03
[mob][photos] Remove dialog for maps and add privacy description instead

anandbaburajan:
## Description

- Remove maps enable dialog: Enable maps directly when users tap map-related actions
    - Gallery app bar "Show on map" action
    - Search tab locations section
    - File info location widget
    - Empty location state "Go to map" button
- Add maps privacy notices: Display privacy notice in settings and trip layout picker
    - Advanced settings below Maps toggle (always visible)
    - Trip layout picker below options (when Trip is selected)
- Clean up code:
    - Remove enable_map.dart dialog file
    - Delete unused enableMapsDesc translation string


JaffaKetchup:
Hey, just wondering what the privacy text really means?

> Approximate coordinates will be sent to privacy conscious third parties like OpenStreetMap to show the map and location.

This doesn't sound quite right to me. It's not possible to send anything to OpenStreetMap, except a HTTP GET request for a specific tile. See https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames.

If the map loads around a given coordinate, then OpenStreetMap will technically know the tile coordinates that have been loaded (as any website which keeps logs of GET requests), but:

 - Without grouping a set of tile requests together from a single client load (which just doesn't happen), it would be impossible to determine the center tile coordinates loaded
 - Even if those tile coordinates were known, that (assuming a sufficiently zoomed out zoom level) gives very little information about the true coordinates, only a square range, of which the coordinates could be located anywhere within that square.

It is wrong to say "approximate coordinates will be sent to OpenStreetMap", if my understanding is correct.

mnvr:
> Approximate coordinates ...

This change is tied to a different feature - trip layouts (which is almost complete but still behind a feature flag - e.g. see these PRs - https://github.com/ente-io/ente/pulls?q=is%3Apr+trip+layout+is%3Aclosed - we're hoping to release it this week!)

1. You're right, but even tile locations do send data, so we'd like to say that. Plus, trip layouts will require reverse geocoding. 2. Also, multiple users have given us feedback that the current dialog is too obtrusive and verbose. 3. We'd like to keep reuse various strings, esp long ones, to reduce the burden on the volunteers who translate the app.

So to handle all these three, these changes were made - the new privacy description tries to cover all these scenarios.

mobile/apps/photos/lib/ui/viewer/file_details/location_tags_widget.dart
```diff
@@ -299,15 +298,10 @@ class _InfoMapState extends State<InfoMap> {
...
-                                  unawaited(
-                                    requestForMapEnable(context).then((value) {
-                                      if (value) {
-                                        setState(() {
-                                          _hasEnabledMap = true;
-                                        });
-                                      }
-                                    }),
-                                  );
+                                  await flagService.setMapEnabled(true);
```
ua741: As discussed offline, we need to handle errors/exception while triggering this flow (can turn off network for testing

ua741 (approved):

PR #7332, opened 2025-10-05
Add endpoint select dialog to auth.

DiIVIiCH:
## Description
Added a dialog box for selecting an endpoint for the auth site. 

In fact, this is a copy-paste from the photo page.  I have no experience working with React, and I hope it turned out well enough.

## Tests
I compiled a local docker image for the web and checked:
- adding a new account on localhost:8080
- opening the auth page on localhost:8080 with an active login (the password entry dialog opens first)
- logging in to the auth instance deployed on the local network, after changing the endpoint address
- logging in to the instance deployed on the local network with an active login

### Before
2025-10-05 14 18 42 localhost c3868a1e09c1

### After
2025-10-05 17 19 12 localhost 158834f7ca52
2025-10-05 17 19 21 localhost 3b7fa6c50b83


mnvr:
Thanks! For getting this through, we'll need to do a more focused change though. This is adding images and other information from the photos app which is not applicable to the auth app. We just want to include the functionality that handles the clicks and shows the dialog. Also note that this is pending CLA.

DiIVIiCH:
@mnvr 
I think that CLA is broken, i already agreed to it 3 times and it's still not done. 
Screenshot_20251007-224934_Vivaldi


Maybe you have at least Auth image to put into the tappable container? Or you just want to have no images at all right now?

mnvr:
> Maybe you have at least Auth image to put into the tappable container? Or you just want to have no images at all right now?

We can have images, but that's a much bigger change than is needed for solving the issue. The PR should only solve the issue of allowing people to set the custom domain, and not go into redesigning the page (that can be done, but it is best to do that as a separate exercise). I think only a very small amount of code would be needed to integrate the click detector + dialog into the existing page. 

Re the CLA, I'm not sure what's going on in this PR, it seems to be worknig elsewhere. Could you close this PR and reopen another one, since anyways the changes you'll need to make will be much smaller than this PR, and I think it is better to start afresh.

DiIVIiCH:
Ok, got you. Closing this. Will return with a new one a bit later.

PR #7342, opened 2025-10-06
[mob][photos] fix bg tasks

prateekmedia:
## Description

Add more logs and bump workmanager.

### DIFF
--- a/mobile/apps/photos/pubspec.yaml
+++ b/mobile/apps/photos/pubspec.yaml
@@ -223,7 +223,7 @@ dependencies:
   wakelock_plus: ^1.1.1
   wechat_assets_picker: ^9.5.1
   widgets_to_image: ^1.0.0
-  workmanager: ^0.9.0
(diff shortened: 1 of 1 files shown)


mobile/apps/photos/lib/utils/bg_task_utils.dart
```diff
@@ -95,11 +94,20 @@ class BgTaskUtils {
...
       );
       $.info("WorkManager configured");
+
+      // Check if task is scheduled (Android only)
+      if (Platform.isAndroid) {
+        final isScheduled = await workmanager.Workmanager()
+            .isScheduledByUniqueName(backgroundTaskIdentifier);
+        $.info(
+          "Periodic task scheduled status: $isScheduled for $backgroundTaskIdentifier",
+        );
```
ua741, line 109: It might be better to log warning if it's not scheduled instead of logging info here. Easier to parse and less noise in the logs

PR #7353, opened 2025-10-07
[server] Config for embed albums

ua741:


mnvr:
@claude do the changes you suggested here https://github.com/ente-io/ente/pull/7353#issuecomment-3375155034

PR #7362, opened 2025-10-07
[workflow] set fetch-depth to 1 in claude workflows

prateekmedia:
## Description

Set depth to 1


mnvr:
Superceded by https://github.com/ente-io/ente/pull/7378

PR #7398, opened 2025-10-09
[mob] set default flavor for android

prateekmedia:
## Description

Currently if we don't specify a flavor then android build fails, resulting in a re-run which is tedious. This fixes that.

## Tests

- [x] Test android defaults to independent flavor
- [ ] Test if iOS builds successfully


prateekmedia:
default-flavor is being used for scheme selection of iOS, so can't use that.

Issue #7045, opened 2025-09-02
Serious: Ignoring Faces Doesn't Keep them Ignored


RayBB:
### Description

When ignoring faces in Ente Photos, the action does not persist reliably. Faces that are marked as ignored sometimes reappear in the list, especially when reviewing faces via Show More or while using Review mode. Initially, ignored faces disappear as expected, but after continuing to ignore or add new faces, previously ignored ones can resurface. This makes it difficult to manage faces consistently and is particularly frustrating for users who rely on the faces feature.

I can't reliably reporoduce this but it seems to be more of an issue of photos with many faces.

Edit: I have found an instance where this bug is reliably reproduceable.
I uploaded a video [here](https://send.bitwarden.com/#bQfXRJMHjkSMF7NMAU946A/ipux1kxDZK3pWGjteU3dAg) so you can see it (password: ente).
Basically at the 0:25 second mark you can see there are some ignored faces for a photo. But then when you tap edit, pick a face, then tap back, all the ignored faces go away.
Apologies for the somewhat strange video, there was supposed to be audio but it didn't get recorded.

### Version

1.2.0

### What product are you using?

Ente Photos

### What platform are you using?

Mobile - Android

comminutus:
This seems like it regressed and is an issue again in v1.2.15

DVSLabs:
Can we get another look at this as the issue is extremely annoying and makes the faces feature nearly unusable for me @laurenspriem 
Thank you

manuelleduc:
In my experience, this feature never worked. This makes it very difficult to achieve a good classification since the same irrelevant faces are suggested over and over again.
Please let me know if I can be of any help to fix this issue (e.g, by providing debugging data).

RayBB:
I was just going through photos yesterday and confirm it's still a major issue. 

It's especially annoying when you have a photo of like a crowd at an event and there's tons of faces that are popping up and you have to just keep ignoring them again and again.

@laurenspriem do you have any insight on what it would take to improve this?

laurenspriem:
Unfortunately, I'm not able to reproduce it from my side, which makes it difficult to find the root cause and verify any fix. 

@RayBB are you able to reproduce it still?
If someone is able to reproduce the issue consistently, please share the flow for doing so, and any other information that could be of help.

manuelleduc:
That's the steps for me:
- click on the magnifying glass et the bottom of the android UI
- click on people
- scroll down to not identified faces
[...]

PR #7400, opened 2025-10-09
[mob][photos] Bug fixes

laurenspriem:
## Description

- Fixes [ignored faces resurfacing issue](https://github.com/ente-io/ente/issues/7045)
- Other minor fixes

### DIFF
--- a/mobile/apps/photos/lib/services/machine_learning/face_ml/person/person_service.dart
+++ b/mobile/apps/photos/lib/services/machine_learning/face_ml/person/person_service.dart
@@ -267,11 +267,7 @@ class PersonService {
     required String personID,
     required String clusterID,
   }) async {
-    final person = await getPerson(personID);
-    if (person == null) {
-      logger.severe("Person $personID not found in removeClusterToPerson");
-      throw Exception("Person $personID not found");
-    }
+    final person = (await getPerson(personID))!;
     final personData = person.data;
     final clusterInfo = personData.assigned.firstWhere(
       (element) => element.id == clusterID,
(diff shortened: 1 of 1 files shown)


laurenspriem:
@codex review

mobile/apps/photos/lib/services/machine_learning/face_ml/person/person_service.dart
```diff
@@ -495,7 +502,11 @@ class PersonService {
       );
     }
 
-    final person = (await getPerson(p.remoteID))!;
+    final person = await getPerson(p.remoteID);
+    if (person == null) {
+      logger.severe("Person ${p.remoteID} not found in updateAvatar");
+      throw Exception("Person ${p.remoteID} not found");
+    }
```
ua741: This seems to be happening at multiple place. I don't think we need this change. We will get Null reference error if the expectation mismatch happens. 
If we want, we get create mustGetPerson that will never return null value, and throw exception.

mobile/apps/photos/lib/services/machine_learning/face_ml/person/person_service.dart
```diff
@@ -485,6 +488,10 @@ class PersonService {
   }
 
   Future<PersonEntity> updateAvatar(PersonEntity p, EnteFile file) async {
+    if (file.uploadedFileID == null) {
+      logger.severe("File has no uploadedFileID in updateAvatar");
+      throw Exception("File has no uploadedFileID");
+    }
```
ua741: Can remove this as well?

mobile/apps/photos/lib/services/machine_learning/face_ml/person/person_service.dart
laurenspriem: Thanks, I've reverted those changes now. I think we can merge now (or after freeze)

PR #7401, opened 2025-10-09
[mob][photos] fix inconsistent behavior when deleting from Ente

prateekmedia:
## Description

Fixed inconsistent deletion behavior in the photo viewer with context-aware handling for device folders.

**Problem:**
When deleting a photo that exists both locally and on Ente, "Delete from both" moved to the next photo but "Delete from Ente" kept the deleted photo visible in the viewer until navigating away.

**Solution:**
Implemented context-aware deletion that properly handles both regular galleries and device folder (local-only) contexts:

- Regular galleries: All deletion actions move to next photo
- Device folders: Files remain visible after "Delete from Ente" since they still exist locally in that view

**Changes:**

- Added `isLocalOnlyContext` parameter throughout the deletion flow
- Device folders detected via `device_folder:` tag prefix
- Updated deletion logic to respect context:
  - "Delete from Ente": Removes from viewer UNLESS in device folder context
  - "Delete from device": Removes from viewer if local-only OR in device folder context
  - "Delete from both": Always removes from viewer

**Files Modified:**

- `lib/ui/actions/file/file_actions.dart`: Added context-aware deletion logic
- `lib/ui/viewer/file/file_bottom_bar.dart`: Pass through context flag
- `lib/ui/viewer/file/detail_page.dart`: Added isLocalOnlyContext to configuration
- `lib/ui/viewer/gallery/component/gallery_file_widget.dart`: Detect device folder context

## Tests

- [x] Regular gallery + "Delete from Ente" → moves to next photo
- [x] Regular gallery + "Delete from device" → moves to next photo
- [x] Regular gallery + "Delete from both" → moves to next photo
- [x] Device folder + "Delete from Ente" → stays visible (has local copy)
- [x] Device folder + "Delete from device" → moves to next photo
- [x] Device folder + "Delete from both" → moves to next photo

### DIFF
--- a/mobile/apps/photos/lib/ui/viewer/gallery/component/gallery_file_widget.dart
+++ b/mobile/apps/photos/lib/ui/viewer/gallery/component/gallery_file_widget.dart
@@ -4,6 +4,7 @@ import "package:media_extension/media_extension.dart";
 import "package:media_extension/media_extension_action_types.dart";
 import "package:photos/core/constants.dart";
 import 'package:photos/models/file/file.dart';
+import "package:photos/models/gallery_type.dart";
 import "package:photos/models/selected_files.dart";
 import "package:photos/services/app_lifecycle_service.dart";
 import "package:photos/theme/ente_theme.dart";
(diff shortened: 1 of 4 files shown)


mobile/apps/photos/lib/ui/viewer/gallery/component/gallery_file_widget.dart
```diff
@@ -212,11 +212,15 @@ class _GalleryFileWidgetState extends State<GalleryFileWidget> {
 
   void _routeToDetailPage(EnteFile file, BuildContext context) {
     final galleryFiles = GalleryFilesState.of(context).galleryFiles;
+    // Device folders are local-only contexts where files should remain visible
+    // even after deleting from Ente (remote)
+    final isLocalOnlyContext = widget.tag.startsWith("device_folder:");
```
ua741: Can there be better/future proof way do to this. If for any reason tag changes, this will result in unexpected behaviour

prateekmedia: Used GalleryType.localFolder instead.

prateekmedia: Fixed and tested.

PR #7404, opened 2025-10-09
[auth] Add minimize to tray on windows

tristan-derez:
## Description

This implements the minimize to tray feature requested in the discussion #7375 

image

### DIFF
--- a/mobile/apps/auth/lib/l10n/arb/app_en.arb
+++ b/mobile/apps/auth/lib/l10n/arb/app_en.arb
@@ -376,6 +376,7 @@
   "shouldHideCode": "Hide codes",
   "doubleTapToViewHiddenCode": "You can double tap on an entry to view code",
   "focusOnSearchBar": "Focus search on app start",
+  "minimizeToTrayOnClose": "Minimize to tray on close",
   "confirmUpdatingkey": "Are you sure you want to update the secret key?",
   "minimizeAppOnCopy": "Minimize app on copy",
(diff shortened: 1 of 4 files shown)


mobile/apps/auth/lib/ui/settings/general_section_widget.dart
```diff
@@ -147,6 +147,24 @@ class _AdvancedSectionWidgetState extends State<AdvancedSectionWidget> {
           ),
           sectionOptionSpacing,
         ],
+        if (Platform.isWindows) ...[
```
prateekmedia, line 150: Can we do this for all three desktop platforms?

prateekmedia: Can you try testing on mac or linux, if it works we can remove this flag.

prateekmedia: Just to be clear, we can merge the current PR as this is already an improvement for windows. At the same time, it would be awesome if this was not platform specific.

tristan-derez: I dont know how system tray works on linux/mac. Sadly I cant test since i dont have linux nor a mac

prateekmedia: Cool then, we can merge and test + extend later on.

PR #7433, opened 2025-10-10
[mob][photos] Swipe to Select ✨

ashilkn:
## Description

This PR introduces a swipe-to-select feature for the gallery, allowing users to quickly select multiple photos by swiping across them. The feature is behind a feature flag

## Tests

- Verified that it works for all galleries without bugs (an exception is MapPullUpGallery where auto-scroll doesn't work sometimes)
- Have reviewed the code myself
- Verified the feature doesn't work nor does any intensive code related to it run if behind feature flag.


ashilkn:
@claude You suggested this change:
> Recommendation: Simplify to:
if (flagService.internalUser && !limitSelectionToOne) {
  _allFilesWithDummies.addAll(groupFiles);
}

Could you make this change?

ashilkn:
@codex review

Issue #7154, opened 2025-09-12
Ente auth app crash on close


yagamikira88:
### Description

This was previously fixed by Ente Auth on their last update, but it appears after the latest Patch Tuesday from windows update, Ente Auth app is beginning to crash at exit once again. Windows Reliability History shows below:

Problem signature
Problem Event Name:	APPCRASH
Application Name:	auth.exe
Application Version:	4.4.3.443
Application Timestamp:	687dfac4
Fault Module Name:	flutter_windows.dll
Fault Module Version:	0.0.0.0
Fault Module Timestamp:	66e1eff3
Exception Code:	c0000005
Exception Offset:	000000000001d9f0
OS Version:	10.0.26100.2.0.0.256.48
Locale ID:	1033
Additional Information 1:	a2c0
Additional Information 2:	a2c05a2894fb1899a2080422d7cf8448
Additional Information 3:	671a
Additional Information 4:	671a075d4fe64771bc1ac9836d98ccce

### Version

4.4.3.443

### What product are you using?

Ente Auth

### What platform are you using?

Windows 11

kinghat:
i get a crash on close as well when using the flatpak on fedora:

<img width="398" height="110" alt="Image" src="https://github.com/user-attachments/assets/fdaf88ac-38bd-434f-bded-0c8b3a4fbe65" />

> Operating System: Fedora Linux 43
> KDE Plasma Version: 6.5.3
> KDE Frameworks Version: 6.20.0
> Qt Version: 6.10.1
> Kernel Version: 6.17.8-300.fc43.x86_64 (64-bit)
> Graphics Platform: Wayland

> $ flatpak run io.ente.auth 
> enteauth: /usr/lib/x86_64-linux-gnu/libcurl.so.4: no version information available (required by /app/share/enteauth/lib/libsentry.so)
> Gdk-Message: 10:15:16.083: Unable to load  from the cursor theme
>  [ente_logging][INFO] [2025-11-29 10:15:16.134323] log file for today: File: '/home/kinghat/.var/app/io.ente.auth/data/enteauth/logs/2025-11-29.txt' with prefix 
> [ente_logging][INFO] [2025-11-29 10:15:16.134323] app version: '4.4.10+476'
> [ente_logging][INFO] [2025-11-29 10:15:16.134323] 
>  [main][INFO] [2025-11-29 10:15:16.134444] Starting app in foreground
[...]

PR #7448, opened 2025-10-11
[auth] Fix crash on window close on Windows

### LINKED ISSUE #7419: Ente Auth crash on exit
### Description

Ente Auth will crash on exit (can be seen through Windows Reliability History) if you exit using the right-click method on the system tray.

But Ente Auth won't crash if you close the app by simply clicking the X button of the window.

### Version

4.4.4

Ente Auth

Desktop - Windows

tristan-derez:
## Description

closes #7419 and #7154

On Windows, closing the app window caused a crash. Explicitly terminating the process prevents this.

**Note:** This may be related to [flutter_inappwebview#2512](https://github.com/pichillilorenzo/flutter_inappwebview/issues/2512), though the exact cause is unclear.

**Solution:** Call `TerminateProcess()` on the current process handle during app closure on Windows.

## Tests

Tested locally on Windows.

### DIFF
--- a/mobile/apps/auth/lib/app/view/app.dart
+++ b/mobile/apps/auth/lib/app/view/app.dart
@@ -214,9 +214,10 @@ class _AppState extends State<App>
         if (Platform.isWindows) {
           final int hProcess = GetCurrentProcess();
           TerminateProcess(hProcess, 0);
+        } else {
+          windowManager.setPreventClose(false);
+          windowManager.destroy();
         }
-        windowManager.setPreventClose(false);
-        windowManager.destroy();
         break;
     }
   }
@@ -232,9 +233,10 @@ class _AppState extends State<App>
       if (Platform.isWindows) {
         final int hProcess = GetCurrentProcess();
         TerminateProcess(hProcess, 0);
+      } else {
+        windowManager.setPreventClose(false);
+        windowManager.destroy();
       }
-      windowManager.setPreventClose(false);
-      windowManager.destroy();
     }
   }
 }


mobile/apps/auth/lib/app/view/app.dart
```diff
@@ -210,6 +211,11 @@ class _AppState extends State<App>
         windowManager.setSkipTaskbar(false);
         break;
       case 'exit_app':
+        if (Platform.isWindows) {
+          final int hProcess = GetCurrentProcess();
+          TerminateProcess(hProcess, 0);
+        }
+        windowManager.setPreventClose(false);
         windowManager.destroy();
```
prateekmedia: Will these and similar lines in window destroy even run if you've killed the process?

mobile/apps/auth/lib/app/view/app.dart
```diff
@@ -223,6 +229,10 @@ class _AppState extends State<App>
       windowManager.hide();
       windowManager.setSkipTaskbar(true);
     } else {
+      if (Platform.isWindows) {
+        final int hProcess = GetCurrentProcess();
+        TerminateProcess(hProcess, 0);
+      }
       windowManager.setPreventClose(false);
       windowManager.destroy();
```
prateekmedia: Same as above.

mobile/apps/auth/lib/app/view/app.dart
tristan-derez: Nope, they wont run at all, but thats the point

prateekmedia: So are they even needed then?

prateekmedia: When Platform.isWindows

tristan-derez: You mean that you would like a else statement?

tristan-derez: Because apart if im wrong the process would only be killed on windows. windowManager would be prefered for other OS imo

prateekmedia: Yes like an else. So on windows it would always kill and on other's it would do normal.

tristan-derez: well its already doing that right now but i can add the else for clarity

prateekmedia: Cool, I'm fine with current approach as well.

prateekmedia: Thanks for resolving!

PR #7463, opened 2025-10-13
[mob][photos] [i] Add more logging for internal users

prateekmedia:
## Description

Only thing this PR does is add logging and only for internalUser for background upload related tasks

### DIFF
--- a/mobile/apps/photos/lib/main.dart
+++ b/mobile/apps/photos/lib/main.dart
@@ -23,7 +23,6 @@ import 'package:photos/core/errors.dart';
 import 'package:photos/core/network/network.dart';
 import "package:photos/db/ml/db.dart";
 import 'package:photos/ente_theme_data.dart';
-import "package:photos/extensions/logger_extension.dart";
 import "package:photos/extensions/stop_watch.dart";
 import "package:photos/l10n/l10n.dart";
 import "package:photos/service_locator.dart";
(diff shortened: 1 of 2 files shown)


mobile/apps/photos/lib/main.dart
```diff
@@ -177,27 +177,44 @@ Future<void> _runMinimally(String taskId, TimeLogger tlog) async {
 
     // Begin Execution
     // only runs for android
-    _logger.info("[BG TASK] update notification");
+    if (flagService.internalUser) {
+      _logger.info("[BG TASK] update notification");
+    }
```
ua741: Do we want to keep these changes behind internal flag?

mobile/apps/photos/lib/main.dart
```diff
@@ -177,27 +177,44 @@ Future<void> _runMinimally(String taskId, TimeLogger tlog) async {
...
     updateService.showUpdateNotification().ignore();
-    _logger.info("[BG TASK] sync starting");
+    if (flagService.internalUser) {
+      _logger.info("[BG TASK] sync starting");
+    }
```
ua741: Let's define an extension for Logger and call it internalLog that. We can have condition iniside that (if not internal, don't log otherwise log).

mobile/apps/photos/lib/main.dart
prateekmedia: Resolved.

mobile/apps/photos/lib/main.dart
prateekmedia: Resolved.

PR #7502, opened 2025-10-16
[mob][photos] Update widget image dimensions + cache

prateekmedia:
## Description
- resize widget images to a 1280px max dimension and encode as JPEG without upscaling
- cache these images

## Tests
- [ ] Test on android
- [ ] Test on iOS
- [ ] Test with different formats like HEIC


prateekmedia:
superseded by #7598

Issue #7493, opened 2025-10-15
ente-cli completion fails with "error getting password from keyring"


kastl-ars:
### Description

Dear maintainers,

I tried packaging the ente-cli for openSUSE. The build was fine in 0.2.1, the shell completions could be generated without errors during package build.

ente-cli 0.2.2 and 0.2.3 fail to generate the shell completions:

```
[   26s] + install -D -m 0755 bin/ente /home/abuild/rpmbuild/BUILD/ente-cli-0.2.3-build/BUILDROOT//usr/bin/ente
[   26s] + mkdir -p /home/abuild/rpmbuild/BUILD/ente-cli-0.2.3-build/BUILDROOT/usr/share/bash-completion/completions/
[   26s] + /home/abuild/rpmbuild/BUILD/ente-cli-0.2.3-build/BUILDROOT//usr/bin/ente completion bash
[   26s] 2025/10/15 07:06:44 error getting password from keyring: exit status 1
[   26s]           Refer to https://help.ente.io/self-hosting/troubleshooting/keyring
```

Is this intentional? Could the old behaviour be restored?

Kind Regards,
Johannes

### Version

v0.2.2 and v0.2.3

Blurazzle:
This is because you are running the Ente CLI headless, meaning without a desktop environment. Or ad-least a desktop environment without a keyring installed. The problem should resolve by installing the gnome keyring package from your distribution 

Alternatively, you can also use `ENTE_CLI_SECRETS_PATH` environment variable. In my environment the command to start Ente CLI looks something like this: 
> ENTE_CLI_SECRETS_PATH=~/.local/share/ente/secrets.txt ~/.local/share/ente/bin/ente

kastl-ars:
Thanks for the fast reply. I am not talking about the end user generating those completions. We want to have them packaged, so we are creating them during package build. This is offline, without any GUI running.

Regarding the `ENTE_CLI_SECRETS_PATH` path, is it enough for the target file to exist or does it need actual content?

Blurazzle:
I’m just a helpful internet stranger who noticed the “error getting password from keyring” subject line. In the past I’ve resolved that issue on my Raspberry Pi by setting the ENTE_CLI_SECRETS_PATH environment variable. Forgot to not read the rest of the issue and problem you try to resolve :).

I hope someone from Ente can help you.

PR #7505, opened 2025-10-16
Fix URL in Ente CLI error message for keyring issue

Blurazzle:
It seems the information to resolve the keyring error when using Ente CLI has moved to  https://ente.io/help/self-hosting/troubleshooting/cli

These docs might also help to resolve issue #7493.


mnvr (approved):
Thank you!

PR #7572, opened 2025-10-24
[mob] Change log

ua741:
## Description
Also fix regression in query for new -installs. Seems like additional comma was added by mistake, identified it via sentry logs.


ua741:
Verified that lint check passes locally. The workflow seems to be failing while trying to install rust.
```
flutter analyze --no-fatal-infos
```

PR #7574, opened 2025-10-24
[desktop] Disable differential downloads to fix Windows NSIS update issues

anandbaburajan:
https://github.com/electron-userland/electron-builder/issues/9181


mnvr:
Note that this won't fix the Windows update issue in the immediate next release (1.7.15) since the auto updater itself has the problem, so the changes will make an impact in the next to next release (1.7.16).

PR #7578, opened 2025-10-24
[infra] Harden headers

mnvr:


server/pkg/controller/file.go
```diff
@@ -802,21 +803,31 @@ func (c *FileController) cleanupDeletedFile(qItem repo.QueueItem) {
...
+func (c *FileController) getHotDcSignedUrl(objectKey string, objType ente.ObjectType) (string, error) {
 	s3Client := c.S3Config.GetHotS3Client()
-	r, _ := s3Client.GetObjectRequest(&s3.GetObjectInput{
+	input := &s3.GetObjectInput{
 		Bucket: c.S3Config.GetHotBucket(),
 		Key:    &objectKey,
-	})
+	}
+	input.ResponseContentDisposition = aws.String("attachment")
+	if objType == ente.FILE {
```
ua741: As discussed offline, can also set if for Thumbnail, rest looks good.

mnvr: Done

PR #7582, opened 2025-10-24
[mob] Skip background tasks when foreground is active

ua741:
## Description
Prevents race conditions and duplicate uploads between foreground and background processes by checking if foreground was recently active. Background tasks defer to foreground when it's running to avoid conflicts with upload URLs, file locks, and sync operations.

### DIFF
--- a/mobile/apps/photos/lib/main.dart
+++ b/mobile/apps/photos/lib/main.dart
@@ -129,16 +129,12 @@ Future<void> runBackgroundTask(
   String mode = 'normal',
 }) async {
   // Check if foreground is recently active to avoid conflicts
-  final prefs = await SharedPreferences.getInstance();
-  await prefs.reload();
-
-  final now = DateTime.now().microsecondsSinceEpoch;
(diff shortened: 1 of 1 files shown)


mobile/apps/photos/lib/main.dart
```diff
@@ -128,6 +128,28 @@ Future<void> runBackgroundTask(
...
 }) async {
+  // Check if foreground is recently active to avoid conflicts
+  final prefs = await SharedPreferences.getInstance();
+  await prefs.reload();
+
+  final now = DateTime.now().microsecondsSinceEpoch;
+  final lastFGBeat = prefs.getInt(kLastFGTaskHeartBeatTime) ?? 0;
+
+  // If FG was active in last 30 seconds, skip BG work
+  if (now - lastFGBeat < 30000000) {
```
prateekmedia: Should we use this check instead, as it seems to check fg heart beat and timeout:
_isRunningInForeground

Just like firebase push method is using.

ua741: Done. Also, it seems like we were not reporting bg heartbeat. Have also made that change.

prateekmedia (approved):

PR #7598, opened 2025-10-26
[mob][photos] optimize home widget image to 1024px

prateekmedia:
## Description

Previously home widget image was restricted to 512px because of thumbnail image. This PR changes to following:
- Image (remote/local): 1024px
- Video/Live Photo (local): 1024px
- Others or error case: 512px / Max Thumbnail Size

Behind a useWidgetV2 flag until it is proven to not have any regression.

### DIFF
--- a/mobile/apps/photos/scripts/internal_changes.txt
+++ b/mobile/apps/photos/scripts/internal_changes.txt
@@ -1,4 +1,6 @@
 - Prateek: (i) Optimize home widget image quality with V2 implementation (1024px vs 512px, uses PhotoManager thumbnails for local files, isolate processing for remote images)
+- Laurens: remove archived photos from memories
+- Neeraj: Fix thumbnail bg in native video player widget
 - Ashil: Fix full screen state getting disabled when swiping to video
 - Ashil: Remove long-press icon to switch to media kit player 
 - Neeraj & Laurens: fix OCR iOS issues and remove feature flag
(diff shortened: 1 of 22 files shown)


mobile/apps/photos/lib/services/home_widget_service.dart
```diff
@@ -182,6 +231,105 @@ class HomeWidgetService {
...
+      else if (file.fileType == FileType.image) {
+        final File? imageFile = await getFileFromServer(file);
+
+        if (imageFile == null) {
+          _logger.warning("Failed to get file for V2 widget ${file.displayName}");
+          return false;
+        }
+
+        // Read and validate in isolate to avoid blocking main thread
+        final Uint8List imageBytes = await compute(
```
ashilkn: This will spawn a new isolate. Shouldn't we use already existing pool of isolates that is spawned on app init by the `Computer` package? If it is much easier to spawn a new isolate here instead of using one from the pool, let it be.

mobile/apps/photos/scripts/internal_changes.txt
```diff
@@ -1,3 +1,4 @@
+- Prateek: Optimize home widget image quality with V2 implementation (1024px vs 512px, uses PhotoManager thumbnails for local files, isolate processing for remote images)
```
ashilkn: Missing `(i)` prefix.

mobile/apps/photos/lib/services/home_widget_service.dart
prateekmedia: Resolved.

mobile/apps/photos/scripts/internal_changes.txt
prateekmedia: Resolved.

PR #7617, opened 2025-10-27
[web] Specify content-length and md5 while requesting upload urls

ua741:


mnvr:
Notice for people who search on GitHub:

Newer Ente builds include a `Content-MD5` header on upload requests. If you've added an explicit CORS allowlist on your S3 provider instead of `["*"]`, **make sure the Content-MD5 header are present in `AllowedHeaders`**, otherwise preflight checks will fail, and uploads will be blocked.

PR #7624, opened 2025-10-28
Update index.md

Epicboy-Manas:
Updated old Authy ios MiTM guide with the new guide which is improved and more easy to use.


Epicboy-Manas:
Sure, that sounds good.

mnvr (approved):

PR #7678, opened 2025-10-30
[auth] Fix auth data key persistence on cold start

prateekmedia:
## Description
- Ensure the auth data key is persisted when the app cold starts

## Tests
- [x] Tested on Mac

### DIFF
--- a/mobile/apps/auth/lib/core/configuration.dart
+++ b/mobile/apps/auth/lib/core/configuration.dart
@@ -32,12 +32,21 @@ class Configuration extends BaseConfiguration {
     );
     sqfliteFfiInit();
     await _initOfflineAccount();
+    await _initOnlineAccount();
   }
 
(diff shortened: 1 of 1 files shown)


mobile/apps/auth/lib/core/configuration.dart
```diff
@@ -32,6 +32,7 @@ class Configuration extends BaseConfiguration {
     );
     sqfliteFfiInit();
     await _initOfflineAccount();
+    await _initOnlineAccount();
```
ua741, line 35: Was there a regression? I remember adding this method when i worked on offline mode.

prateekmedia: Yes! During common config change: https://github.com/ente-io/ente/blob/134314c2853431caa8129ae343a182013b09716c/mobile/apps/auth/lib/core/configuration.dart

PR #7731, opened 2025-11-03
[lock_screen] Make hide content setting independent of app lock

prateekmedia:
## Description

Hide content is independent to app lock toggle.

Affects photos, locker and auth.

## Tests

- [x] Tested on emulator, it works independently


ua741:
Was there any reason for putting it inside app lock before? Otherwise LGTM

ua741 (approved):

PR #7742, opened 2025-11-04
Implement sealed share secret flow

neeraj-pilot:
## Summary
- introduce 12-character base62 link secrets on mobile and encrypt file keys before link creation
- persist encrypted key material on server and expose it via /file-link/info for share clients
- update web share app to derive file keys using the new metadata while keeping legacy support


neeraj-pilot:
Superseded by #7753 with clean history.

PR #7744, opened 2025-11-04
[auth] Add import from gallery button to qr scanner screen

prateekmedia:
## Description

Adds Import from gallery to qr scanner screen.

https://github.com/user-attachments/assets/af95e91b-f239-4e9c-9610-50940dc5c1be

## Tests

- [x] Tested on emulator


ua741:
Instead of additional option, should we show import from Gallery on the scan page?

PR #7746, opened 2025-11-04
[photos] [auth] Add a warning re multi-app accounts for account deletion

anandbaburajan:


anandbaburajan:
@ashilkn same for locker :D

ashilkn:
@anandbaburajan Locker uses `mobile/packages/accounts/lib/pages/delete_account_page.dart` for deleting account. Since you've made changes to this file, it will reflect on locker too.

PR #7754, opened 2025-11-05
[auth] Fix custom order button to open rearrange screen on first tap

prateekmedia:
## Description

Previously, tapping "Custom order" from a different sort order would only switch to custom order but not open the rearrange screen. Users had to tap the sort button again and click "edit" beside Custom order to access it.

Fixed by removing the condition that checked if the current sort was already manual, so now selecting custom order immediately opens the rearrange screen.

## Tests

- [x] Tested on iOS sim


ua741:
closing as discussed offline

PR #7831, opened 2025-11-08
[auth] Add Android Quick Settings tile to launch the app

ExtendedGuru4883:
Users can now add a Quick Settings tile to quickly launch the app. This provides a convenient shortcut to access 2FA codes without searching for the app on the home screen or app drawer.

Changes:
- Android: Added QuickTileService to the manifest.
- Android: Implemented QuickTileService, the tile launches the main activity.
- Android: Included a monochrome icon for the tile.

## Description
This PR introduces a Quick Settings tile on Android to launch the Ente Auth app.

The following changes have been made:

- **AndroidManifest.xml** Registered the new QuickTileService.
- **QuickTileService.kt** Implemented the service that handles the tile's onClick event, which launches the MainActivity. The implementation handles the deprecated startActivityAndCollapse for backward compatibility.
- **drawable/ic_qs_monochrome.xml** Added a monochrome icon for the tile, as required for the Quick Settings panel.

Related discussion #1026 

## Tests
Physical Device and emulator on Android 16 (SDK 36)
Result: The tile is correctly displayed and launches the app successfully.

Emulator: Android 5.0 (SDK 21)
Result: This test was performed to ensure backward compatibility on Android versions below API 24 (which do not support Quick Settings tiles). The app runs without any crashes or issues. The new code is correctly bypassed on older versions.

https://github.com/user-attachments/assets/1b804753-907d-4b0f-a978-b66d1fa97f64

### DIFF
--- a/mobile/apps/auth/android/app/src/main/kotlin/io/ente/authenticator/QuickTileService.kt
+++ b/mobile/apps/auth/android/app/src/main/kotlin/io/ente/authenticator/QuickTileService.kt
@@ -11,6 +11,7 @@ import androidx.annotation.RequiresApi
 class QuickTileService : TileService() {
 
     override fun onStartListening() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) return
         val tile = qsTile
         tile.subtitle = "Open Ente Auth"
         tile.updateTile()


brog-io:
I wonder would it be possible to copy a specific code from quick tile?

ExtendedGuru4883:
> I wonder would it be possible to copy a specific code from quick tile?

Technically it should be possible, for example Bitwarden has a tile to autofill credentials. However this would require accessibility permissions to detect the active app or website when the tile is activated, and each entry in the authenticator would need to include its associated URLs.
In its current form, the tile still greatly improves the authentication flow: the user can use the tile to open the app and copy the code, and if the app is set to minimize on copy they are immediately back to the previous page to paste it.

ExtendedGuru4883:
> Is it possible for Ente Auth tile to look like Flashlight tile (basically without the right arrow at the end).

Here is how it looks now
<p>
  
  
</p>

prateekmedia:
@codex review

ExtendedGuru4883:
@codex review

prateekmedia (approved):

PR #7848, opened 2025-11-10
[auth] remove x icon

prateekmedia:
## Description

- Use the one from simple-icons, which switches color correctly based on theme.
- Also remove altname from Twitter custom icon to fix it switching to different one

## Tests

- [x] Tested on simulator

x_dark
x_light


ua741:
Can you add more details about it? This used to work before. Was there any regression?

PR #7904, opened 2025-11-13
[locker] Fix edit file name

AmanRajSinghMourya:

### DIFF
--- a/mobile/apps/locker/lib/utils/file_actions.dart
+++ b/mobile/apps/locker/lib/utils/file_actions.dart
@@ -151,14 +151,11 @@ class FileActions {
         }
       }
 
+      showToast(context, context.l10n.fileUpdatedSuccessfully);
+
       await CollectionService.instance.sync();
 
       await dialog.hide();
-
-      if (!context.mounted) {
-        onSuccess?.call();
-        return;
-      }
 
       onSuccess?.call();
     } catch (e) {
@@ -315,10 +312,10 @@ class FileActions {
       await dialog.hide();
 
       if (context.mounted) {
-      showToast(
-        context,
-        context.l10n.fileDeletedSuccessfully,
-      );
+        showToast(
+          context,
+          context.l10n.fileDeletedSuccessfully,
+        );
       }
 
       onSuccess?.call();


mobile/apps/locker/lib/utils/file_actions.dart
```diff
@@ -59,6 +60,28 @@ class FileActions {
     await dialog.show();
 
     try {
+      final currentTitle = file.displayName;
+      final currentCaption = file.caption ?? '';
```
vishnukvmd: We are not using caption anymore?

mobile/apps/locker/lib/utils/file_actions.dart
```diff
@@ -137,7 +160,12 @@ class FileActions {
       if (!context.mounted) {
         return;
       }
-      showToast(context, context.l10n.fileUpdatedSuccessfully);
+
+      if (hasMetadataChanged) {
+        showToast(context, context.l10n.fileUpdatedSuccessfully);
+      } else {
+        showToast(context, context.l10n.fileUpdatedSuccessfully);
```
vishnukvmd: We can skip the toast in this case?

AmanRajSinghMourya: Okay

mobile/apps/locker/lib/utils/file_actions.dart
```diff
@@ -294,10 +315,10 @@ class FileActions {
       await dialog.hide();
 
       if (context.mounted) {
-        showToast(
-          context,
-          context.l10n.fileDeletedSuccessfully,
-        );
+      showToast(
```
vishnukvmd: Unintentional formatting?

mobile/apps/locker/lib/utils/file_actions.dart
```diff
@@ -135,9 +156,9 @@ class FileActions {
       await dialog.hide();
 
       if (!context.mounted) {
+        onSuccess?.call();
         return;
       }
-      showToast(context, context.l10n.fileUpdatedSuccessfully);
```
vishnukvmd, line 140: Shouldn't we show this toast if metadata was updated successfully?

mobile/apps/locker/lib/utils/file_actions.dart
```diff
@@ -135,9 +156,9 @@ class FileActions {
       await dialog.hide();
 
       if (!context.mounted) {
+        onSuccess?.call();
```
vishnukvmd: I'm not sure what we're doing `onSuccess`, but the caller might not be aware that there's a dependency on `context` being mounted to receive a callback - and that might lead to bugs. If the mount state is not important, we could skip this check.

mobile/apps/locker/lib/utils/file_actions.dart
```diff
@@ -59,6 +60,26 @@ class FileActions {
...
     try {
+      final currentTitle = file.displayName;
+      final hasMetadataChanged = result.title != currentTitle;
+
+      if (hasMetadataChanged) {
+        _logger.info('Updating file metadata: title changed');
+        final metadataUpdateSuccess = await MetadataUpdaterService.instance
+            .editFileName(file, result.title);
+
+        if (!metadataUpdateSuccess) {
```
vishnukvmd, line 71: I think the code could be further simplified by
1. Moving `dialog.hide()` outside, since it has to be hidden regardless of the outcome, and 
2. Showing the toast if context is mounted (flipping the if condition), to reduce a return statement

PR #7906, opened 2025-11-13
[auth] add semantics

prateekmedia:
## Description

Fixes screen readers not working.

## Tests

- [x] Tested with voiceover


ua741 (approved):

prateekmedia:
From docs: This is only needed for web, for other platforms this is auto-enabled. So we can close the PR.
Reference: https://docs.flutter.dev/ui/accessibility/assistive-technologies

PR #7957, opened 2025-11-17
[locker] Locker download

AmanRajSinghMourya:
## Description
Switched download saving to use FileSaver.saveAs on both Android and iOS, which opens the system’s save/location picker.

## Tests
Tested on Android and iOS
https://wormhole.app/pAxDZd#dU5fJLimooFL8MPhy9vcVQ

### DIFF
--- a/mobile/apps/locker/lib/utils/file_util.dart
+++ b/mobile/apps/locker/lib/utils/file_util.dart
@@ -152,16 +152,17 @@ class FileUtil {
 
         final sanitizedName = _sanitizeFileName(file.displayName);
         final baseName = _baseNameWithoutExtension(sanitizedName);
-        final extension = _extensionWithoutDot(sanitizedName);
+        final fileExtension = _extensionWithoutDot(file.displayName);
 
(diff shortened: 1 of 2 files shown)


mobile/apps/locker/lib/utils/file_util.dart
```diff
@@ -131,40 +137,33 @@ class FileUtil {
...
+            );
           }
           continue;
         }
 
-        final targetFileName =
-            extension.isEmpty ? baseName : "$baseName.$extension";
+        final sanitizedName = _sanitizeFileName(file.displayName);
+        final baseName = _baseNameWithoutExtension(sanitizedName);
+        final extension = _extensionWithoutDot(sanitizedName);
```
ashilkn: This is a reserved keyword in dart, use a different name.

ashilkn:
Can remove `<uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />` at`mobile/apps/locker/android/app/src/main/AndroidManifest.xml`

PR #7992, opened 2025-11-18
[server] Fix file copy for objects larger than 5GB

ua741:
## Description
Copying large files was resulting in EntityTooLarge error.


ua741:
@codex review

Issue #7943, opened 2025-11-16
Ente Auth 4.4.10: extremely oversized icons and codes


Omoeba:
### Description

Since updating to Ente Auth 4.4.10 on macOS 26.1, some of my codes have grown absolutely enormous, taking up multiple pages and making the other codes impossible to access without searching.

<img width="1624" height="980" alt="Image" src="https://github.com/user-attachments/assets/8c8edbba-3442-44c9-af29-9ed09f6acb4d" />

<img width="1624" height="980" alt="Image" src="https://github.com/user-attachments/assets/b8926b8d-a8ff-4c04-84e2-c540b747f9dc" />

### Version

4.4.10

### What product are you using?

Ente Auth

### What platform are you using?

Desktop - macOS

prateekmedia:
Thanks for report, we're actively trying to solve this issue. We've created a beta release with increased logging, can you try that build and send us logs at `support@ente.io` (Drawer > Support > Report a bug > Export logs):
https://github.com/ente-io/ente/releases/download/auth-v4.4.11-beta/ente-auth-v4.4.11-beta.dmg

The extremely oversized icons seems to be from the Show large icons setting ( Drawer > General > Show large icons ).

Omoeba:
The issue occurs which the Show Large Icons settings disabled as well. I'll install the beta release and upload logs when I have the opportunity to do so.

prateekmedia:
We've identified the issue, it was related to the Parallels icon that had an issue, it's fixed and will be available soon in the beta build along with the next release as well.
https://github.com/ente-io/ente/releases/tag/auth-v4.4.11-beta

prateekmedia:
Duplicate of:
https://github.com/ente-io/ente/discussions/7878

PR #7995, opened 2025-11-18
[auth] add error handling to icon utils & icon picker

prateekmedia:
## Description

Wrap IconUtils method with try-catch

Also fixes the same issue in Custom icon picker

Resolves #7878 #7979 #7943 #7941 

## Tests

- [x] Verified in both home and icon picker page after this patch


mobile/apps/auth/lib/ui/utils/icon_utils.dart
```diff
@@ -176,38 +238,70 @@ class IconUtils {
...
+          (blue - red).abs() <= tolerance;
+    } catch (e, s) {
+      Logger("IconUtils").warning(
+        "Failed to check if color is neutral grey",
+        e,
+        s,
+      );
+      return false; // Assume not neutral grey on error
+    }
   }
```
ua741: We don't need try catch everywhere, Just top level getIcon and getAllIcons should be fine?

prateekmedia:
@ua741 Resolved!

PR #8005, opened 2025-11-19
[auth] Fix add tag

prateekmedia:
## Description

Only select intersection of tags and on deselection of intersected one deselect from all, on selection add to all.

### DIFF
--- a/mobile/apps/auth/lib/ui/home/add_tag_sheet.dart
+++ b/mobile/apps/auth/lib/ui/home/add_tag_sheet.dart
@@ -38,18 +38,13 @@ class _AddTagSheetState extends State<AddTagSheet> {
 
   Future<void> _loadInitialState() async {
     final allTagsFromServer = await CodeDisplayStore.instance.getAllTags();
-    final initialTagsForSelection = <String>{};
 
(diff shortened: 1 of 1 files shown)


mobile/apps/auth/lib/ui/home/add_tag_sheet.dart
```diff
@@ -41,24 +40,51 @@ class _AddTagSheetState extends State<AddTagSheet> {
...
-    for (final code in widget.selectedCodes) {
-      initialTagsForSelection.addAll(code.display.tags);
+    // Calculate intersection of tags from all selected codes
+    if (widget.selectedCodes.isNotEmpty) {
+      // Start with tags from the first code
+      initialTagsForSelection.addAll(widget.selectedCodes.first.display.tags);
+
+      // Keep only tags that exist in ALL selected codes
+      for (final code in widget.selectedCodes.skip(1)) {
+        initialTagsForSelection.retainAll(code.display.tags);
```
ua741: Is it possible to simplify the logic. These feels very hard to understand.

prateekmedia: Resolved.

PR #8123, opened 2025-11-26
[auth][custom_icons] Delete questrade.svg

prateekmedia:
## Description

This icon isn't visible in the auth app, so removing it.


4fosaefp:
Hi @prateekmedia, I'm not sure why this svg doesn't work, but I found an alternative. Do I create a new PR?

prateekmedia:
@4fosaefp you can, if it works we will merge that.

4fosaefp:
> @4fosaefp you can, if it works we will merge that.

Any idea what make a SVG valid or invalid? I have no idea why the other two icons in my original PR works compared to questrade icon...

Anyways, icon updated here https://github.com/ente-io/ente/pull/8124. Hope this one works

prateekmedia:
Generally it's rendering issue, dependent on plugin we're using

PR #8125, opened 2025-11-26
[server] Support for admin to remove files from collection & suggest deletion

neeraj-pilot:
## Description

We are now also updating collection_files.created_at value when an entry from collection is added after deletion.


neeraj-pilot:
@codex review , have applied your suggestions

PR #8176, opened 2025-12-01
[rust] Rust integration scaffold for both web/mobile

mnvr:


mnvr:
(the web lint is failing because I've not wired in `yarn build:wasm` in the lint flow so the generated files are not there)

ua741 (approved):

PR #8179, opened 2025-12-01
[auth] adds andOTP import support with encrypted backup handling

joicemjoseph:
Support importing authenticator codes from andOTP, including both plain text JSON and encrypted (.aes) backup files. The importer:

- Decrypts AES-GCM encrypted backups using PBKDF2 key derivation
- Handles TOTP, HOTP, and Steam codes
- Preserves tags from andOTP entries
- Shows which entries were skipped (e.g., unsupported MOTP)

ref: https://github.com/ente-io/ente/discussions/481#discussioncomment-10436994

### DIFF
--- a/mobile/apps/auth/lib/ui/settings/data/import/andotp_import.dart
+++ b/mobile/apps/auth/lib/ui/settings/data/import/andotp_import.dart
@@ -67,13 +67,10 @@ Future<void> _pickAndOTPFile(BuildContext context) async {
 
   try {
     String path = result.files.single.path!;
-    final importResult = await _processAndOTPFile(context, path, progressDialog);
+    final count = await _processAndOTPFile(context, path, progressDialog);
     await progressDialog.hide();
-    if (importResult != null) {
-      await importSuccessDialog(context, importResult.importedCount);
-      if (importResult.skippedEntries.isNotEmpty) {
-        await _showSkippedEntriesDialog(context, importResult.skippedEntries);
-      }
+    if (count != null) {
+      await importSuccessDialog(context, count);
     }
   } catch (e, s) {
     _logger.severe('Exception while processing andOTP import', e, s);
@@ -86,13 +83,6 @@ Future<void> _pickAndOTPFile(BuildContext context) async {
   }
 }
 
-class _ImportResult {
-  final int importedCount;
-  final List<String> skippedEntries;
-
-  _ImportResult({required this.importedCount, required this.skippedEntries});
-}
-
 class _DecryptParams {
   final Uint8List fileBytes;
   final String password;
@@ -100,7 +90,7 @@ class _DecryptParams {
   _DecryptParams({required this.fileBytes, required this.password});
 }
 
-Future<_ImportResult?> _processAndOTPFile(
+Future<int?> _processAndOTPFile(
   BuildContext context,
   String path,
   ProgressDialog dialog,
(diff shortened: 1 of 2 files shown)


mobile/apps/auth/lib/ui/settings/data/import/andotp_import.dart
```diff
@@ -0,0 +1,295 @@
...
+    );
+  }
+}
+
+class _ImportResult {
+  final int importedCount;
+  final List<String> skippedEntries;
+
+  _ImportResult({required this.importedCount, required this.skippedEntries});
+}
```
prateekmedia: Why to introduce a new type for this?

prateekmedia (changes requested):

prateekmedia:
I tested the PR, the progress dialog animation is getting blocked, probably as you're doing too much work in main thread and it is causing a jank.

Can you check how other imports work and introduce isolate or something to do the heavy work for decryption so the progress dialog's animation is not causing jank.

mobile/apps/auth/lib/ui/settings/data/import/andotp_import.dart
joicemjoseph: AndOTP supports the legacy [MOTP](https://motp.sourceforge.net/) format, which Ente Auth doesn't currently support. 

When encountering unsupported types, other importers like Aegis, which 
  [prevents](https://github.com/ente-io/ente/blob/5fe4ff3e1973843fcead17c02057789672a79954/mobile/apps/auth/lib/ui/settings/data/import/aegis_import.dart#L173) users from importing any of their codes - even the valid ones.

My approach here is to import all supported OTP types and gracefully skip the unsupported ones, while [informing](https://github.com/ente-io/ente/commit/efbaaad3a39acdbac371e963f2901a7edf98bb4f#diff-bf29f8f82f3da5dd11a5babea6a6427acf6243b305f72e25d7b9c05c6922782fR275-R295) users which entries couldn't be imported. This way they don't lose access unknowingly.

  That said, I'm open to suggestions! If you'd prefer a different pattern, let me know.

joicemjoseph:
@prateekmedia Thank you very much for the review.

I have moved decryption logic from the main thread - now, the animation is smooth.

mobile/apps/auth/lib/ui/settings/data/import/andotp_import.dart
prateekmedia: Okay. Seems good then.

mobile/apps/auth/lib/ui/settings/data/import/andotp_import.dart
```diff
@@ -0,0 +1,310 @@
...
+
+  try {
+    String path = result.files.single.path!;
+    final importResult = await _processAndOTPFile(context, path, progressDialog);
+    await progressDialog.hide();
+    if (importResult != null) {
+      await importSuccessDialog(context, importResult.importedCount);
+      if (importResult.skippedEntries.isNotEmpty) {
+        await _showSkippedEntriesDialog(context, importResult.skippedEntries);
+      }
```
prateekmedia: Will this trigger two consecutive dialogs, if skipped entries are present?

Ideally only only dialog informing what were imported and optionally what were skipped is fine.

joicemjoseph: Makes sense. I will revert the changes. I will just show the imported otps like other importers.

joicemjoseph:
I've simplified the import, matching other importers. Removed showing skipped entries.

PR #8257, opened 2025-12-06
[mob][photos] Fix backups out of sync

prateekmedia:


prateekmedia:
@ua741 Here I have used both skipping while polling next in queue and also travesing through all and dropping based on deselection.

Let me know if I should only keep the polling queue skip logic and remove dropping all from queue that doesn't match new criteria at once.

prateekmedia:
@ua741 can you review?

PR #8287, opened 2025-12-08
[workflow] Update macOS runner to use latest version

prateekmedia:
## Description

Fixes 
```
The macOS-13 based runner images are now retired. For more details, see https://github.com/actions/runner-images/issues/13046.
```

## Tests

Triggered here: https://github.com/ente-io/ente/actions/runs/20033248420

### DIFF
--- a/.github/workflows/auth-release.yml
+++ b/.github/workflows/auth-release.yml
@@ -275,7 +275,7 @@ jobs:
                   updateOnlyUnreleased: true
 
     build-macos:
-        runs-on: macos-latest
+        runs-on: macos-15
 
         defaults:
             run:


.github/workflows/auth-beta-release.yml
```diff
@@ -332,7 +332,7 @@ jobs:
                   path: mobile/apps/auth/artifacts/*
 
     build-macos:
-        runs-on: macos-13 # latest is 12
+        runs-on: macos-latest
```
ua741: Instead of latest, can we pin specific version to avoid any surprise related to builds?

prateekmedia: Done!

PR #8306, opened 2025-12-10
[server] Like and Comment

neeraj-pilot:


server/pkg/controller/collections/collection.go
```diff
@@ -214,6 +218,9 @@ func (c *CollectionController) HandleAccountDeletion(ctx context.Context, userID
 		if err != nil {
 			return stacktrace.Propagate(err, "")
 		}
+		if cleanupErr := c.removeUserSocialActivity(ctx, shareCollection.CollectionID, shareCollection.ToUserID); cleanupErr != nil {
+			return cleanupErr
+		}
 	}
```
ua741, line 224: Test this flow

PR #8309, opened 2025-12-10
[mob][photos] Fix iOS back navigation for Trash and other page

prateekmedia:
## Description

Now we calculate canPop before so back gesture works on iOS.

## Tests

- [x] Test on simulator
- [x] Test on emulator


ua741:
Please check if the behaviour works fine on android.

PR #8318, opened 2025-12-10
[mob][photos] Fix route shifting when bottom sheet appears

prateekmedia:
## Description

Fixes the screen shifting to left on iOS when bottom sheet appears.

## Tests

- [x] Tested on iOS sim


ashilkn:
@prateekmedia  Setting `rootNavigator = true` by default will  close the app instead of closing the modal when system back gesture is used. See https://github.com/ente-io/ente/commit/0a8d403ea0c47e80105ac74a76fe667dc0c8e2f1

ashilkn (approved):

PR #8335, opened 2025-12-11
Fix Ente Auth Wayland app_id vs desktop file mismatch

### LINKED ISSUE #4410: Icon in Ubuntu Dock is generic gear wheel
### Description

I have pinned ente auth to the dock in Ubuntu 24.04. When I click the icon, the applications launches. Instead of a dot right next to the icon, a generic icon is shows up for the running application.

Current state:
image

How it should look:
image

The problem seems to be fixed when adding the following line to enteauth.desktop:

`StartupWMClass=Io.ente.auth`

### Version

v4.1.6

Ente Auth

Desktop - Linux

siskinsec:
**Summary**

This PR fixes the duplicate “gear” icon issue in GNOME Wayland when running Ente Auth by aligning the installed desktop file name with the Wayland app_id.

**Background**

When running Ente Auth with Wayland debug enabled:

```
WAYLAND_DEBUG=1 enteauth 2>&1 | grep -i app_id
# xdg_toplevel@32.set_app_id("io.ente.auth")
```

The Wayland app_id is:

`io.ente.auth`

However, the Linux packaging currently installs a desktop file named:

`enteauth.desktop`

GNOME Wayland requires the desktop file basename to match the app_id. Because the names don’t match, GNOME treats the running window as a separate application, resulting in the generic gear icon.

**Changes**

Renamed:
enteauth.desktop → io.ente.auth.desktop

Updated packaging:
Modified build_rpm.sh to install the new desktop filename.

Updated AppStream data:
<launchable> and <id> entries now reference io.ente.auth.desktop

**Result**

After this change, Ente Auth correctly appears under its own icon in GNOME Wayland, with no duplicate gear icon.

Tested on:

Ubuntu 24.04
GNOME 46 on Wayland

**Fixes #4410**


prateekmedia:
Have commented on the original issue.

PR #8341, opened 2025-12-12
[Docs]Add FAQ for Google Photos mobile import

Shanthy17:


Shanthy17:
Is this FAQ appropriate? Would it make more people choose mobile and then lead to issues in future?

ua741:
May be we should not recommend using mobile for Google takeout upload.  Only desktop/web supports metadata parsing from sidecar json file. If someone has uploaded on google photos in compressed form, we will end up importing those with incorrect date.

PR #8355, opened 2025-12-13
[auth] Add US Army Icon

NateroniPizza:
The Army's DS Login site just gives a raw "secret" - no TOTP code, so don't have an issuer associated. Ideally this would remain a generic "US Army" icon, rather than specific to DS Login, for whenever they change MFA sites in the future.

title: US Army

Apologies if there's anything wrong about this PR - I've only done a couple before, and the last one was years ago.

### DIFF
--- a/mobile/apps/auth/assets/custom-icons/_data/custom-icons.json
+++ b/mobile/apps/auth/assets/custom-icons/_data/custom-icons.json
@@ -1840,6 +1840,10 @@
       "title": "Upstox"
     },
     {
+      "title": "US Army",
+      "slug": "us_army"
+    },
+    {
       "title": "US Mobile",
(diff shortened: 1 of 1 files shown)


AmanRajSinghMourya:
https://github.com/ente-io/ente/blob/main/mobile/apps/auth/docs/adding-icons.md 
Please add the Icon name to the custom-icons.json as well.

NateroniPizza:
> https://github.com/ente-io/ente/blob/main/mobile/apps/auth/docs/adding-icons.md Please add the Icon name to the custom-icons.json as well.

Sorry, missed that one - done.

AmanRajSinghMourya (approved):

PR #8401, opened 2025-12-17
[mob][photos] Christmas animation

AmanRajSinghMourya:


mobile/apps/photos/lib/ui/settings/advanced_settings_screen.dart
```diff
@@ -157,6 +159,24 @@ class AdvancedSettingsScreen extends StatelessWidget {
                           },
                         ),
                       ),
+                      const SizedBox(height: 24),
+                      MenuItemWidget(
```
ashilkn: We should only show this when the banner is shown by default.

PR #8412, opened 2025-12-17
[mob][photos] Hide on device section when permission skipped

prateekmedia:
## Description

- Hide on device collection when permission was skipped


prateekmedia:
https://github.com/ente-io/ente/pull/8420

PR #8420, opened 2025-12-18
[mob][photos] empty state ui for skipped + on device flow

prateekmedia:
## Description

On device section updated when permission was skipped.


prateekmedia:
Have to use styled text instead of text separation.

PR #8483, opened 2025-12-23
[mob][locker] Allow account creation  

AmanRajSinghMourya:

### DIFF
--- a/mobile/packages/accounts/lib/pages/password_entry_page.dart
+++ b/mobile/packages/accounts/lib/pages/password_entry_page.dart
@@ -434,6 +434,9 @@ class _PasswordEntryPageState extends State<PasswordEntryPage> {
 
   void _updatePassword() async {
     final logOutFromOthers = await logOutFromOtherDevices(context);
+    if (logOutFromOthers == null) {
+      return;
+    }
     final dialog =
         createProgressDialog(context, context.strings.generatingEncryptionKeys);
     await dialog.show();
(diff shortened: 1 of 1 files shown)


mobile/apps/auth/pubspec.yaml
```diff
@@ -21,34 +21,64 @@ dependencies:
   device_info_plus: ^9.1.1
   dio: ^5.8.0+1
   dir_utils:
-    path: ../../packages/dir_utils
+    git:
+      url: https://github.com/ente-io/ente.git
```
ua741: why we are adding dependency on commit ? it will close clone the whole repo for building.

mobile/apps/locker/lib/services/trash/trash_service.dart
```diff
@@ -192,6 +194,14 @@ class TrashService {
...
         latestUpdatedAtTime,
       );
+    } on DioException catch (e, s) {
+      _logger.severe(e, s);
+      if (e.response?.statusCode == 401) {
+        _logger.info("Session expired in trash sync, triggering logout");
+        Bus.instance.fire(TriggerLogoutEvent());
+        throw UnauthorizedError();
+      }
+      rethrow;
```
ua741: Why is this require here? Ideally, during collections diff sync itself or remote sync we should get this exception and handle it at that place instead of doing this at every api place

mobile/apps/auth/pubspec.yaml
AmanRajSinghMourya: This is done to avoid breaking for auth as the redesign is done on the `packages/account` which is used for auth as well. So I pinned these packages to main

ua741: This feels like bad design as it breaks the primary purpose of having common packages. It completely defeats the purpose of having common packages. 
Let's discuss this in person

PR #8570, opened 2026-01-01
[mob][locker] Add multiselect support for files and collections

AmanRajSinghMourya:
## Summary
- Add file multiselect on HomePage and CollectionPage with scroll-responsive overlay
- Add collection multiselect on AllCollectionsPage with ownership checks
- Add ownership filtering for shared files/collections


jackxu0302:
I think the option should be added

PR #8611, opened 2026-01-05
[mob][locker] Add support for language selection

AmanRajSinghMourya:


ua741:
In the actual selection, what all languages we are planning to show? We only enable a language once we have 90% strings that are translated.

AmanRajSinghMourya:
> In the actual selection, what all languages we are planning to show? We only enable a language once we have 90% strings that are translated.

For now we can start with English and will keep adding languages which are 90% translated.

ua741 (approved):

Issue #8561, opened 2025-12-31
Icon fails to update after rename Issuer


corobin:
### Description

Hello,

### Summary

Under certain circumstances, after you edit an existing entry and rename the Issuer, the entry icon does not update to reflect the new name.

It appears to be failing when two conditions are met:
1. There is an icon for the new issuer name, AND
2. It is the first time that this entry is renamed

### Tests and results

For convenience, I will use these two short hands:
KNOWN - can be any name that there IS an icon for
UNKNOWN - can be any name that there IS NOT an icon for
the exact name does not matter besides that.

**Test 1**

1. Create a new entry with UNKNOWN issuer
2. Edit entry and rename to KNOWN issuer

Result: icon does *not* automatically update

3. Then edit entry again to KNOWN issuer

Result: icon does automatically update

**Test 2**

1. Create a new entry with UNKNOWN issuer
2. Edit entry and rename to a different UNKNOWN issuer (with different first letter)

Result: icon does automatically update

**Test 3**

1. Create a new entry with KNOWN issuer
2. Edit entry and rename to a different KNOWN issuer

Result: icon does *not* automatically update
![Image](https://github.com/user-attachments/assets/65a16031-aaf0-40e4-815f-19a1972f3bda)

3. Then edit entry again to KNOWN issuer

Result: icon does automatically update

**Test 4**

1. Create a new entry with KNOWN issuer
2. Edit entry and rename to UNKNOWN issuer

Result: icon does automatically update

3. Then edit entry again to KNOWN issuer

Result: icon does automatically update

### Analysis

As you can see, my tests were not *entirely* exhaustive. However, what I have tested seem to suggest that the failure only occurs when the issuer is renamed TO something that there is an icon for, and only the first time that a particular entry is renamed. Where there is not an icon for the issuer, it appears to update every time, and subsequent renames after the first time also appear to work.

### Version

4.4.15

### What product are you using?

Ente Auth

### What platform are you using?

Mobile - Android

devbindal:
opened a pr for this
if found relevant assign this to me 😊

PR #8612, opened 2026-01-05
fix(icon-update): fix icon not updating on changing issuer

devbindal:
* fix icon not updating on changing issuer

there were two contradicting if conditions
issue link : https://github.com/ente-io/ente/issues/8561

before issuer XYZ
Screenshot_1767604691
renamed to google
Screenshot_1767604820


devbindal:
heyy any update..?

PR #8634, opened 2026-01-07
[web][photos] Adding support for downloading files as zip

AswinAsok:
## Description

Albums now are being downloaded as a group of zip files, unless it's a single file.

image
image
image

### DIFF
--- a/web/apps/photos/src/components/DownloadStatusNotifications.tsx
+++ b/web/apps/photos/src/components/DownloadStatusNotifications.tsx
@@ -17,32 +17,13 @@ import { Notification } from "ente-new/photos/components/Notification";
 import { t } from "i18next";
 
 /** Maximum characters for album name before truncation */
-const MAX_ALBUM_NAME_LENGTH = 30;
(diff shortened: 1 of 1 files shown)


anandbaburajan:
- Lint is failing
- Current naming is: Himalayas On Cycle - 1K KM to Leh (17 files)-4. Let's do Himalayas On Cycle - 1K KM to Leh (Part 4 - 17 files) and the first one should have "Part 1" as well
- I tried download two albums, and in both of them, a few files failed. I clicked the retry button in both of them and the failed files in both of them successfully downloaded. Which makes me curious, why were the files failing at all? I briefly checked those few files, and didn't find anything weird about them. Note that the two albums were large (200+ photos)
- I tried one of those albums again and didn't get any failed files :D
- When clicking download on an album, the notification/toast says "Downloading ... 34/232 files" but since nothing is being downloaded until the zip is ready, it might be confusing. We could say "Processing" maybe? Not sure

anandbaburajan (approved):

Issue #333, opened 2023-11-08
A concept of Trash "location" for removed items


x80486:
## Description

I would like to see the addition of a `Trash` location for removed items. Currently, when we remove items, they are deleted permanently, which can be problematic if someone accidentally removes an important item.

Having this would offer an added layer of security, ensuring that users can easily recover mistakenly removed items — as it just happened to me :grimacing:

## Describe the Solution You Would Like

A locations where removed items are placed before actually removing them permanently.

## Describe Alternatives You Have Considered

N/A

## Additional Context

Some notes/observations from other applications:

- Users should be able to restore items from the `Trash`
- Items in the `Trash` might be automatically deleted after a certain period to manage storage — not really a pleasant setting, but could be useful for some

PR #8670, opened 2026-01-09
[web][photos] QoL Added settings to clear large files.

AswinAsok:


anandbaburajan:
- Deleting should show a confirmation dialog
- If I select a few photos, then want to select all photos, no option to do that
- When I click on Large files / Deduplicate files, there's a flicker where I see the main sidebar with the green box, then page opens
- Select all button should be Select all / Select all photos / Select all videos based on All / Photos / Videos, otherwise could be confusing
- The sort button interaction is confusing. Let's do a dropdown (clicking on the sort icon) with Newest, Oldest and Largest (default)?
- Just curious, is the logic for finding large files same/similar to what we do on mobile, or is it something different?

anandbaburajan:
I asked Claude for a review. Not sure if all these make sense; you can just give all this to Claude and then decide. 

Issues & Suggestions

  1. Potential bug: State dependency in useEffect (large-files.tsx:306-316)

  useEffect(() => {
      dispatch({ type: "analyze" });
      void findLargeFiles(state.filter)  // ❌ uses state.filter
          .then(...)
          .catch(...);
  }, [state.filter]);  // filter in deps, but state.filter is stale in closure
  This works because state.filter changes trigger a re-run, but the closure captures the old state object. Consider passing filter directly:

  const filter = state.filter;
  useEffect(() => {
      dispatch({ type: "analyze" });
      void findLargeFiles(filter)
          .then(...)
          .catch(...);
  }, [filter]);

  2. Missing dependency in useCallback (large-files.tsx:318-330)

  const handleDeleteFiles = useCallback(() => {
      dispatch({ type: "delete" });
      void deleteSelectedLargeFiles(state.largeFiles, ...)  // uses state.largeFiles
          ...
  }, [state.largeFiles, onGenericError]);  // ✓ correct deps, but dispatch is missing
  dispatch should technically be in the dependency array (though React guarantees it's stable).

  3. Mutation of array item (large-files.ts:1413-1415)

  case "toggleSelection": {
      const largeFiles = [...state.largeFiles];
      const item = largeFiles[action.index]!;
      item.isSelected = !item.isSelected;  // ❌ mutates the item object
  This mutates the original item. Should create a new object:

  const item = { ...largeFiles[action.index]! };
  item.isSelected = !item.isSelected;
  largeFiles[action.index] = item;

  4. Non-null assertion without bounds check (large-files.ts:1414)

  const item = largeFiles[action.index]!;
  If action.index is out of bounds, this will fail silently. Consider adding validation.

  5. Inefficient iteration in deleteCompleted (large-files.ts:1477-1489)

  const removedFileIDs = new Set(
      state.largeFiles
          .filter(({ id }) => action.removedIDs.has(id))
          .map(({ file }) => file.id),
  );
  Two iterations where one would suffice. Could be:
  const removedFileIDs = new Set<number>();
  const largeFiles = state.largeFiles.filter(({ id, file }) => {
      if (action.removedIDs.has(id)) {
          removedFileIDs.add(file.id);
          return false;
      }
      return true;
  });

  7. Edge case: Empty progress state (large-files.tsx:890-893)

  variant={deleteProgress === 0 ? "indeterminate" : "determinate"}
  The progress starts at 0 in the "delete" action, which shows indeterminate. However, tickProgress immediately increments, so the indeterminate state is very brief. This is fine but could be confusing.

  8. Hardcoded color values (large-files.tsx:1010, 1025, 1053)

  background-color: #ddd;
  border: solid #333;
  border-color: #ddd;
  Should use theme colors for consistency:
  background-color: ${theme.vars.palette.common.white};

  ---
  Minor/Nitpick

  1. Inconsistent comparison operators: Mix of == and === (e.g., large-files.ts:1189 uses ==, large-files.ts:1207 uses !=)
  3. HoverOverlay is defined but never effectively visible - opacity is always 0 and no hover state changes it in the current code
  4. Translation key inconsistency: delete_files_button vs remove_duplicates_button_count - different naming patterns

anandbaburajan (approved):

PR #8673, opened 2026-01-09
[web][photos] QoL Added-support-for-asc-and-desc

AswinAsok:
## Description
Added new options so that users can sort in descending manner also.

image


anandbaburajan:
Let's do:

A-Z
Newest
Recently updated

Rest of those are very rare use cases, and if someone really wants that, they can just scroll to the end.

ua741:
Can't se use UP/Down icon to indicate current sort? If user Click on Name sorting show (up arrow).. If they again click on name, show down arrow and reverse the sorting order. This is what we are doing on mobile.

anandbaburajan:
@ua741 I think reverse of "A-Z", "Newest" and "Recently updated" are very rare use cases, and if someone wants that for one-off actions, they can just scroll to the end. Also @AswinAsok, thinking about it more, "Newest" is also rare - "Recently updated" covers most of what people actually need when looking for something they were just working on.

anandbaburajan (approved):

PR #8689, opened 2026-01-11
[mobile][auth] Fix keyboard not appearing on search bar auto-focus

### LINKED ISSUE #279: [Bug] Focus search on app start doesn't make the keyboard popup
When opening auth the search bar is focused, but my keyboard doesn't popup. It shows up once I tap the search bar, but that makes the feature somewhat useless. I'm using an Android device and the Samsung keyboard

AmanRajSinghMourya:
On some Android devices (particularly Samsung keyboard), the TextField's autofocus property doesn't reliably trigger the soft keyboard to appear. This fix adds an explicit requestFocus() call after the widget tree is built, which reliably shows the keyboard on all devices.

Fixes #279

### DIFF
--- a/mobile/apps/auth/lib/ui/custom_icon_page.dart
+++ b/mobile/apps/auth/lib/ui/custom_icon_page.dart
@@ -39,6 +39,15 @@ class _CustomIconPageState extends State<CustomIconPage> {
     _showSearchBox = _autoFocusSearch;
     searchBoxFocusNode = FocusNode();
     ServicesBinding.instance.keyboard.addHandler(_handleKeyEvent);
+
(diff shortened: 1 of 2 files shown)


ua741:
Does this fix the issue in release mode? IIRC, there were certain fixes that work on debug, but not in the release variant.

AmanRajSinghMourya:
I did some more changes and it does work on app start but if I switch the app then the keyboard does not pops up. Happens on Android devices mostly Samsung. There are issues open on flutter repo [#122994](https://github.com/flutter/flutter/issues/122994) [#52599](https://github.com/flutter/flutter/issues/52599)

I will keep a track on flutter issues and check for any fix, closing this now

PR #8729, opened 2026-01-13
[Docs]Add note about swipe to select settings

Shanthy17:


docs/docs/photos/features/utilities/swipe-to-select.md
```diff
@@ -7,6 +7,8 @@ description: Select multiple photos at once by swiping across them in the galler
 
 Swipe to Select lets you quickly select multiple photos by dragging your finger across the gallery, making it easy to organize large collections of photos.
 
+> **Note:** Swipe to select can be enabled or disabled at Settings > General > Gallery > Swipe to select.
```
ashilkn, line 10: Let's not add this. I'll be removing this option soon. 

Enabling swipe to select introduced issues on Samsung S series and this is the only range of devices that are affected as far as we know. The toggle was added in case this issue exists on any other devices and if a user reports it we could ask them to use this toggle. 

I'll be soon removing this toggle so I don't think docs for it is necessary.

Shanthy17: okay

PR #8733, opened 2026-01-13
[auth] remove flutter submodule + update fdroid icon

prateekmedia:
## Description

- Remove flutter submodule as it isn't needed for fdroid now, similar to photos.
- Update fdroid icon yaml to match pubspec.yaml

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,10 +2,6 @@
 	path = auth/thirdparty/sentry-dart
 	url = https://github.com/ente-io/sentry-dart.git
 	branch = sentry_flutter_ente
-[submodule "auth/flutter"]
-	path = mobile/apps/auth/flutter
-	url = https://github.com/flutter/flutter.git
-	branch = stable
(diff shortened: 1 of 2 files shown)


ua741:
Is the flutter version pinned on fdroid or it uses flutter stable? On auth, I can cause unexpected issues, and risk might be little higher than photos. Just thinking out loud about this.

prateekmedia:
@ua741 It uses similar approach to photos fdroid build, where we grep the flutter version from the auth-release.yml, so whatever we use for building release auth will be same there.

We checkout stable first then we switch to that pinned flutter version in fdroid.

ua741:
> @ua741 It uses similar approach to photos fdroid build, where we grep the flutter version from the auth-release.yml, so whatever we use for building release auth will be same there.
> 
> 
> 
> We checkout stable first then we switch to that pinned flutter version in fdroid.

Awesome. I wasn't aware that it is happening this way for photos. Makes sense.

PR #8735, opened 2026-01-13
[server] Add fileCount to FamilyMember API response

AmanRajSinghMourya:
Returns fileCount for each family member in the user details API.


ua741:
Let's call this endpoint as usage, in the controller, request request for any other app except locker.

PR #8740, opened 2026-01-13
Add import-google-takeout-folder CLI Command to Ente cli

korjavin:
## Summary
This PR adds a new CLI command to import Google Takeout exports directly to Ente Photos. It preserves metadata (timestamps, location, etc) from Takeout JSON sidecars and includes built-in duplicate detection.

## Motivation
Many users manage their media on headless environments like a NAS or a home server. Currently, importing large Google Takeout exports requires a GUI (Desktop/Web client). This command provides a robust, command-line alternative for these environments, allowing for efficient, unattended imports.

## Key Implementation Details
Duplicate Detection
Duplicate detection is implemented and enabled by default. It follows the logic used in the web/desktop clients:

Content Hash: Computes a BLAKE2b hash of the file content.
Metadata Comparison: Before uploading, the command fetches existing file metadata from the target album and compares hash + title + fileType.
Efficiency: This ensures that interrupted imports can be resumed without re-uploading existing files, and multiple imports from the same source don't create clusters of duplicates.
Chunked Encryption for Large Files
To ensure compatibility with the web client's decryption process for large files, we've implemented 
EncryptChaCha20poly1305Chunked()
. This function encrypts data in 4MB chunks (matching streamEncryptionChunkSize), which resolves "Failed to fetch" errors during playback of large videos.

## Testing
Tested on macOS with:

Large-scale collection: Imported a collection of ~6,400 files.
Deduplication reliability: Interrupted and restarted the import multiple times at random intervals; verified that the duplicate detection correctly identified and skipped already-uploaded files, resuming the process seamlessly.
Resilience: Encountered and successfully handled transient 500 errors from the storage provider (Wasabi) via built-in retries, with all files eventually being uploaded correctly.
Media types: Verified playback of 100MB+ videos and visibility of diverse JPG/PNG/HEIC files.
Usage
# Basic usage
ente import-google-takeout-folder /path/to/takeout "Album Name"
# Enable debug output
ente import-google-takeout-folder /path/to/takeout "Album Name" --debug

## Questions for Maintainers
FileType Requirement
While implementing this, I noticed the Zod schema for file metadata seems to require a fileType field (0 for image, 1 for video), otherwise files can remain invisible in the web client. I have implemented this as a requirement in the CLI metadata - could you clarify if this is the intended behavior or if there's a more flexible way to handle this?

### DIFF
--- a/cli/cmd/import.go
+++ b/cli/cmd/import.go
@@ -249,7 +249,7 @@ func runImport(folderPath, albumName string, dryRun, skipMetadata bool) error {
 
 		var metadata *uploader.SimpleFileMetadata
 		if !skipMetadata {
-			if takeoutMeta := scanResult.GetMetadataForFile(fi.FileName, int(collection.ID), pathPrefix); takeoutMeta != nil {
+			if takeoutMeta := scanResult.GetMetadataForFile(fi.FileName, dummyCollectionID, pathPrefix); takeoutMeta != nil {
 				metadata = &uploader.SimpleFileMetadata{}
 				if takeoutMeta.CreationTime != nil {
 					metadata.CreationTime = *takeoutMeta.CreationTime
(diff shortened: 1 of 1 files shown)


ua741:
Hey @korjavin 
Thank you for the PR. Unfortunately, at this point we won't have the bandwidth to regularly maintain and test the import flow in the CLI. Also, at all other places we are currently using the OG libsodium library for any encryption during upload. This changes that, which would require more demand and attention from the engineering team to validate and maintain it.

That said, we are slowly moving some core operations to Rust. Once things are stabilized in that world, it would be easy for us to build the import flow via the CLI.

PR #8764, opened 2026-01-14
Update email.go to add Date: ... header

ube80:
The `Date:` email header is actually required.

Emails without the `Date:` header may be considered malformed by some content scanners and not get delivered.

This adds a `Date:` header in the correct format to emails.


ube80:
Oops

PR #8889, opened 2026-01-22
[mob][photos] Remove redundant spinner

laurens-pilot:
## Description

Remove redundant spinner found during testing.


laurenspriem:
I personally feel it's mostly redundant, as the refresh doesn't take that long and in the meanwhile the search tab is still usable. 

Plus with the change of position of the search tab it's not clear where the spinner should be, the current position now feels off. So even if we're not removing it we at least have to change its position then.

Plus on app startup when you navigate to the search tab the user sees two different spinners now, which is quite weird.

Let's discuss in office once and take a call.

Issue #8906, opened 2026-01-24
Rituals show photos from hidden album


misuto:
### Description

When creating a new ritual along with a new album, and then hiding the album. The photo is correctly added to the hidden album but also showed in the ritual. This can be confusing for users since users expect that the album is hidden, but recent photos are shown in a "unhidden" part of the app.

The preferred solution in my opinion is to allow rituals to be hidden along with their underlying album. If not possible, I think it should be impossible to hide a album used for rituals to not give users the expectation that the photos are hidden when they are not.

### Version

1.3.0

### What product are you using?

Ente Photos

### What platform are you using?

Mobile - Android

laurenspriem:
Thanks for reporting!

This was fixed on release 1.3.10. Rituals tied to hidden albums now no longer show any thumbnails or route till the album till the user has performed authentication.

PR #8944, opened 2026-01-28
[mob][photos] Mobile fixes

laurens-pilot:
## Description

Multiple fixes for mobile

### DIFF
--- a/mobile/apps/photos/lib/ui/rituals/all_rituals_screen.dart
+++ b/mobile/apps/photos/lib/ui/rituals/all_rituals_screen.dart
@@ -1,3 +1,5 @@
+import "dart:async";
+
 import "package:ente_icons/ente_icons.dart";
 import "package:ente_pure_utils/ente_pure_utils.dart";
 import "package:flutter/material.dart";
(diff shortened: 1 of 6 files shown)


laurens-pilot:
Should fix #8906

laurenspriem:
> ### 💡 Codex Review
> https://github.com/ente-io/ente/blob/bdf3b28f9725fee398803b64fd0634058376cd34/mobile/apps/photos/lib/ui/rituals/all_rituals_screen.dart#L309-L311
> 
> 
> **P1 Badge Gate ritual camera access for hidden rituals**
> The new hidden-ritual auth gate only wraps the card tap; the camera tile still calls `openRitualCamera` directly. For a hidden ritual, today’s tile still shows the camera (via `shouldShowCamera`), so a user who hasn’t authenticated can open the ritual camera and interact with the hidden album from the overview. This bypasses the lock you just added; consider hiding the camera tile for hidden rituals or wrapping `openRitualCamera` in `requestHiddenRitualAccess` when the ritual is hidden.
> 
> ℹ️ About Codex in GitHub

This is fine. A user can take a new photo and add to a hidden album without authentication, but they can't use this route to see the album. So all the existing photos from the hidden album stay hidden till authentication happens.

PR #9102, opened 2026-02-10
[mob][photos] resolve caption before sync

ghost:
## Summary
- add `GET /files/public-magic-metadata?fileID=&collectionID=` to return the latest **file object** for that file+collection (including `pubMagicMetadata`) when caller has access
- on client 409 from `PUT /files/public-magic-metadata`, fetch latest data for only the conflicted file(s), refresh local state, and retry once
- keep optimistic local metadata/version updates only after successful write

## Why
Editing caption during/just-after backup can hit metadata version conflicts (409).
Previously the client triggered a full silent sync before retrying, which is expensive.
This changes recovery to a targeted fetch for the specific file+collection and retries once.

## Validation
- `dart format` on changed Dart files
- `flutter analyze lib/gateways/files/file_magic_gateway.dart lib/services/file_magic_service.dart`
- `gofmt` on changed Go files
- `go test ./pkg/api ./pkg/controller ./pkg/repo`

## Notes
- Full `go test ./...` fails in `pkg/repo/storagebonus` in local non-test environment with: `Not running tests in non-test environment` (pre-existing env guard).

### DIFF
--- a/mobile/apps/photos/lib/services/file_magic_service.dart
+++ b/mobile/apps/photos/lib/services/file_magic_service.dart
@@ -75,6 +75,15 @@ class FileMagicService {
         rethrow;
       }
 
+      if (files.length != 1) {
+        _logger.warning(
+          "public magic metadata conflict on bulk update, skipping per-file recovery",
+          e,
+          s,
+        );
+        rethrow;
+      }
+
       _logger.warning(
         "conflict while syncing public magic metadata, fetching latest metadata and retrying once",
         e,


mobile/apps/photos/lib/services/file_magic_service.dart
```diff
@@ -62,44 +62,93 @@ class FileMagicService {
...
+      if (e.response?.statusCode != 409) {
+        rethrow;
+      }
+
+      _logger.warning(
+        "conflict while syncing public magic metadata, syncing and retrying once",
+        e,
+        s,
+      );
+      await RemoteSyncService.instance.sync(silently: true);
```
ua741: I think this will wait for upload to finish. For now, either we can ignore this or build api to fetch latest medatada for files that are out of sync

server/pkg/api/file.go
```diff
@@ -417,6 +417,28 @@ func (h *FileHandler) UpdatePublicMagicMetadata(c *gin.Context) {
 	c.Status(http.StatusOK)
 }
 
+// GetPublicMagicMetadata returns latest public magic metadata for a single file
+// in a given collection if caller has access.
+func (h *FileHandler) GetPublicMagicMetadata(c *gin.Context) {
```
ua741: Let's change it to return the File object? We will be able to reuse this API for other cases as well in future if needed.

mobile/apps/photos/lib/services/file_magic_service.dart
```diff
@@ -62,44 +62,90 @@ class FileMagicService {
...
+      if (e.response?.statusCode != 409) {
+        rethrow;
+      }
+
+      _logger.warning(
+        "conflict while syncing public magic metadata, fetching latest metadata and retrying once",
+        e,
+        s,
+      );
+      await _refreshPublicMetadataFromServer(files);
```
ua741, line 92: Let's only try to patch if when only single file is getting updated? For bulk update, this can take way too long if we try to update for each image.  Another option is that we create new endpoint for update, and, the new endpoint will return list of fileIDs that are out of sync. If we are just patching for caption, restricting to single file would more sense,

PR #9153, opened 2026-02-12
[mob][photos] Enable locally available file download for single file

ashilkn:
## Description

Allow download from the file's popup menu in it's appbar. 

Simulator Screenshot - iPhone 17 Pro - 2026-02-12 at 09 51 22

## Tests
- Tested on Android and iOS

### DIFF
--- a/mobile/apps/photos/lib/l10n/intl_en.arb
+++ b/mobile/apps/photos/lib/l10n/intl_en.arb
@@ -1004,6 +1004,7 @@
       }
     }
   },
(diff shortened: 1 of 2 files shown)


ua741:
For android, should we also subtly or in some other way display the full path?

ua741 (approved):

ashilkn:
> For android, should we also subtly or in some other way display the full path?

That made the dialog very noisy, so decided not to show it.

PR #9169, opened 2026-02-12
[locker][android] enable 16KB page-size native packaging without Flutter upgrade

AmanRajSinghMourya:
Summary:
- pin Locker Android NDK to 28.2.13676358
- restrict Locker ABIs to armeabi-v7a, arm64-v8a, x86_64
- disable legacy JNI packaging and enable uncompressed native libs in bundle
- keep Flutter/Dart constraints unchanged

Verification:
- flutter pub get succeeds on current toolchain
- flutter build apk --debug --flavor independent succeeds
- zipalign 16KB check passed with Verification successful for libflutter.so and libsodium.so

Note: local release packaging requires signing storeFile.

### DIFF
--- a/mobile/apps/locker/pubspec.yaml
+++ b/mobile/apps/locker/pubspec.yaml
@@ -1,7 +1,7 @@
 name: locker
 description: "Ente Locker – Safe space for your important documents"
 publish_to: "none"
-version: 1.0.0+90
+version: 1.0.0+91
 environment:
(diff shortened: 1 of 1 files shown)


AmanRajSinghMourya:
> Not upgrading sodium_libs?

The build succeeded without the updating sodium_libs so I think the update of lib sodium is not needed

PR #9178, opened 2026-02-13
Revert "[web][Photos]: SidebarDrawer can be now closed by swiping"

anandbaburajan:
Reverts ente-io/ente#9045


anandbaburajan:
Some problems with this so reverting; will get to this when we work on onboarding and feel like the sidebar experience needs to be improved on mobile

PR #9187, opened 2026-02-13
[server] Schedule deletion for inactive accounts

ua741:


ua741:
@coderabbitai Act as a group of product manager and engineer and identify if there are big P1 issues or miss in our implementation and reminders around account deletion. Think of all edge cases (ignore the case there token table doesn't contain any token, we will handle that seprately)

PR #9189, opened 2026-02-14
feat: keep optimized copy when free up storage

notAunty:
## Description

This PR adds a feature that supplements the "Free up device storage". 

The feature keeps an optimized "low resolution image" on the mobile device and load in from the cloud when the user zoom into to the low resolution image while previewing it. There is also a setting to immediately load full resolution when the image is tapped on from the home page called "Eager load". 

This "Keep a low resolution copy" feature have a on/off switch in the "Free up device storage" page. 

The low resolution copy uses a high efficiency format, and show clear details in a small footprint (~100KB). The low resolution copy shall replace the original file in the System Gallery (Photos on iOS, File system on Android).

## Tests

- `mobile/apps/photos/test/services/ignored_files_service_optimized_copy_test.dart`
- `mobile/apps/photos/test/utils/local_settings_optimized_copy_test.dart`

## PS

This PR is heavily AI-assisted. I have some Flutter knowledge from the pre-v1 era

### DIFF
--- a/mobile/apps/photos/test/services/ignored_files_service_optimized_copy_test.dart
+++ b/mobile/apps/photos/test/services/ignored_files_service_optimized_copy_test.dart
@@ -1,0 +1,46 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:photos/models/file/file.dart';
+import 'package:photos/models/ignored_file.dart';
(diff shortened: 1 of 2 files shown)


ua741:
Request you to please use discussion thread for new feature requests.

PR #9325, opened 2026-02-23
[mob][photos] Improve feed and local notifications

ashilkn:
## Description

- Separate feed item and notification for newly shared albums
- Local notification for memories added to shared albums

## Tests

Tested, all good.


ua741:
@ashilkn We can close this?

PR #9333, opened 2026-02-24
[mob][photos] Fix offline mode CTA

ashilkn:


mobile/apps/photos/lib/ui/settings/developer_settings_page.dart
```diff
@@ -9,6 +10,26 @@ import "package:photos/service_locator.dart";
...
 import "package:photos/utils/dialog_util.dart";
+import "package:photos/utils/local_settings.dart";
+
+Future<AppMode?> _maybeChangeAppMode(
+  String input, {
+  required LocalSettings settings,
+}) async {
+  if (input == "offline") {
+    await settings.setShowOfflineModeOption(true);
+    await settings.setAppMode(AppMode.offline);
```
ua741: As discussed offline, we should be setting appMode. Just configure setShowOfflineModeOption

PR #9369, opened 2026-02-26
[mob][photos] bg ml

ghost:
## Description

Making ML run in background.

### DIFF
--- a/mobile/apps/photos/lib/services/machine_learning/ml_service.dart
+++ b/mobile/apps/photos/lib/services/machine_learning/ml_service.dart
@@ -135,6 +135,9 @@ class MLService {
   }
 
   Future<void> _maybePredownloadLocalModels() async {
+    if (isProcessBg) {
+      return;
+    }
     if (!hasGrantedMLConsent) {
       return;
     }
@@ -174,10 +177,6 @@ class MLService {
   }
 
   Future<void> runAllML({bool force = false}) async {
-    if (_isRunningML) {
-      _logger.info("runAllML called while already running, skipping");
-      return;
-    }
     try {
       final MLMode mode = isOfflineMode ? MLMode.offline : MLMode.online;
       final mlDataDB = _dbForMode(mode);


mobile/apps/photos/lib/main.dart
```diff
@@ -216,9 +221,14 @@ Future<void> _runMinimally(String taskId, TimeLogger tlog) async {
     _logger.info("[BG TASK] home widget sync");
     await _homeWidgetSync(true);
 
-    // await MLService.instance.init();
-    // await PersonService.init(entityService, MLDataDB.instance, prefs);
-    // await MLService.instance.runAllML(force: true);
+    if (flagService.enableMLInBackground) {
+      final isDeviceHealthy = await computeController.isDeviceHealthyFuture();
+      if (isDeviceHealthy) {
+        await MLService.instance.init();
```
ua741: Do we need to add check for device healthy here? Ideally mL service should listen and take care of this

mobile/apps/photos/lib/services/machine_learning/compute_controller.dart
```diff
@@ -68,6 +69,9 @@ class ComputeController {
 
   // Directly assign the values + Attach listener for compute controller
   Future<void> init() async {
+    // we don't need listeners to be initialized in background
+    if (isProcessBg) return;
```
ua741: Should we not listen to thermal state and other changes here?

ua741 (commented):
Cc @laurenspriem

mobile/apps/photos/lib/services/machine_learning/compute_controller.dart
laurenspriem: We should I think

ghost: Fixed. Now it will listen and react to those. Also it will acquire / release compute properly.

PR #9388, opened 2026-02-27
[mob][locker] Prevent duplicate default collections during signup

AmanRajSinghMourya:
## Summary
- prevent concurrent default-collection setup calls in Locker by introducing a single in-flight gate inside CollectionService
- route both startup or signed-in setup and HomePage fallback through ensureDefaultCollections()
- keep readiness check centralized in ensureDefaultCollections() using hasConfiguredAccount()

## Root Cause
During first-time signup, two async paths could call default setup concurrently (service init path plus Home fallback), both observing missing defaults and creating duplicate collections (including duplicate Documents).

### DIFF
--- a/mobile/apps/locker/lib/services/collections/collections_service.dart
+++ b/mobile/apps/locker/lib/services/collections/collections_service.dart
@@ -774,6 +774,13 @@ class CollectionService {
       return;
     }
 
+    if (!hasCompletedFirstSync()) {
+      _logger.info(
+        "Skipping default collections setup - first sync not completed yet",
+      );
+      return;
+    }
+
     final inFlight = _defaultSetupInFlight;
     if (inFlight != null) {
       await inFlight;


mobile/apps/locker/lib/services/collections/collections_service.dart
```diff
@@ -771,6 +766,27 @@ class CollectionService {
...
+    if (inFlight != null) {
+      await inFlight;
+      return;
+    }
+
+    final setupFuture = Future<void>.sync(setupDefaultCollections);
+    _defaultSetupInFlight = setupFuture.whenComplete(() {
+      _defaultSetupInFlight = null;
+    });
+    await _defaultSetupInFlight;
```
ua741, line 794: Even before queuing first request, we should ensure collections are fetched, and such collection doesn't existing locally.

AmanRajSinghMourya: fixed

PR #9404, opened 2026-02-27
auth: remove slow unlock transition on iOS

lnilluv:
## Summary
- use zero-duration app-lock route transitions for lock/unlock paths
- avoid blocking unlock on invalid-attempt counter reset by performing it asynchronously
- keep lock-screen behavior unchanged while reducing post-auth latency

## Test Plan
- flutter analyze lib/ui/app_lock.dart lib/ui/lock_screen.dart
- flutter test test/app_lock_route_test.dart

### DIFF
--- a/mobile/packages/lock_screen/lib/ui/lock_screen.dart
+++ b/mobile/packages/lock_screen/lib/ui/lock_screen.dart
@@ -1,4 +1,3 @@
-import 'dart:async';
 import 'dart:io';
 import 'dart:math';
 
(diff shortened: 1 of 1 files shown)


AmanRajSinghMourya:
@codex review

AmanRajSinghMourya:
@lnilluv Could you tell us a reason why to not keep it delayed. As considering the fact that the delay is just 300ms and I hope that delay provides a smoother transition to the app then the current implementation which can feel snappy

lnilluv:
> @lnilluv Could you tell us a reason why to not keep it delayed. As considering the fact that the delay is just 300ms and I hope that delay provides a smoother transition to the app then the current implementation which can feel snappy

I tested several variants today (route timing, first paint, keyboard timing, privacy blur, auth handoff), and none gave a real improvement. At this point the remaining delay looks like a native iOS handoff issue and is above my skill level to fix in this PR. I’m closing this instead of adding AI slop / guessy changes. Sorry for the time wasted.

PR #9405, opened 2026-02-27
[auth] keep OTP code readable with larger text scales

lnilluv:
## Summary

This fixes the iOS accessibility text scaling issue in the Auth code card.

At the largest iOS text sizes, the OTP card could become too tall and cramped, making the title/account name and codes hard to use. The fix keeps the existing layout, but caps the card text scale on iOS at `2.0x`, matching the effective ceiling Android already has.

The latest maintainer feedback is addressed by keeping the change narrow:
- Android behavior is unchanged.
- Smaller iOS text sizes are preserved.
- The next-code/HOTP action is not hidden when space is tight.

## Changes

- Add an iOS-only `MediaQuery` override around the code card.
- Cap only text scales above `2.0`; smaller user-selected scales still pass through.
- Add unit coverage for the text-scale cap behavior.

## Test plan

- `flutter analyze lib/ui/code_widget.dart lib/ui/code_widget_layout_utils.dart test/ui/code_widget_layout_test.dart`
- `flutter test test/ui/code_widget_layout_test.dart`

<details>
<summary>Screenshots</summary>

<details>
<summary>Android accessibility</summary>
Android - Max Accessibility
</details>

<details>
<summary>iOS clipping issue at max accessibility settings</summary>
iOS clipping issue
</details>

<details>
<summary>iOS after fix at max accessibility settings</summary>
iOS after fix
</details>

</details>

### DIFF
--- a/mobile/apps/auth/lib/ui/code_widget.dart
+++ b/mobile/apps/auth/lib/ui/code_widget.dart
@@ -15,6 +15,7 @@ import 'package:ente_auth/store/code_display_store.dart';
 import 'package:ente_auth/store/code_store.dart';
 import 'package:ente_auth/theme/ente_theme.dart';
 import 'package:ente_auth/ui/code_timer_progress.dart';
+import 'package:ente_auth/ui/code_widget_layout_utils.dart';
 import 'package:ente_auth/ui/components/auth_qr_dialog.dart';
 import 'package:ente_auth/ui/components/note_dialog.dart';
 import 'package:ente_auth/ui/home/shortcuts.dart';
@@ -337,7 +338,8 @@ class _CodeWidgetState extends State<CodeWidget> {
       );
     }
 
-    return Container(
+    final bool isIOS = !kIsWeb && Platform.isIOS;
+    final Widget content = Container(
       margin: widget.isCompactMode
           ? const EdgeInsets.only(left: 16, right: 16, bottom: 6, top: 6)
           : const EdgeInsets.only(left: 16, right: 16, bottom: 8, top: 8),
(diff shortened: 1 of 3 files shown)


ua741:
@lnilluv Thank you for the change. Can you please add before and after screenshots, that would help us understand this PR better.

prateekmedia:
Not sure if we should just hide the next code in that scenario

lnilluv:
> Not sure if we should just hide the next code in that scenario

I agree I shouldn’t hide the next code if we can avoid it.
I’m seeing a potential Android regression with my current approach, so I’d like to scope this fix to iOS only (issue 2005 is iOS-specific) and keep Android behavior unchanged, I belive it fits better since iOS & Android apps are quite different.
I’ll share before/after iOS screenshots first. If the UI looks OK on your side, I’ll finalize and push the commit.

lnilluv:
Updated the PR with screenshots. 

I forced push to keep the history clean. The fix now properly targets only iOS. 
Only the OTP text had a cap and the card title was still rendering at 3x. The cap is now at the card level to scale everything together and both totp should be visible.

ua741:
@codex review

ua741:
@codex review

ua741:
This seems to change behavior beyond the iOS fix. `shouldShowNextTotpCode()` can hide the visible next-code UI on narrower Android cards at larger text scales, and `clamp(1.0, 2.0)` also forces any iOS text scale below `1.0` back to default size. I think the safe fix is to keep the iOS upper cap, preserve smaller user-selected scales, and avoid hiding the next-code action when space is tight.

lnilluv:
Updated the PR based on your feedback.

The fix is narrower now:
- Android behavior is unchanged.
- The next-code/HOTP action is no longer hidden when space is tight.
- Smaller iOS text scales are preserved. Only values above `2.0x` are capped.

I also updated the tests for the text-scale cap.

ua741:
@codex review

ua741:
@codex review

ua741 (approved):

PR #9459, opened 2026-03-03
[mob][photos] Add separate tiff decoder as fallback for image crate tiff decoding

laurens-pilot:
## Description

Add separate tiff decoder as fallback for image crate tiff decoding


laurenspriem:
@codex review

PR #9464, opened 2026-03-03
[mob][photos] Show newly shared collections in feed + notifications

ashilkn:
## Description

- Show feed item and notification for newly shared album
- Show notification for new memories added to a shared album


ashilkn:
@codex Review

PR #9501, opened 2026-03-04
[auth] Add search-by-intent via ente-auth://search?query= deep link

bekoeppel:
## Description

- Support launching Ente Auth with a pre-filled search query via deep link, e.g. `ente-auth://search?query=gmail`
- Useful for automation tools (Tasker, shortcuts) to jump directly to a specific OTP code
- Fix deep link stream listener not being retained, so deep links now work when the app is already open

## Tests

- Build and install the app
- `adb shell am start -a android.intent.action.VIEW -d "enteauth://search?query=gmail"` with app closed: should open with search filtered
- Same command with app already open: should update the search
- `adb shell am start -a android.intent.action.VIEW -d "enteauth://abcd?query=gmail"`: should be ignored (only `/search` path is accepted)
- Verify existing `otpauth://` deep links still work

### DIFF
--- a/mobile/apps/auth/lib/ui/home_page.dart
+++ b/mobile/apps/auth/lib/ui/home_page.dart
@@ -1914,14 +1914,18 @@ class _HomePageState extends State<HomePage> {
     lastScanTime = DateTime.now().millisecondsSinceEpoch;
     final lowerLink = link.toLowerCase();
     if (mounted && (lowerLink.startsWith("ente-auth://") || lowerLink.startsWith("enteauth://"))) {
-      final uri = Uri.parse(link);
-      if (uri.host != "search") return;
-      final searchQuery = uri.queryParameters['query'];
-      if (searchQuery != null && searchQuery.isNotEmpty) {
-        _showSearchBox = true;
-        _textController.text = searchQuery;
-        _searchText = searchQuery;
-        _applyFilteringAndRefresh();
+      try {
+        final uri = Uri.parse(link);
+        if (uri.host != "search") return;
+        final searchQuery = uri.queryParameters['query'];
+        if (searchQuery != null && searchQuery.isNotEmpty) {
+          _showSearchBox = true;
+          _textController.text = searchQuery;
+          _searchText = searchQuery;
+          _applyFilteringAndRefresh();
+        }
+      } catch (e) {
+        _logger.warning("Malformed ente-auth deep link: $link", e);
       }
     } else if (mounted && lowerLink.startsWith("otpauth://")) {
       try {
--- a/mobile/apps/auth/lib/ui/home_page.dart
+++ b/mobile/apps/auth/lib/ui/home_page.dart
@@ -1907,11 +1907,6 @@ class _HomePageState extends State<HomePage> {
     if (!(isAccountConfigured || isOfflineModeEnabled) || link == null) {
       return;
     }
-    if (DateTime.now().millisecondsSinceEpoch - lastScanTime < 1000) {
-      _logger.info("Ignoring potential event for same deeplink");
-      return;
-    }
-    lastScanTime = DateTime.now().millisecondsSinceEpoch;
     final lowerLink = link.toLowerCase();
     if (mounted && (lowerLink.startsWith("ente-auth://") || lowerLink.startsWith("enteauth://"))) {
       try {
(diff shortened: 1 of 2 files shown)


bekoeppel:
@codex review

bekoeppel:
@codex review

AmanRajSinghMourya:
@bekoeppel Thanks for the contribution but could you also add a video recording or a screenshot too.

bekoeppel:
Hi @AmanRajSinghMourya! Sure, happy to. Here's a quick video how I intend to use this new intent (pun intended).

https://github.com/user-attachments/assets/b8739bdf-4cf0-45fc-b5d5-7b90511b272e

bekoeppel:
Hi @AmanRajSinghMourya, hope you're doing well! Were you able to take a look at the video I uploaded? Do you have any feedback for my pull request? Thank you! Benedikt

mobile/apps/auth/lib/ui/home_page.dart
```diff
@@ -1908,12 +1907,27 @@ class _HomePageState extends State<HomePage> {
...
+          _showSearchBox = true;
+          _textController.text = searchQuery;
+          _searchText = searchQuery;
+          _applyFilteringAndRefresh();
+        }
+      } catch (e) {
+        _logger.warning("Malformed ente-auth deep link: $link", e);
+      }
+    } else if (mounted && lowerLink.startsWith("otpauth://")) {
+      if (DateTime.now().millisecondsSinceEpoch - lastScanTime < 1000) {
```
prateekmedia, line 1925: It would be better to extract this above the if-else ladder, so it applies to both enteauth and otpauth

mobile/apps/auth/lib/ui/home_page.dart
```diff
@@ -1908,12 +1907,27 @@ class _HomePageState extends State<HomePage> {
...
       return;
     }
-    if (DateTime.now().millisecondsSinceEpoch - lastScanTime < 1000) {
-      _logger.info("Ignoring potential event for same deeplink");
-      return;
-    }
-    lastScanTime = DateTime.now().millisecondsSinceEpoch;
-    if (mounted && link.toLowerCase().startsWith("otpauth://")) {
+    final lowerLink = link.toLowerCase();
+    if (mounted && (lowerLink.startsWith("ente-auth://") || lowerLink.startsWith("enteauth://"))) {
```
prateekmedia: I am not sure about ente-auth:// pattern, any reason you included it here?

mobile/apps/auth/lib/ui/home_page.dart
bekoeppel: Thanks for your feedback! 

I skipped the throttling only for `ente-auth://search` links on purpose, because it's OK to repeat a search query (it's idempotent, as it just filters the list). 

However, the throttle already existed for OTP code additions (with `otpauth://` links), I assume this is because `otputh://` links add elements to the list, so we wanted to debounce that logic.

If I moved the throttle above the if-else, it adds an unnecessary delay to search deep links.

mobile/apps/auth/lib/ui/home_page.dart
bekoeppel: I saw `ente-auth://` pattern (with a `-`) is registered in [AndroidManifest.xml](https://github.com/ente-io/ente/blob/164efcf74ae05de37d24ba98b66f7cb718455d02/mobile/apps/auth/android/app/src/main/AndroidManifest.xml): https://github.com/ente-io/ente/blob/164efcf74ae05de37d24ba98b66f7cb718455d02/mobile/apps/auth/android/app/src/main/AndroidManifest.xml#L19-L24

along with the `enteauth://` pattern (without a `-`): https://github.com/ente-io/ente/blob/164efcf74ae05de37d24ba98b66f7cb718455d02/mobile/apps/auth/android/app/src/main/AndroidManifest.xml#L33-L38

Let me know if I should remove the one with the `-`.

ua741:
Can we keep this to `enteauth://search` only and drop `ente-auth://` here? `ente-auth` is only registered on Android today, so supporting it in new code makes the behavior inconsistent across platforms, and if we’re not relying on it anymore we can remove that scheme from the Android manifest too. While touching this, could we also narrow the handler to the exact search route instead of early-returning for any non-search `enteauth://...` link, so future deep links don’t get accidentally swallowed here?

bekoeppel:
Thanks for your feedback. I removed the `ente-auth://` from both the handler and the Android manifest, and narrowed the match to `enteauth://search` directly so other `enteauth://` routes won't be swallowed.

ua741 (approved):

Issue #9532, opened 2026-03-05
Publish `App Signing Certificate's SHA-256 Hash` for all the (4) apps upfront.


hrsiit:
Hi, I was recently trying to install ente's app [ente auth, ente locker, ente photos, ensu] on my android device without google play store, directly through github using obtainium. 
But was stuck / blocked in doing so, because I couldn't find the **App Signing Certificate's SHA-256** for any of the app to verify the app's integrity using [AppVerifier](https://github.com/soupslurpr/AppVerifier) after downloading it.

Ultimately via [ `https://github.com/ente-io/ente/issues/4219` ] I got to know that atleast the hash is published somewhere, even though the link provided there was broken.

Soon enough with little bit of digging, 
I finally found the hash for ente auth at [ `https://github.com/ente-io/ente/blob/main/mobile/apps/auth/README.md` ]
and ente photos at [ `https://github.com/ente-io/ente/blob/main/mobile/apps/photos/README.md` ].

Unfortunately, I still couldn't find the hash for ente locker and ensu (likely because they are new app), 
I tried looking up their readme at [ `https://github.com/ente-io/ente/blob/main/mobile/apps/locker/README.md` ] 
and [ `https://github.com/ente-io/ente/blob/main/mobile/native/android/apps/ensu/README.md` ] respectively.

I would thereby request ente team to please **publish / provide** `App Signing Certificate's SHA-256 Hash` for all their current and future apps timely.
Preferably upfront at [ `https://github.com/ente-io/ente/blob/main/README.md` ]  
and [ `https://ente.io/help` ], so that new users don't have to go through so many hops to find them in future.

AmanRajSinghMourya:
@hrsiit Yes will add the fingerprints for locker and auth on the readme

hrsiit:
Thankyou.
If possible, please try to publish all of these certificate hash upfront on the **`ente-io/ente/README.md`** file instead of hidden inside the repo files.

hrsiit:
Would really appreciate if the hash can also be published on **`ente.io/help`** . 
Security experts say that ideally certificate hash should be published somewhere different than where the apk is hosted.
This is done so to prevent undetected tampering, in the case an attacker gets access to the hosted repo, it can not just alter the hash too. Plus posting the SHA-256 hash of the developer's app signing key in multiple locations would make changes detectable easily.

pingu-the-penguin:
[...]

PR #9550, opened 2026-03-06
[locker] Add certificate fingerprints to README

AmanRajSinghMourya:
## Summary
- add Locker certificate fingerprints to the app README
- include the same apksigner verification command used in Auth and Photos

## Notes
- fingerprints were extracted from the official locker-v1.0.0 GitHub release APK using apksigner verify --print-certs
- partially addresses #9532

### DIFF
--- a/mobile/apps/locker/README.md
+++ b/mobile/apps/locker/README.md
@@ -29,8 +29,12 @@ flutter run -t lib/main.dart --flavor independent
 
 For iOS, use `flutter run`
 
-## TODOs
+## Certificate Fingerprints
 
-Refactor and merge
-- [ ] Verify correctness for `PackageInfoUtil.getPackageName()` on Linux and Windows
-- [ ] Update `file_url.dart` to download only via CF worker when necessary
+- **SHA1**: EF:BA:46:7F:BE:E2:F7:9A:0C:A7:76:A2:9D:AA:70:13:B9:B7:AE:D2
+- **SHA256**: 6E:5B:71:61:B0:FA:F1:01:B6:AF:3D:33:C6:B0:8C:AD:AC:4A:8B:DF:85:E5:BE:A5:06:83:AA:FA:74:05:0D:B1
+
+To verify these fingerprints, use the following command:
+```bash
+apksigner verify --print-certs <path_to_apk>
+```


hrsiit:
Hi @AmanRajSinghMourya, I see you are a contributor of this project.
Does that mean you also have the signing certificate secret private key used to sign the legitimate app?

---

If that's not the case and you ran `apksigner verify --print-certs <path_to_apk>` to get the hash of the apk to publish here, it's not **"exactly"** the same thing as publishing the _"App's Signing Certificate's SHA-256 Hash"_ since that requires access to the private key, what you got from that command is the hash of the public certificate embedded directly inside the APK. 

No offence, And I don't mean to demean you or your efforts in anyway, but anyone can technically extract that hash and post it on GitHub, but only the developer can provide the authoritative hash. One can use this method i.e. the `apksigner verify --print-certs <path_to_apk>` command to verify the app then.

---

This is because of the root of trust.

For example, hypothetically speaking, If an attacker has access to this repo and and replaced the legitimate APK with a malware-infected APK signed with the it's own certificate. This would mean when you downloaded the apk and ran the apksigner command, your apksigner command would have outputted the hacker's SHA-256 hash, not the developer's true private signing key hash. 

The ente team possesing the key (or you too, if you have the private key) are the only one who can publish the hash so users have a safe baseline to compare against.

AmanRajSinghMourya:
@hrsiit In this case, the Locker fingerprint was verified from our official release signing setup. The GitHub/independent APKs are built by our release workflow using our signing key, and the published fingerprint matches that signing certificate. So the value in the README is intended as the trusted baseline for users installing via GitHub/Obtainium.

As far as you are concerned about me as a contributor you can find me [here](https://ente.io/about/)

hrsiit:
Perfect! @AmanRajSinghMourya 
Thank you so much for the clarification.
And so sorry lol, I was honestly just worried / confused about you being a random-friendly-helpful contributor and not from the ente team directly.
Since your profile doesn't mention anything about you working in / for ente.io unlike other people in the team, 
for example - `Amrithesh-Kakkoth`, `prateekmedia`, `mnvr`, `ua741`, `laurenspriem`. Would highly recommend adding something like that in your profile too for the future.

ashilkn (approved):

PR #9591, opened 2026-03-10
[server] Implement storage warning lifecycle with terminal access lockout

neeraj-pilot:
## Summary
- add the staged storage warning workflow for expired subscriptions and active overage accounts, including mail templates, cadence enforcement, rollout metrics, and scheduled-deletion handling
- add the daily storage warning cron, the notification-history grouping migration, and the repo/controller plumbing needed to evaluate individual accounts and family recipients
- block Photos and Locker logins after the scheduled-deletion stage and add an admin endpoint to clear that login block when support needs to recover an account
- normalize expired warning stage IDs to the anchor-relative `0d/30d/60d/90d/119d` sequence and document that family candidate selection intentionally stays per-member

## Operational Notes
- the mailer runs from `museum` on a 24 hour cadence via `SendStorageWarningMails`
- rollout is still gated by `storageWarningRolloutPercentage = 0`; `@ente.io` accounts bypass the gate for internal validation
- the scheduled-deletion stage resets Photos and Locker access and records the lockout state in `notification_history`, but it does not enqueue data cleanup yet
- active-overage history is grouped so prior-cycle reminders can be cleaned up when an account leaves the overage bucket
- family candidate selection is intentionally based on an individual member crossing the threshold, even though the downstream family snapshot evaluates aggregate family usage
- a new admin API route, `POST /admin/user/unblock-storage-warning-login`, clears the scheduled-deletion login block

## Tests
- `go test ./pkg/repo ./pkg/controller/email ./pkg/controller/user`

### DIFF
--- a/server/pkg/controller/email/storage_warning.go
+++ b/server/pkg/controller/email/storage_warning.go
@@ -172,6 +172,10 @@ func (c *EmailNotificationController) SendStorageWarningMails() {
 		log.Error("Skipping storage warning mails: controller dependencies are not fully configured")
 		return
 	}
+	if c.UserRepo.IsLikelySelfHosted() {
+		return
+	}
+	log.Info("Running storage warning mails for cloud instance")
 
 	lockStatus := c.LockController.TryLock(StorageWarningMailLock, time.MicrosecondsAfterHours(24))
 	if !lockStatus {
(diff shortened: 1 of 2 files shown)


ua741:
@codex Review

ua741:
@codex review

PR #9594, opened 2026-03-10
[mob][photos] don't prepare upload on bg when metered

prateekmedia:
## Description

- Only runs bg sync on mobile data when "Backup over mobile data" is enabled
- This is done through the "unmetered" and "connected" setting of WorkManager
- Also re-schedule bg task on the setting change

### DIFF
--- a/mobile/apps/photos/scripts/internal_changes.txt
+++ b/mobile/apps/photos/scripts/internal_changes.txt
@@ -19,5 +19,5 @@
 - Laurens: Use vectorDB for more accurate suggestions in the people overview
 - Laurens: (I) Use cloudflare worker for upload for internal users
 - Neeraj: (I) Add internal-only resumable gallery download queue with persistence/restart fixes and completion banner updates
-- Prateek: Background sync now short-circuits upload prep on mobile data when "Backup over mobile data" is disabled
 - Prateek: (i) Enable BG sync notifications by default for internal users on iOS
+- Prateek: Background sync now short-circuits upload prep on mobile data when "Backup over mobile data" is disabled


mobile/apps/photos/lib/utils/bg_task_utils.dart
```diff
@@ -77,10 +78,13 @@ class BgTaskUtils {
     const androidPeriodicTask = "io.ente.photos.androidPeriodicTask";
     final backgroundTaskIdentifier =
         Platform.isIOS ? iOSBackgroundAppRefresh : androidPeriodicTask;
+    final shouldBackupOverMobileData =
```
ua741: Let's only make this change for iOS? On Android, sync anyways run frequently, and we will be still able to show sharing related notifications. Also, let's disable this check for for offline mode too.

prateekmedia: Done! Only checking wifi-only for iOS && online mode.

prateekmedia:
@ua741 Have updated the logic to short-circuit upload prep in background on mobile data and Backup over mobile data is off.

mobile/apps/photos/lib/services/sync/remote_sync_service.dart
```diff
@@ -151,6 +157,10 @@ class RemoteSyncService {
         fileDataService.syncFDStatus().ignore();
       }
 
+      if (!canPrepareUploadsInBackground) {
+        throw WiFiUnavailableError();
```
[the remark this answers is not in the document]
prateekmedia: Expected change! That's what it is about

PR #9653, opened 2026-03-13
[mob][photos] Ensure syncAll triggers remote sync when mappings change

prateekmedia:
Summary
- make `syncAll` report success when existing mappings change so the follow-up remote sync still runs
- document that repairing mappings without new files still needs the remote sync path to promote entries

Testing
- Not run (not requested)


ua741:
Just a note: I don't think this should fix the issue where new files were not getting discovered as mapping won't change for those.

PR #9682, opened 2026-03-16
[mob][photos] Add minimal sync recovery diagnostics

prateekmedia:
## Description

Diagnostics for sync recovery


ua741 (commented):
Let's add a separate flag for this so that it's easy to clean them up later?

PR #9701, opened 2026-03-17
[server] Add device, app, and IP to login notification email

sv3nnie:
**TLDR: Include app name (Photos, Auth, Locker), device/browser info, and IP address in login notification emails**

The current login notification email only includes the timestamp, making it difficult for users to determine whether a login was legitimate. Since Ente has various different products and users may use multiple products with the same account, including the app name also helps distinguish which service was logged into and makes it easier to identify unauthorized access.

I saw some requests about this in GitHub issues/discussions and missed this myself as well. Feel free to close this PR if it doesn't align with your vision or priorities. It required minimal effort, most of the data (IP, user-agent, app) was already available in `AddTokenAndNotify` since it's used by the session manager. It just wasn't being passed to the email template.

## Changes

- **`server/mail-templates/on_login.html`** — Added a table displaying the app, device, and IP address
- **`server/pkg/controller/user/userauth.go`** — Resolved app to a display name, parsed user-agent into a readable string via `GetPrettyUA`, and passed all three new fields to the email template

CleanShot 2026-03-17 at 18 59 33

### DIFF
--- a/server/mail-templates/on_login.html
+++ b/server/mail-templates/on_login.html
@@ -103,6 +103,7 @@
       <p>
       We detected a new login to your Ente account at {{ .Date }} UTC.
       </p>
+      {{ if .App }}
       <table>
         <tr>
           <td><strong>App</strong></td>
@@ -117,6 +118,7 @@
           <td>{{ .IP }}</td>
         </tr>
       </table>
+      {{ end }}
       <p>
       If this is not you, please reset your password. For any assistance or questions, please reply to this email.
       </p>
--- a/server/pkg/controller/user/userauth.go
+++ b/server/pkg/controller/user/userauth.go
@@ -447,29 +447,34 @@ func (c *UserController) AddTokenAndNotify(ctx context.Context, userID int64, ap
 		return nil
 	}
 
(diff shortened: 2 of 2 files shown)


vishnukvmd:
Hey @sv3nnie, could you please address the review comments and enable this for just internal users (`strings.HasSuffix(emailUtil.NormalizeEmail(user.Email), "@ente.io")`)? We'd like to test this out internally before rolling it out in production. Thanks!

sv3nnie:
> Hey @sv3nnie, could you please address the review comments and enable this for just internal users (`strings.HasSuffix(emailUtil.NormalizeEmail(user.Email), "@ente.io")`)? We'd like to test this out internally before rolling it out in production. Thanks!

Addressed the comments and made sure it'll only include the new fields for `@ente.io` addresses

ua741:
@sv3nnie could you please rebase this on `main` and push an updated version? Looks good otherwise.

sv3nnie:
> @sv3nnie could you please rebase this on `main` and push an updated version? Looks good otherwise.

Done, rebased on main and pushed.

ua741 (approved):

PR #9754, opened 2026-03-23
fix(ios): exclude sensitive app directories from iCloud backup

### LINKED ISSUE #9741: Ente Photos includes photos in backups, possibly exposing them unencrypted to Apple
### Description

I noticed that when Ente Photos is included in iCloud backup, it includes several hundred megabytes of data, which didn't seem right if it only included basic configuration data as I would expect. My main concern with this is that Apple can access backups if Advanced Data Protection is not enabled, raising privacy concerns and undermining Ente's promise that only the user (not any cloud provider) can access their photos. While I can't explore iCloud backups directly, I took a backup to my Mac and uncovered several worrying or unnecessary things included in the backup:

1. Face recognition thumbnails. This exposes the people stored in the user's library.
2. Cached data from the Share Extension. This included _**several entire photos**_ imported via the share extension, which should not be backed up to iCloud given that they could be sensitive.
3. A general database (ente.files.db), containing possibly sensitive metadata such as GPS coordinates for certain photos
4. ML models. This is not a privacy concern, but likely contributes to the hundreds of megabytes stored in iCloud, and is unnecessary to back up individually for each user.

From my understanding, there are [ways](https://developer.apple.com/documentation/foundation/optimizing-your-app-s-data-for-icloud-backup) to exclude certain data from backups, and I saw these in use in a few places throughout the codebase, but it does not appear to cover all cases where user photos and sensitive data could be backed up unencrypted via iCloud Backup.

If I'm incorrect and these files do happen to be properly excluded from iCloud Backup yet not from iTunes Backup, please feel free to disregard. However, I still feel that this is worth looking into as I did not like what I was able to extract from a backup, assuming that this data is also included in unencrypted iCloud backups.

### Version

1.3.24

Ente Photos

Mobile - iOS

lumm2509:
## Problem
  Ente Photos stores sensitive data in directories that are included in
  iCloud Backup by default. Without Advanced Data Protection enabled,
  Apple holds the decryption keys for iCloud Backup and can access this
  data in response to legal requests — directly contradicting Ente's E2EE
  privacy model.

  Directories affected:
  - `Documents/person-face-thumbnail-cache/` — decrypted face crop images
  - `Documents/ente-shared-media/` — full-resolution photos pending upload
  - `Documents/*.db` — SQLite databases with file metadata
  - `Library/Application Support/` — downloaded ML model assets

  Reported in #9741 and discussed in #9075.

  ## Fix
  Registers a dedicated platform channel (`io.ente.photos/backup`) in
  `AppDelegate.swift` that calls `setResourceValue(_:forKey: .isExcludedFromBackupKey)`
  on a given path. The exclusion is applied to `Documents/` and
  `Application Support/` on every app init — intentionally, so the flag
  is re-applied after device restores (which recreate directories without
  the filesystem attribute).

  No new dependencies. `pubspec.lock` is unchanged.

  ## Changes
  - `ios/Runner/AppDelegate.swift` — registers backup exclusion channel
  - `lib/utils/ios_backup_util.dart` — Dart wrapper with logging on failure
  - `lib/core/configuration.dart` — applies exclusion during app init
  - `test/utils/ios_backup_util_test.dart` — 4 unit tests

  ## Tests
  flutter test test/utils/ios_backup_util_test.dart
  +4: All tests passed!

  Covers: correct channel invocation, warning on `false` return, warning
  on `PlatformException`, no-op on non-iOS platforms.

  Closes #9741
  Ref: #9075

### DIFF
--- a/mobile/apps/photos/lib/core/configuration.dart
+++ b/mobile/apps/photos/lib/core/configuration.dart
@@ -41,6 +41,7 @@ import 'package:photos/services/search_service.dart';
 import 'package:photos/services/sync/sync_service.dart';
 import 'package:photos/services/video_preview_service.dart';
 import 'package:photos/utils/file_uploader.dart';
+import 'package:photos/utils/ios_backup_util.dart';
 import "package:photos/utils/lock_screen_settings.dart";
 import 'package:photos/utils/validator_util.dart';
 import 'package:shared_preferences/shared_preferences.dart';
@@ -95,6 +96,10 @@ class Configuration {
         ),
       );
       _documentsDirectory = (await getApplicationDocumentsDirectory()).path;
+      final appSupportDirectory = await getApplicationSupportDirectory();
+      // Exclude Documents (SQLite, thumbnails, decrypted media) and Application Support (ML models) from backups since they’re server-derivable or must remain within the device’s E2EE boundary.
+      await excludeFromBackup(_documentsDirectory);
+      await excludeFromBackup(appSupportDirectory.path);
       _tempDocumentsDirPath = _documentsDirectory + "/temp/";
       final tempDocumentsDir = Directory(_tempDocumentsDirPath);
       await _cleanUpStaleFiles(tempDocumentsDir);
--- a/mobile/apps/photos/lib/utils/ios_backup_util.dart
+++ b/mobile/apps/photos/lib/utils/ios_backup_util.dart
@@ -38,5 +38,10 @@ Future<void> _invokeExcludeFromBackup(String path) async {
     }
   } on PlatformException catch (e) {
     _logger.warning('Failed to exclude path from backup: $path — ${e.message}');
+  } on MissingPluginException {
+    // Channel not registered in headless background execution (Workmanager).
+    // Safe to ignore: backup exclusion is a best-effort attribute set on
+    // foreground init; background tasks do not trigger iCloud backups.
+    _logger.info('excludeFromBackup skipped: channel unavailable (background)');
   }
 }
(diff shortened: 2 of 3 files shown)


gitgizmo7:
@ua741 or @lumm2509 can you confirm that this also covers the share extension? Also, for the share extension, I'm not sure why it's actually keeping the photos in the first place. I have several photos in there from like a month ago that were already successfully imported into Ente and seem like they should have been deleted. Is there a process for this or does this also need to be added?

lumm2509:
@gitgizmo7  

Hi! I don’t think this fully covers the share extension yet
* This PR excludes the main app’s `Documents/` and `Application Support/` from backup during init.
* It also handles `MissingPluginException`, so background runs won’t crash if the channel isn’t available.
* It does not implement any cleanup for leftover files.

For the share extension:
* I haven’t verified how it stores files or whether backup exclusion is applied there.

Suggestion:
Might be worth testing the extension directly to confirm whether those files are included or excluded.

Happy to be proven wrong, just don’t want to invent architecture that isn’t in the diff

gitgizmo7:
@ua741 Until this is confirmed to be fully fixed, I believe my issue should be re-opened.

ua741:
> I have several photos in there from like a month ago that were already successfully imported into Ente and seem like they should have been deleted. Is there a process for this or does this also need to be added?

Can you please confirm the directory path for these files in your iCloud backup?

gitgizmo7:
It's something like App Groups/AppDomainGroup-group.io.ente.frame.ShareExtension. I couldn't tell you whether iMazing is somehow abstracting/simplifying the iOS file system but that's what I'm able to see.

PR #9755, opened 2026-03-23
[mob][photos] add grace window plugin for ios

prateekmedia:
## Description

Allow operations to run for a short interval in background using grace window.


prateekmedia:
https://github.com/ente-io/ente/pull/9810

PR #9766, opened 2026-03-24
auth: select all text on double-tap in search field

mariow:
## Summary
Wraps the search TextField in a GestureDetector that selects all text when the user double-taps the search input field. This makes it easy to replace the current search query without manually selecting text.

## Changes
- `mobile/apps/auth/lib/ui/home_page.dart`: Wrapped the search `TextField` in a `GestureDetector` with an `onDoubleTap` handler that sets the `TextEditingController` selection from offset 0 to the end of the text.

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/mobile/apps/auth/lib/ui/home_page.dart
+++ b/mobile/apps/auth/lib/ui/home_page.dart
@@ -1484,27 +1484,35 @@ class _HomePageState extends State<HomePage> {
       ),
       title: !_showSearchBox
           ? const AuthLogoWidget(height: 18)
-          : TextField(
-              autocorrect: false,
-              enableSuggestions: false,
-              autofocus: _autoFocusSearch,
-              controller: _textController,
-              onChanged: (val) {
-                _searchText = val;
-                _applyFilteringAndRefresh();
+          : GestureDetector(
+              onDoubleTap: () {
+                _textController.selection = TextSelection(
+                  baseOffset: 0,
+                  extentOffset: _textController.text.length,
+                );
               },
-              onSubmitted: (_) {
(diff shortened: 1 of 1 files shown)


ua741:
@mariow Double tapping on text selects all text on android and ios. Isn't not happening for you on any particular platform? 

Also, as per flutter docs, textfield already register double tap gestures. They don't seem to encourage wrapping textfield with GestureDectector for double tap.

mariow:
@ua741 You are right, tapping and clicking on the text works fine, my case is a bit different: I typically have a very short search strings in the search field on desktop as Ente Auth runs all day, something like "aws". The search field is way longer than that text and I tend to double-click somewhere in the search field rather than directly on the text. The change makes sure that the text still gets selected, so I can double-click anywhere in the field and start typing a new search term.

ua741:
> @ua741 You are right, tapping and clicking on the text works fine, my case is a bit different: I typically have a very short search strings in the search field on desktop as Ente Auth runs all day, something like "aws". The search field is way longer than that text and I tend to double-click somewhere in the search field rather than directly on the text. The change makes sure that the text still gets selected, so I can double-click anywhere in the field and start typing a new search term.

Instead of clicking, on OSX, you can use ``cmd + f`` followed by ``cmd +a`` to select everything. Would this work for you instead of adding this particular change?

mariow:
@ua741 No, for me it doesn't. I like the flow of double-clicking the field and then starting to type. But I understand that this might be an individual problem, so perfectly fine for me if the decision is to reject the PR.

PR #9809, opened 2026-03-26
[mob][photos] improve bg sync

prateekmedia:
## Description

- Adds BgProcessingTask
- Reduces iOS frequency to 15mins instead of 30mins
- Dynamic expiration time instead of the fixed timeout based expiration on iOS
- Firebase sync uses _runMinimally


prateekmedia:
@codex review

PR #9832, opened 2026-03-27
Update app_ru.arb

bubblecrabs:
The Russian translation of the button has been corrected.


mobile/apps/auth/lib/l10n/arb/app_ru.arb
```diff
@@ -10,7 +10,7 @@
   "featureBackupCodes": "Сквозное шифрование резервного копирования и синхронизация",
   "featureSearchEtc": "Поиск, теги, поделиться,\nследующий одноразовый код и многое другое",
   "featureOpenSource": "Открытый исходный код\nпроверенный экспертами по безопасности",
-  "signUp": "Войти",
+  "signUp": "Регистрация",
```
ua741, line 13: Thank you for the PR. We’ve fixed this in Crowdin, which is the source of truth for all localization.
[Ref: ](https://github.com/ente-io/ente/blob/main/CONTRIBUTING.md#translate)

PR #9856, opened 2026-03-30
[mob][photos] Switch QR detection to ZXing-CPP native

Amrithesh-Kakkoth:
## Summary
- Replace Java ZXing with ZXing-CPP (C++ native) for QR code detection, reducing scan time from ~4.8s to ~50-280ms (~17-90x faster)
- Add efficient two-pass image loading with `inSampleSize` downsampling (max 1024px), reducing memory from ~80MB to ~5-8MB per scan
- Fix QR overlay tap regions not receiving gestures due to zero-sized inner `Stack` (wrapped with `SizedBox.expand()`)
- Add `ente_qr` Photos-specific plugin with platform channel for Android (ZXing-CPP) and iOS (Vision framework)


laurenspriem:
@codex review

Issue #4749, opened 2025-01-16
Ente Auth iOS: slow unlock


LivioZ:
### Description

With app lock set to Device lock (in my case it is Face ID), when I open the app (or other authentication, for example to enter app lock settings) the "hide content screen" is _relatively_ slow to disappear after authentication is successful. The delay is small, but noticeable.

To make it clearer what I expect, you can set "Focus search on app start". With this option, the keyboard appears before the codes, I expect to see codes the moment the keyboard pops up.

![Image](https://github.com/user-attachments/assets/bb7e9c5d-b7a7-42f2-8b48-87a9c7bc0eba)

### Version

4.2.8

### What product are you using?

Ente Auth

### What platform are you using?

Mobile - iOS

kybos:
10 months later, version 4.4.4:

This issues is still there.

After the app has been unlocked by FaceID, there is approx. 1 second black screen before the code list is shown (not using the Search focus option). 
This on a pretty fast iPhone 16. Such issue is not present with any other app using FaceID for unlocking.

Really a shame, as this black screen delay is really annoying and disturbs the flow in this otherwise very good auth app.

simbabimba-dev:
@kybos @LivioZ  Does this still occur for you guys?

LivioZ:
Hi @LakshmiBhaskarPVL , yes still occurs. Version 4.4.15, iOS 26.2.1

To add something, the same issue is on the Bitwarden app, but not on 2FAS Auth.

kybos:
Yes, it's still there. Same OS/app version as LivioZ.

Bitsnbytes14:
Looked into this and the closed PR (#9884). The delay fix looks good, but I think the regression is that EnteScreenCover gets disabled after unlock and never re-enabled, so the app-switcher blur stops working after the first unlock.
I was thinking of:
Calling setHideAppContent(false, persist: false) in _showLockScreen() after successful auth.
Calling setHideAppContent(true, persist: false) in AppLock.didChangeAppLifecycleState() on paused/inactive (if getShouldHideAppContent() is true).
Using persist: false would just toggle the live cover without changing the user's preference.
Does that sound like the right approach? Happy to open a PR if so.

nikhilmali-02:
[...]

PR #9884, opened 2026-04-01
Fix slow unlock on iOS by disabling privacy screen immediately

brian-c11:
## Description
Fix slow unlock on iOS when using Face ID (device lock). The blur/privacy screen was persisting briefly after successful authentication before showing the codes.

## Changes
- Added `await PrivacyScreen.instance.disable()` immediately after successful authentication in `_showLockScreen()`, before calling `didUnlock()`
- This immediately removes the blur overlay instead of waiting for the lock screen navigation to complete

## Issue
Fixes https://github.com/ente-io/ente/issues/4749

## Testing
- Tested on iOS with Face ID enabled as device lock
- Verified that codes appear immediately when keyboard pops up with "Focus search on app start" option enabled

## Screenshots
Before: Noticeable delay between Face ID success and codes appearing
After: Codes appear instantly when keyboard pops up

### DIFF
--- a/mobile/packages/lock_screen/lib/ui/lock_screen.dart
+++ b/mobile/packages/lock_screen/lib/ui/lock_screen.dart
@@ -13,6 +13,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter_animate/flutter_animate.dart';
 import 'package:logging/logging.dart';
+import 'package:privacy_screen/privacy_screen.dart';
 
 class LockScreen extends StatefulWidget {
   final BaseConfiguration config;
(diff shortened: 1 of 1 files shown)


AmanRajSinghMourya:
Thanks for the PR! The UX issue you're describing is real — the blur lingering after Face ID is noticeable.    
                                                                                     
However, this fix has a security concern. Calling PrivacyScreen.instance.disable() here permanently disables the privacy blur for subsequent app backgrounding. The privacy screen is what hides auth codes in the iOS task switcher. After this      change, once the user unlocks:  

  1. They use the app normally
  2. They swipe to the task switcher or background the app
  3. Auth codes are visible with no blur — the privacy screen was disabled and
  nothing re-enables it                                                              
  4. The lock screen only appears after the auto-lock timer delay
                                                                                     
The setHideAppContent() method in LockScreenSettings is the only place that re-enables the privacy screen, and it's only called when the user toggles the setting — not on app lifecycle.
    
Closing this for now as the current approach introduces a security regression, but we've noted the underlying UX issue and will look into fixing it properly on our end. Thanks for identifying the problem!

PR #9975, opened 2026-04-07
[web] Fix timeline sort for mixed image/video uploads

### LINKED ISSUE #8321: Photos sorting incorrectly in timeline despite correct timestamps
### Description

I'm experiencing a sorting issue where photos appear out of chronological order in my albums, even though they display the correct capture time when I view their info.

What's happening:

A photo taken at 7:39 PM is appearing before photos taken at 7:29 PM and 4:26 PM on the same day. The photo info panel shows the correct time (7:39 PM), but it's just sorted in the wrong position. If I edit the photo itself and modify the date/time manually, it goes to the correct location.

Example:
One of the affected photos has this EXIF data:

```
Date/Time Original              : 2013:08:05 19:39:41
Create Date                     : 2013:08:05 19:39:41
Modify Date                     : 2013:08:07 21:53:57
Software                        : Adobe Photoshop CS6 (Macintosh)
Time Created                    : 19:39:41+00:00
File Modification Date/Time     : 2024:03:25 14:51:39-07:00
```

This photo was edited in Photoshop a couple days after it was taken, which created multiple date fields. It seems like Ente might be using the wrong date field for sorting in some cases.

This has happened on another album with mp4s not being sorted properly despite me editing their metadata.

The timestamps are being read correctly (they show up right in the info panel), but the sorting logic seems to be off.

### Version

v1.7.16

Ente Photos

Desktop - macOS

willsoto:
Fixes https://github.com/ente-io/ente/issues/8321

Modifies `fileCreationTime()` to derive its epoch from the `dateTime` string
in `pubMagicMetadata` instead of `metadata.creationTime`. The `dateTime` field
always represents local date/time regardless of source format, ensuring
consistent sort ordering between images (EXIF local time) and videos (FFmpeg
UTC time with "Z" suffix). This fixes both existing and new uploads since
`dateTime` is already stored for all files with extracted metadata.
User-edited times still take priority.

### DIFF
--- a/web/packages/media/file-metadata.ts
+++ b/web/packages/media/file-metadata.ts
@@ -476,9 +476,30 @@ export const fileFileName = (file: EnteFile) =>
  *
  * While sometimes the epoch timestamp is the correct value to use, it is also
  * possible that {@link fileCreationPhotoDate} might be more appropriate.
+ *
+ * When the file has a {@link dateTime} string in its public magic metadata, we
+ * derive the epoch from it instead of using the baked {@link creationTime}.
+ * This ensures consistent sort ordering across images (whose EXIF dates are
+ * local time without offset) and videos (whose FFmpeg dates are often UTC with
+ * a "Z" suffix). The {@link dateTime} field always represents the local
+ * date/time regardless of how the original metadata encoded the timezone.
+ *
+ * See: https://github.com/ente-io/ente/issues/8321
+ *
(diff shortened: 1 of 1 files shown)


mnvr:
Thank you for the PR! 

Superceded by https://github.com/ente-io/ente/pull/10302 that fixes the issue but in a different way.

PR #10007, opened 2026-04-10
[mob][photos] fix add to shared album for copied files

AmanRajSinghMourya:
## Description
Adding an already-uploaded photo from a shared album into another shared album fails when the acting user is a collaborator/admin but does not own the destination album.

The failure happens in apps/photos/lib/services/collections_service.dart:1987 because foreign-owned files go through _copyToCollection(...), and apps/photos/lib/services/collections_service.dart:2088 requires the destination collection to be
 owned by the current user. 
 
 For foreign-owned files:
  - if the destination album is owned by the current user, behavior stays the
    same and uses the old direct copy flow
  - if the destination album is not owned by the current user but canAutoAdd(...) is allowed, the service now:
      1. copies those files into the user’s uncategorized album
      2. takes those newly user-owned copies and adds them to the destination album through the normal add flow

### DIFF
--- a/mobile/apps/photos/lib/services/collections_service.dart
+++ b/mobile/apps/photos/lib/services/collections_service.dart
@@ -1828,7 +1828,7 @@ class CollectionsService {
         final currentUserID = _config.getUserID()!;
         final shouldCopyViaUncategorized = dstCollection != null &&
             !dstCollection.isOwner(currentUserID) &&
-            dstCollection.canAutoAdd(currentUserID);
+            dstCollection.canAdd(currentUserID);
 
         final int copyDestinationCollectionID;
         if (shouldCopyViaUncategorized) {
--- a/mobile/apps/photos/lib/services/collections_service.dart
+++ b/mobile/apps/photos/lib/services/collections_service.dart
@@ -1847,7 +1847,7 @@ class CollectionsService {
         if (shouldCopyViaUncategorized) {
           await _addToCollection(
             dstCollectionID,
-            copiedFiles.map((file) => file.copyWith()).toList(),
+            copiedFiles,
           );
         }
       }
--- a/mobile/apps/photos/lib/services/collections_service.dart
+++ b/mobile/apps/photos/lib/services/collections_service.dart
@@ -39,6 +39,7 @@ import "package:photos/models/metadata/collection_magic.dart";
 import "package:photos/service_locator.dart";
 import 'package:photos/services/app_lifecycle_service.dart';
 import "package:photos/services/favorites_service.dart";
+import "package:photos/services/hidden_service.dart";
 import 'package:photos/services/memory_share_service.dart';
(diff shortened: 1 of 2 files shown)


Trekky12:
What if the photos are already uploaded in the destination album? So for example when photos are shared in one album by the admin a collaborator has them on their home view. If this collaborator selects photos to add to the shared album and marks these photos already in this album and shared to him they shouldn’t be copied another time to the album in my opinion. This is misleading for non technical people…

AmanRajSinghMourya:
@Trekky12 
This PR is addressing a different issue.

The bug here was that a collaborator/admin could not add files from one shared album to another shared album when they did not own the destination album. That failed because foreign-owned files cannot be directly added through that path.

The fix is to first create a user-owned copy in the acting user’s uncategorized album, and then add that owned copy into the destination album.

So in the case you’re describing, the file in the destination would not be the same file record as the already-shared one. It is a different user-owned copy with a different `uploadedFileID`, not the same shared file being re-added.

Trekky12:
Okay thank you for your response. So in my described use case this leads to the issue that the photo is duplicated in the album, right? Is there a solution for that?

mobile/apps/photos/lib/services/collections_service.dart
```diff
@@ -1822,33 +1823,74 @@ class CollectionsService {
...
-            .contains(file.uploadedFileID)) {
-          _logger.warning(
-            "skip copy, duplicate ID: ${file.uploadedFileID} in collection "
-            "${file.collectionID}",
+      if (filesToCopy.isNotEmpty) {
+        final dstCollection = _collectionIDToCollections[dstCollectionID];
+        final currentUserID = _config.getUserID()!;
+        final shouldCopyViaUncategorized = dstCollection != null &&
+            !dstCollection.isOwner(currentUserID) &&
+            dstCollection.canAutoAdd(currentUserID);
```
ashilkn: `canAutoAdd` seems to be something smart-album specific? If the goal is to check if this user can add to the destination collection, let's create a different helper method like `canAdd()`. Same as `canAutoAdd()` but without the `isFavoritesOrUncategorized` check.

mobile/apps/photos/lib/services/collections_service.dart
```diff
@@ -1822,33 +1823,74 @@ class CollectionsService {
...
+
+        final copiedFiles = await _copyFilesToCollection(
+          filesToCopy,
+          dstCollectionID: copyDestinationCollectionID,
+        );
+
+        if (shouldCopyViaUncategorized) {
+          await _addToCollection(
+            dstCollectionID,
+            copiedFiles.map((file) => file.copyWith()).toList(),
```
ashilkn: Redundant copyWith(). Check line 1875

ashilkn (approved):

PR #10023, opened 2026-04-12
[mobile][photos] Enable text selection during zoom

Amrithesh-Kakkoth:
- Add ZoomTransform (scale + offset) propagated via InheritedDetailPageState
- Subscribe to PhotoViewController.outputStateStream in ZoomableImage to push continuous transform updates on every gesture frame
- Wrap OCR overlay in a Transform that mirrors the zoom instead of hiding it; text boundaries stay aligned with the image during pinch-to-zoom
- Overlay becomes interactive 200ms after gesture settles
- Use a Listener to count active pointers; force IgnorePointer when 2+ fingers are down so pinch gestures always reach PhotoView
- Increase teardrop handle hitbox from 52 to 72px for easier grabbing


Amrithesh-Kakkoth:
@codex review

laurenspriem:
@codex review

PR #10027, opened 2026-04-12
[web][paste] improving ente paste onboarding copy, send loader, and link/QR actions

vnay-dev:
### Headline on the create screen
New visitors only see a paste field, so it isn’t obvious what happens next. I added a short headline so it’s clear upfront that you’re creating a secure, one-time link to share private text—not just “pasting” into nowhere.

### Fewer duplicate trust badges
The headline already explains the idea at a glance, so the extra pills felt repetitive. I trimmed them so the page doesn’t repeat the same message in two places.

### Send button loading state
In production, the spinner on the send button looked like it was orbiting instead of spinning in place. I adjusted the layout so the loader stays fixed in the center and only the indicator rotates.

### Copy next to the generated link
After a link is created, copying it is the natural next step. The “copy” hint below the row doesn’t read as a control, so I added a clear copy action beside the link, with feedback when it’s copied.

### Close control for the QR code
The QR toggle wasn’t obvious—I didn’t realize the same icon shows and hides the code. I added a dedicated close control on the QR view that doesn’t cover the code, so people can dismiss it without guessing.

pr-review1

### DIFF
--- a/web/apps/paste/src/features/paste/components/PasteCreatePanel.tsx
+++ b/web/apps/paste/src/features/paste/components/PasteCreatePanel.tsx
@@ -43,27 +43,14 @@ export const PasteCreatePanel = ({
     const isNearCharLimit = inputText.length >= nearLimitThreshold;
     const isCreateDisabled = isInputEmpty;
     const privacyPills = [
+        "Private",
         isMobile ? "E2EE" : "End-to-end encrypted",
+        "One-time view",
         "Auto-deletes after 24 hours",
     ];
 
     return (
         <Box sx={{ width: "100%", maxWidth: "100%", minWidth: 0 }}>
-            <Typography
-                component="h2"
-                sx={{
-                    color: tokens.text.primary,
-                    fontWeight: 600,
-                    fontSize: { xs: "1.08rem", sm: "1.18rem" },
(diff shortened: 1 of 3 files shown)


AswinAsok:
hey @vnay-dev 

first of all thanks for your contribution.

i went through your changes, two things i'd like reverted:
- the heading on create screen - the "one-time link" is already prominently shown when the link is generated
- the duplicate trust badges - since we are removing the heading.

also the "Copy" icon next to the link - share/copy only show after link creation so users will get it 
(otherwise we'd need to move share to the same row too for consistency).

btw, i do agree the Send button loading state and Close control for the QR code are valid ones.

thanks for fixing these.

vnay-dev:
Thanks for the feedback!👍 @AswinAsok  
I’ve made the corrections and updated the PR.

AswinAsok:
hey @vnay-dev 

looks great overall 

just noticed one thing, the IconButton code is duplicated in the **PasteLinkCard** component. 
could you move it into the **PasteQrCode** component instead?

also, do keep the current styles as such for the close button.

vnay-dev:
Hi @AswinAsok, I have added the changes.

AswinAsok (approved):

PR #10124, opened 2026-04-18
Nix flake based deployment

itihas:


mnvr:
I'll be closing drafts, but I presume you opened it to get some feedback as you work on this, so: 

While I don't personally use nix and so can't comment on the files themselves, it is unlikely we'll put more file in the root of the repo (esp. since we can't commit to maintaining them), but yes, I do like the idea of having better support for nix! So once they're done, we could keep them in some other folder (e.g., infra/nix) and/or add instructions in the self hosting docs.

Screenshot 2026-04-22 at 9 30 45 AM

PR #10193, opened 2026-04-24
[web] Add extra auth web issuer legacy period parsing

mnvr:
Pre-v3.1.8 Ente Auth built otpauth:// URIs without URI-encoding the issuer, so issuers containing URL-unsafe characters produced malformed query strings where `period=N` folded into the issuer value. The root cause was fixed in 75c3bc1c84 (Sep 2024).

Web has carried a mitigation since the viewer's first commit (4faa05a465, Mar 2023), but it only stripped a trailing literal `period` — i.e. `issuer=Xperiod`. Some exports instead surface as `issuer=Xperiod=30` (different URL parsers react differently to the malformed input), which the old check missed. Handle that second shape too.


mnvr:
@codex review

Issue #3008, opened 2024-08-27
Not possible to add file directly in album where user have collaboration right


Tuscann:
### Description

1. Open https://web.ente.io/gallery
2. Login with user bai ivan
3. Open album Camera Roll where user bai ivan has right for collaborator (Camera Roll is album created from user mancho)
4. Click Upload -> File -> Choose ivan.png
5. Check albums ?? Missing album Camera Roll where to add new file

https://github.com/user-attachments/assets/47d85214-e50e-4c09-99e0-718bd2b9388a

### Version

WEB

### What product are you using?

Ente Photos

### What platform are you using?

Web - Chrome

PR #10278, opened 2026-05-04
Allow collaborators to upload directly to shared albums

dev-goalz:
Closes the gap behind #3008. Previously the museum POST /files endpoint hard-rejected any actor who was not the collection owner, so the mobile client worked around it by uploading to the collaborator's Uncategorized and then symlinking into the shared album. This produced duplicate rows, charged storage to the wrong user, and surprised collaborators when the "copy" persisted in their account after the share ended.

Server (museum):
- Migration 121 adds collection_files.added_by_user_id; legacy rows stay NULL and readers fall back to f_owner_id.
- POST /files now uses AccessCtrl.GetCollection + Role.CanAdd() instead of the owner-only check. file.OwnerID is pinned to the album owner (server-side, ignoring whatever the client sent), quota is charged to the album owner, and the actor is recorded on the collection_files row as added_by_user_id.
- POST /files/copy gets the same loosening so collaborators can copy existing files into a shared album without the Uncategorized detour.
- RemoveFilesV3 is repartitioned by added_by_user_id (with f_owner_id fallback for legacy rows). Collaborators can remove the files they themselves added even though those files are now owned by the album owner; the OWNER/ADMIN/suggest-action paths are preserved.

Web:
- CollectionSelector picker uses canAddToCollection for the upload destination list so collaborator-shared albums show up.

Mobile:
- collections_service.dart drops the shouldCopyViaUncategorized branch in addOrCopyToCollection; with the server changes the direct copy works.

Backwards compatibility: old clients still hitting the two-step add-files flow continue to work — those code paths are untouched. New clients get a single collection_files row per upload to a shared album.

### DIFF
--- a/mobile/apps/photos/lib/services/collections_service.dart
+++ b/mobile/apps/photos/lib/services/collections_service.dart
@@ -1889,32 +1889,10 @@ class CollectionsService {
       }
 
       if (filesToCopy.isNotEmpty) {
-        final dstCollection = _collectionIDToCollections[dstCollectionID];
-        final currentUserID = _config.getUserID()!;
-        final shouldCopyViaUncategorized = dstCollection != null &&
-            !dstCollection.isOwner(currentUserID) &&
-            dstCollection.canAdd(currentUserID);
-
-        final int copyDestinationCollectionID;
-        if (shouldCopyViaUncategorized) {
-          final uncategorizedCollection =
-              await CollectionsService.instance.getUncategorizedCollection();
-          copyDestinationCollectionID = uncategorizedCollection.id;
-        } else {
-          copyDestinationCollectionID = dstCollectionID;
-        }
-
-        final copiedFiles = await _copyFilesToCollection(
+        await _copyFilesToCollection(
           filesToCopy,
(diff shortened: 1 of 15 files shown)


ua741:
Hello,
Thank you for the PR but unfortunately, we will have to close this. The current behaviour is designed in such that where it allows users to add link to their own photos in various albums owned by anyone including their own. 
If/When we add an option where we allow collaborators to directly upload to albums owned by some else, we would want to also give more control to album owner to disable this based on either admin or separately setting as the collaborator can accidentally consume all storage of the album owner.

dev-goalz:
@ua741 if I would be able to extend this PR to your design, would this be a mergeable feature?
The user-case I'm trying to solve is that I would like to share a album with a friend and we both have the option to upload media directly to the album without holding a additional personal "Uncategorized" copy.

ua741:
This would require our product / design team to think of various cases before we can actually roll it out for all folks. Currently, it's not prioritised yet. So, I don't think I can suggest quick changes that would help us merge this change. I would request you to start / upvote / comment on relevant GitHub discussions with your requirements.
Cc @setal

PR #10297, opened 2026-05-05
[locker][web] Fix lint

vishnukvmd:


mnvr:
I just fixed it in a different PR - https://github.com/ente-io/ente/pull/10296 :)

PR #10321, opened 2026-05-07
[mobile] Fix mobile localization placeholders and plurals

mnvr:
Replace hardcoded Photos web/cast URLs in localized strings with normal ARB placeholders, and pass the current web/cast values from code.

Update the affected Photos and Locker source ARB plural selectors from =1 to one. Translated plural selectors are left as-is to preserve locale-specific behavior.


mnvr:
@codex review

PR #10337, opened 2026-05-08
[mob][photos] Batch large free up space deletes

AmanRajSinghMourya:
## Summary

Fixes large free-up-space operations getting stuck/failing when the app asks the OS to delete too many local files in one request.

On Android 11+, `photo_manager` routes deletes through `MediaStore.createDeleteRequest`. For our target SDK, Android caps that request at 2000 URIs. The failing user log showed the Camera album trying to delete 7014 files in one request, which failed with `PlatformException(deleteWithIds failed)` before Android could show the delete confirmation popup.

This change lowers the existing large-delete threshold from 20000 to 1900, so large free-up-space deletes use the existing batch delete flow. The same threshold is used as the retry helper's minimum stop threshold so the helper starts and stays at the Android-safe batch size.


AmanRajSinghMourya:
@codex review

mobile/apps/photos/lib/utils/delete_file_util.dart
```diff
@@ -346,8 +346,10 @@ Future<bool> deleteLocalFiles(
   final List<String> localAssetIDs = [];
   final List<String> localSharedMediaIDs = [];
 
-  const largeCountThreshold = 20000;
-  final tooManyAssets = localIDs.length > largeCountThreshold;
+  // Keep large platform asset deletes in smaller batches. Android 11+ routes this
```
ashilkn: Update this comment for better clarity, and add a todo for refactoring (by removing the recursive batching) if this solution works on users that are stuck .

ashilkn (dismissed):

ashilkn (approved):

PR #10340, opened 2026-05-08
[Ensu] Optimize inference for image attachments across all platforms except web

laurens-pilot:
## Description

- Store images in optimized format in app
- Basic caching of vision encoder
- Better viewing of image attachments in the app

## Testing

Testing is done on following platforms:
- Android
- iOS
- Mac

Pending (for release QA, as it's highly unlikely there are issues here):
- Linux
- Windows

Not needed as no changes made here:
- Web


mobile/native/darwin/Apps/Ensu/scripts/build-rust.sh
```diff
@@ -109,6 +110,16 @@ build_crate() {
             rustflags+=" -C link-arg=$min_flag=$IPHONEOS_DEPLOYMENT_TARGET"
         fi
 
+        local clang_rt
```
mnvr: Instead of this, changing the subsequent line to only build the staticlib

```sh
(cd "$crate_dir" && RUSTFLAGS="$rustflags" cargo rustc "${cargo_flags[@]}" --target "$target" --lib --crate-type staticlib)
```

should work, the dylib is only needed during codegen which happens on the macOS host, not on iOS.

mnvr (approved):

PR #10360, opened 2026-05-10
[mob][photos] Fix HEIC orientation flip on Android viewer

AmanRajSinghMourya:
## Issue

- iPhone HEIC photos rendered correctly as thumbnails but flipped onto their side once the full image loaded on Android.
- Reproduces with any iPhone HEIC having EXIF Orientation 3, 6, or 8 (any portrait shot).
- Same files render correctly in Google Photos and on iOS Ente, so the issue is Android-specific.

## Fix

- Before rendering a HEIC on Android, read its EXIF orientation. If it requires rotation, route the load through `_loadInSupportedFormat` instead of `Image.file`.
- `_loadInSupportedFormat` already calls `FlutterImageCompress.compressWithFile` with `autoCorrectionAngle: true`, which reads EXIF via `androidx.exifinterface.ExifInterface` and physically rotates the bitmap before returning bytes.
- Root cause: Flutter's `Image.file` on Android decodes HEIC via `BitmapFactory`/Skia, neither of which applies the EXIF orientation tag for HEIC. The Rust path that does apply it is still gated to internal users.

https://github.com/user-attachments/assets/354631ee-c8d5-415a-bd3a-3ebceb3921df

https://github.com/user-attachments/assets/4fdd8f0f-cc61-46ff-b38d-bfacf0e7507d

### DIFF
--- a/mobile/apps/photos/lib/ui/viewer/file/zoomable_image.dart
+++ b/mobile/apps/photos/lib/ui/viewer/file/zoomable_image.dart
@@ -26,6 +26,7 @@ import "package:photos/theme/ente_theme.dart";
 import "package:photos/ui/actions/file/file_actions.dart";
 import 'package:photos/ui/common/loading_widget.dart';
 import 'package:photos/ui/viewer/file/thumbnail_widget.dart';
+import "package:photos/utils/exif_util.dart";
 import 'package:photos/utils/file_util.dart';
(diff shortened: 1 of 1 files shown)


AmanRajSinghMourya:
@codex review

ua741 (approved):

PR #10377, opened 2026-05-11
Remove mobile Crowdin source push workflows

mnvr:
## Summary
- Remove the per-app mobile Crowdin source push workflows for Photos, Auth, and Locker
- Keep source uploads tied to the existing Monday mobile Crowdin sync workflow, which also downloads translations

## Testing
- Not run; workflow deletion only


simondubrulle:
@mnvr It makes sense to remove those workflows but is there a reason why you would not run the [mobile-crowdin-sync.yml](https://github.com/ente-io/ente/blob/main/.github/workflows/mobile-crowdin-sync.yml) workflow when a change is detected on any of the `XXX_en.arb` files or any of the `crowdin.yml` files? Especially when the workflow takes less than 1 minute?

I am asking because now it also means that if a new string is introduced, we have to wait for the next monday for the sync to push it to Crowdin, only then contributors can add translations on Crowdin and then we have to wait for the next monday to have them in the main branch (and potentially miss a release). It will take at least 1 week (at best) to 2 weeks (at worst) to get it in the main branch.

I am also asking because I am planing to update the `translations.md` doc which is not correct anymore :

https://github.com/ente-io/ente/blob/29dd464b8201e753a2782b3feaad5f6c317f3428/mobile/apps/photos/docs/translations.md?plain=1#L57

mnvr:
Thanks for the feedback. The change was because sometimes (infrequently) a PR would be merged but then the string would change based on internal feedback. So the delay gave some stabilization buffer, to avoid wasted effort in translating strings that immediately get changed again.

I've reverted the change and restored the source push on main merge — https://github.com/ente-io/ente/pull/10719, it'll now remove the delay. Now I've added marker functions and will encourage contributors to use them in case there is any churn expected.

PR #10382, opened 2026-05-11
[mob][photos] Handle missing thumbnail cache files

ua741:
## Summary
- Treat missing thumbnail cache files as cache misses when Android removes files from the app cache directory between `exists` and `readAsBytes`.
- Recreate the thumbnail cache directory on missing-path write failures, then retry the best-effort cache write.
- Keep this handling local to the thumbnail path instead of adding a Sentry filter.

Sentry: PHOTOS-MOBILE-9PQ

## Testing
- `dart format mobile/apps/photos/lib/utils/thumbnail_util.dart`
- `dart analyze lib/utils/thumbnail_util.dart`
- `flutter analyze` still fails on existing localization/generated-string errors outside this diff.


ua741:
@codex review

PR #10411, opened 2026-05-13
[mob][photos] Initialize ServiceLocator before Configuration main init

laurenspriem:
## Description

This fixes a startup ordering issue where Configuration.init() could trigger auto-logout before ServiceLocator was initialized. The app now bootstraps configuration preferences first, initializes networking and the service locator, and only then runs the full configuration init that may perform cleanup.


ua741:
Closed as part of https://github.com/ente-io/ente/pull/10421

PR #10440, opened 2026-05-14
[server] Allow specific emails to register when registration is disabled

rickiewars:
## Description

When registrations are disabled, there is no way to invite specific users to create an account. The current workaround is to temporarily enable registrations, let the user finish the registration and then disable it again.

To solve this, I added an option to allow certain users to register themselves using one of te allowed emails. This can later be extended to a proper invitation system if needed but this works until then.

## Tests

I have manually tested the functionality against a local instance.
During testing I used the `hardcoded-ott.emails` setting to hardcode a verification code.
I've followed the following test plan:

- In `local.yaml`, set `disable-registration` to `false`
- Restart the museum server
  - Register using `a@example.com`
  - **Verify** that registration succeeds -> **Success**
- In `local.yaml`, set `disable-registration` to `true`
- Restart the museum server
  - Register using `b@example.com`
  - **Verify** that registration fails with a forbidden -> **Success**
  - Login with `a@example.com`
  - **Verify** that the login flow still works -> **Success**
- In `local.yaml`, add `b@example.com` to the `allowed-registrations` list
- Restart the museum server
  - Register using `b@example.com`
  - **Verify** that registration succeeds -> **Success**
  - Login with `a@example.com`
  - **Verify** that the login flow still works -> **Success**
  - Register using `c@example.com`
  - **Verify** that registration fails with a forbidden -> **Success**


mnvr:
This is too niche. Sorry, I understand that you might need this, but every knob increases surface area. In particular for museum we want to minimize the branching factor of the code.

Issue #10470, opened 2026-05-17
[locker] missing header (invalid token error)


albundy83:
### Description

Hello,

I have tried to use and host myself the locker web app.

To give you a bit of background, museum is hosted on something like`museum.youpi.fr` and locker on `locker.youpi.fr`

I have found a smal error that prevent to make it works.

When I connect to my self hosted web app locker.youpi.fr for example, I can connect successfully.

But when I try to download a file that I have uploaded using the android app (because web app did not exist when I start using locker), I have the error `{"error":"invalid token"}` on the call https://museum.youpi.fr/files/download/10021337?token=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

I'm not a developper, so my IA friend help me to fix the issue by modifying the file [remote-read.ts](https://github.com/ente-io/ente/blob/main/web/apps/locker/src/services/remote-read.ts).

Here the updated lines:
https://github.com/ente-io/ente/blob/adea3fb6eeabbe881b768535b3254470f0fcadbf/web/apps/locker/src/services/remote-read.ts#L5
```js
import {
    authenticatedRequestHeaders,
    ensureOk,
    publicRequestHeaders,
} from "ente-base/http";
```

and
https://github.com/ente-io/ente/blob/adea3fb6eeabbe881b768535b3254470f0fcadbf/web/apps/locker/src/services/remote-read.ts#L1047
```js
            response = await fetch(url, {
                headers: publicRequestHeaders(),
            });
```

I will create a pull request.

Thanks a lot for your incredible applications

### Version

latest

### What product are you using?

Ente Photos, Ente Auth

### What platform are you using?

Kubernetes 1.35

mnvr:
Thank you, should be fixed by https://github.com/ente-io/ente/pull/10471

PR #10471, opened 2026-05-17
[locker] Fix missing header when self hosted

albundy83:
## Description
Fix for the self hosted locker web app, see details [here](https://github.com/ente-io/ente/issues/10470)


mnvr (approved):
Thank you!

mnvr:
Fixes - https://github.com/ente-io/ente/issues/10470

PR #10569, opened 2026-05-21
[mob][photos] Use renamed filenames for Photos gallery downloads

### LINKED ISSUE #8541: Renaming an image file is impossible (misleading)
### Description

Steps to reproduce the problem:

- Open the Ente photos app on your mobile device.

- Open an existing photo, tap "Info," and then rename the file, for example, "toto" .jpg.

- Go back and check the file information again. The file displays correctly, it is named "toto.jpg." This is expected.

- Go back and download the image.
- check the file name on your device.

The image is not named "toto.jpg" => KO

Explanation:

Either it's not possible to rename the file, in which case the corresponding field shouldn't allow this action, or renaming is allowed, and in that case, the file should be downloaded with its new name. Does that make sense?

### Version

1.12.31

Ente Photos

Mobile - Android

neeraj-pilot:
## Summary
- use the user-facing `displayName` when saving direct gallery downloads
- keep edited filenames for image, video, and live-photo gallery save paths
- fall back to the original title or local path basename if displayName is unavailable

Fixes #8541

## Tests
- `flutter pub get --enforce-lockfile` in `mobile/apps/photos`
- `flutter analyze lib/utils/file_download_util.dart`
- `git diff --check`

### DIFF
--- a/mobile/apps/photos/lib/utils/file_download_util.dart
+++ b/mobile/apps/photos/lib/utils/file_download_util.dart
@@ -60,6 +60,18 @@ String getDownloadSkipToastFileName(EnteFile file) {
   return title.isNotEmpty ? title : displayName;
 }
 
+String _getGallerySaveTitle(EnteFile file, String fallbackPath) {
+  final displayName = file.displayName;
+  if (displayName.trim().isNotEmpty) {
+    return displayName;
+  }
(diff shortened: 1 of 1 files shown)


ua741 (approved):

ua741:
@codex review

PR #10597, opened 2026-05-22
[ensu] Fix Ensu desktop version display

mnvr:


mnvr:
(the desktop lint is failing for an unrelated reason, fixing in https://github.com/ente-io/ente/pull/10586)

PR #10702, opened 2026-05-27
[mob][photos] Introduced 'archiveCollectionName' translation string

simondubrulle:
## Description

For specific languages (like French), there is a need to make a distinction between "Archive" (the collection name) and "Archive" (the action of archiving). 

In French, the collection name would be "Archive**s**" and the action of archiving would be "Archive**r**".

With the new UI design, we see the "Archive" collection name in the "more" section of the new album management sheet. Currently it is using the translation string "archive" and it therefore shows an action in some languages (which feels off).

## Proposed changes

I have introduced a new translation string named 'archiveCollectionName' in `intl_en.arb`. It is replacing the "archive" translation string in the `albums_manage_sheet.dart`.

I have also added more context to the "archive" and "unarchive" translation strings in `intl_en.arb` to emphasize this is an action and not the archive collection name

I followed instructions in [translations.md](https://github.com/ente-io/ente/blob/main/mobile/apps/photos/docs/translations.md) document.

## Tests
Ran `flutter gen-l10n` succesfully.

### DIFF
--- a/mobile/apps/photos/lib/l10n/intl_en.arb
+++ b/mobile/apps/photos/lib/l10n/intl_en.arb
@@ -3212,5 +3212,4 @@
   "@archiveCollectionName": {
     "description": "Name of the 'Archive' collection"
   }
-
 }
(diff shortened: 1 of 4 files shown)


ua741 (dismissed):

ua741 (approved):

ua741:
Thank you.

PR #10741, opened 2026-05-29
[Ensu] Don't load model into memory on very low RAM devices (below 4 GB)

laurens-pilot:
## Description

Don't load model into memory on very low RAM devices (below 4 GB). Instead it will show a banner explaining that sending messages is disabled.


laurenspriem:
@codex review

mnvr (approved):

Issue #6608, opened 2025-07-22
Volume stays on maximum when sharing video via link


ArtjomSchwenk:
### Description

When sending a video via the "send link" feature and viewing it over the browser on PC (Firefox & OperaGX) the volume is raised to the maximum and cannot be adjusted. I can only mute or unmute the audio.

In my opinion the volume should be managed through the ente player (which looks like youtube?) and not via the system settings for the browser.

### Version

1.1.53

### What product are you using?

Ente Photos

### What platform are you using?

Web - Firefox, Mobile - FDroid, Web - Chrome

Issue #5441, opened 2025-03-26
Exporting codes in encrypted format generates a .txt file instead of a .json


Greysys:
### Description

Explained in the title. When I try to create an encrypted export, after inserting the password, Ente created a .txt file instead of a .json

### Version

4.3.2

### What product are you using?

Ente Auth

### What platform are you using?

Mobile - Android, Mobile - FDroid

AmanRajSinghMourya:
This issue has been fixed [here](https://github.com/ente/ente/pull/9089)
Please update the app to the latest version and lets us know if issue still exist.

Issue #451, opened 2024-02-15
Can't type in desktop app after logging out.


nikunjkumarnakum:
**Describe the bug**
The auth app on desktop when asked for login credentials after logging out, I cant type with my keyboard.

**To Reproduce**
1. Go to 'auth desktop app'
2. Click on 'Logout'
3. Do this 'Now try login and typing your email'
4. It won't type

**Expected behavior**
The app wont type until you restart the app. It will only allow you to paste the email from clipboard.

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Version [2.0.31 Nightly]

prateekmedia:
Temporary fix is to switch focus to other app and then back to ente Auth.

Meltryllis163:
If it could be fixed by switching focus, it maybe similar to #2262.

ua741:
This should be fixed in the available version on Github. Please comment here if that's not the case.

PR #10748, opened 2026-05-30
 Fix quickstart.sh POSIX compatibility, auth export MIME types, video volume slider, and auto-logout keyboard handler

### LINKED ISSUE #6324: Broken quickstart script
### Description

Hey, just trying to test ente, and running the suggested quickstart.sh script fails.

I suspect it’s because the `dcv`value [in the script]( https://github.com/ente-io/ente/blob/main/server/quickstart.sh) is an empty string, whereas in previous versions it was 
`dcv="docker compose version --short 2>/dev/null"`

Or am I missing something obvious?

### Version

_No response_

### Last working version

_No response_

_No response_

_No response_

Dryader:
Some old bugs i saw in your issues, focused on small issues that were clearly bugs and not design decisions on your part that could be easily fixed with a quick ai prompt.

[server] Fix quickstart.sh using bash syntax in #!/bin/sh script — replace [[ =~ ]] with POSIX-compatible [ ] || [ ] so the script works on Debian/Ubuntu where sh is dash.

[auth] Use correct MIME type for encrypted code exports — encrypted exports now use MimeType.json / application/json instead of MimeType.text / text/plain, fixing .txt files being generated instead of .json.

[web] Add volume slider to video player controls — add to the media-chrome control bar in both the gallery and public albums viewers, giving users granular volume control instead of only mute/unmute.

[auth] Fix stale keyboard handler after auto-logout on desktop — the auto-logout path (session expiry) did not clear the route stack, leaving the HomePage mounted with its global keyboard handler registered. Add pushNamedAndRemoveUntil after logout to properly dispose the old widget tree.

Fixes #6324, #5441, #6608, #451

### DIFF
--- a/mobile/apps/auth/lib/ui/account/logout_dialog.dart
+++ b/mobile/apps/auth/lib/ui/account/logout_dialog.dart
@@ -66,4 +66,7 @@ Future<void> _logout(BuildContext context, AppLocalizations l10n) async {
   await dialog.show();
   await Configuration.instance.logout();
   await dialog.hide();
+  if (context.mounted) {
+    Navigator.of(context).pushNamedAndRemoveUntil('/', (route) => false);
+  }
 }
--- a/mobile/apps/auth/lib/ui/settings/data/export_widget.dart
+++ b/mobile/apps/auth/lib/ui/settings/data/export_widget.dart
@@ -155,6 +155,8 @@ Future<void> _exportCodes(
   if (!hasAuthenticated) {
     return;
   }
+  final mimeType = extension == "json" ? MimeType.json : MimeType.text;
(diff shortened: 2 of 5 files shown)


mnvr:
Coding with an agent is quick, the hard part is verification :)

1. Don't dump all things in one PR. This is not reviewable or independently mergable.
2. Make sure you actually run and test them. Please.

About the changes:
1. The POSIX change looks file
2. The web change is a behaviour change. Don't include that — that needs discussion not a unilateral UX change.
3. The other two auth bug fixes also look great. Ask your agent to add a brief entry in mobile/apps/auth/changes.

If you don't want to open a new PR, force push here without the web change, a changes/ entry for auth, and after actually testing these things (maybe you've already done so, then just an explicit ack is enough).

PR #10827, opened 2026-06-04
[Docs]Add FAQ entries for sharing, ML, metadata, and auth features

Shanthy17:


setalp:
@Shanthy17 - I think tiff upload is allowed but it doesnt render correctly on the apps.

PR #10849, opened 2026-06-05
feat(web): conditionally enable IPv6 Nginx listeners at startup

upmcplanetracker:
Supersedes #10280

## Description

* **Problem**: The current Nginx configuration only listens on IPv4 sockets. In dual-stack environments (common in modern home labs and container orchestration), this can lead to connectivity issues when the host or network prefers IPv6. However, naively adding unconditional `listen [::]:PORT;` directives causes Nginx to crash on environments where IPv6 is completely disabled, throwing a fatal `Address family not supported by protocol` error.
* **Solution**: 
  1. Added commented-out IPv6 dual-stack listeners (`#listen [::]:PORT;`) to the Nginx configuration template.
  2. Created a lightweight runtime initialization script (`40-enable-ipv6.sh`) inside `/docker-entrypoint.d/` that checks if the container environment supports IPv6 (via `/proc/net/if_inet6`).
  3. The script dynamically uncomments the IPv6 lines right before Nginx starts up *only* if IPv6 networking is available.
* **Impact**: Ensures seamless, out-of-the-box dual-stack network compatibility for modern deployments without changing application logic or breaking workflows for legacy IPv4-only hosts.
* **Refactoring Note**: Switched the Nginx configuration block to a literal heredoc (`<<'EOF'`) to clean up unnecessary backslash escaping and ensure consistent cross-platform behavior across different container engines.

## Tests

* **Configuration Syntax**: Verified the updated Dockerfile heredoc generates valid syntax and handles the standard `try_files` routing directives perfectly.
* **Runtime Conditional Logic**: Verified that the shell script cleanly toggles the commented lines using `sed` when it detects an IPv6 environment.
* **Fallback Safety**: Confirmed that on IPv4-only environments, the script safely logs a notice and allows Nginx to start without crashing.

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -129,10 +129,7 @@ RUN chmod +x /docker-entrypoint.d/90-replace-ente-env.sh
 COPY <<'EOF' /docker-entrypoint.d/40-enable-ipv6.sh
 #!/bin/sh
 if [ -f /proc/net/if_inet6 ]; then
-    echo "IPv6 supported, enabling dual-stack..."
     sed -i 's/#listen \[::\]:/listen [::]:/g' /etc/nginx/conf.d/default.conf
-else
-    echo "IPv6 not supported, keeping single-stack..."
 fi
 EOF
 RUN chmod +x /docker-entrypoint.d/40-enable-ipv6.sh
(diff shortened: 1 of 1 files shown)


mnvr (approved):
Thank you. I've pushed a commit to remove the echoes, because both of them will print each time the container runs, and are just noise after a while (even a bit misleading, since once the IPv6 is latched, it is not unlatched. Nginx's own `/docker-entrypoint.d/10-listen-on-ipv6-by-default.sh` doesn't print them).

upmcplanetracker:
Thank you so much.

PR #10995, opened 2026-06-15
[web] Show link icon on albums shared with both people and a link

### LINKED ISSUE #2052: Sharable link icon not visible when album is shared
### Description

When an album is both shared and has a sharable link only the shared icon is visible, and the link icon is not visible.

Screenshot 2024-06-07 144923
The album on the left has an active sharable link but the icon is not displayed.

### Version

_No response_

Ente Photos

Web - Chrome

yashs33244:
Closes #2052

When an album was shared with people and also had a public link, the album tile only showed the people icon and hid the link icon, because the two states were treated as mutually exclusive.

Track the presence of a public link with a separate `sharedViaLink` attribute and render the people and link icons independently, so an album shared both ways shows both. `sharedOnlyViaLink` keeps its meaning (the link-only case that picks the primary icon), so no other consumer changes. Tightened the tile icon spacing so all icons still fit when both are shown.

### DIFF
--- a/desktop/changes/show-link-icon-for-mixed-shared-albums.md
+++ b/desktop/changes/show-link-icon-for-mixed-shared-albums.md
@@ -1,0 +1,1 @@
+- Fix link icon visibility for albums shared with people and a public link. By @yashs33244.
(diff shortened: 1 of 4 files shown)


AswinAsok (approved):

AswinAsok:
Hi @yashs33244 

Nice work on the fix. I have merged it.

PR #11000, opened 2026-06-16
[mob][photos] Fix add-to-album bottom sheet behavior

AmanRajSinghMourya:
## Summary
- Fix the add-to-album bottom sheet empty-state layout so the create-album row sits at the top instead of vertically centered.
- Move the add-to-album sheet and shared-album confirmation path onto `ente_components` bottom sheet/button/input components.
- Add opt-in modal dismissal support to `ButtonComponent` and cover it with component tests.
- Add the Photos release note.

image png

Screenshot_20260609_142046_Ente_Photos jpg-2

After
Screenshot 2026-06-16 at 11 27 57 AM

### DIFF
--- a/mobile/apps/photos/lib/ui/collections/collection_action_sheet.dart
+++ b/mobile/apps/photos/lib/ui/collections/collection_action_sheet.dart
@@ -2,6 +2,7 @@ import "dart:async";
 import 'dart:math';
 
 import 'package:collection/collection.dart';
+import "package:ente_components/ente_components.dart";
 import 'package:flutter/material.dart';
 import "package:hugeicons/hugeicons.dart";
 import "package:logging/logging.dart";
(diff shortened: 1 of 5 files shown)


Trekky12:
Does this PR fix the problem on iOS that the "Add" button is behind the keyboard when an album is searched?

AmanRajSinghMourya:
> Does this PR fix the problem on iOS that the "Add" button is behind the keyboard when an album is searched?

No it does not solve that but the reason the `Add` button is behind keyboard is because if we show the `Add` button then there will be very less space for the resultant searched albums here    

Simulator Screenshot - iPhone 16e - 2026-06-16 at 17 00 24

Trekky12:
Thank you for the explanation @AmanRajSinghMourya. Nevertheless I think not that tech-savy people don't get that they first need to click done on the keyboard to see the button. I think the current UX is not ideal.

AmanRajSinghMourya:
@Trekky12 Agreed with the feedback and we will work on this to improve the UX

ashilkn:
@Trekky12 I'm working on some UI/UX improvements that includes improvement of what you mentioned.

mobile/apps/photos/lib/ui/collections/collection_action_sheet.dart
```diff
@@ -218,11 +215,9 @@ class _CollectionActionSheetState extends State<CollectionActionSheet> {
                   ],
                 ),
               ),
-              SafeArea(
-                child: Container(
-                  padding: const EdgeInsets.symmetric(vertical: 8),
-                  child: Column(children: [..._actionButtons()]),
-                ),
+              Padding(
```
ashilkn, line 218: Are we sure removing the SafeArea here will not introduce UI regressions? Maybe it could lead to system bottom UI (android nav bar for example) rendering over app UI (buttons here).

AmanRajSinghMourya: Checked on android and iOS there is more padding to the bottom of the button so removed the safe area

ashilkn (approved):

PR #11009, opened 2026-06-16
[mob][photos] update: Photos login components

AmanRajSinghMourya:
## Description

- Update Photos login/signup flow to use newer components
- Fix iOS keyboard Done behavior on login


mobile/apps/photos/lib/ui/account/login_page.dart
```diff
@@ -36,8 +31,8 @@ class _LoginPageState extends State<LoginPage> {
   final Logger _logger = Logger('_LoginPageState');
 
   @override
-  void didChangeDependencies() {
-    super.didChangeDependencies();
+  void initState() {
```
ashilkn, line 34: I'm not sure why we were using `didChangeDependencies()` instead of `initState()` here but I check and found that I did it intentionally (see [here](https://github.com/ente-io/ente/pull/1434/changes#diff-95871335ee86675dc752492e4310c085f4434493758c39a906645ecc0b10931a)). We should go back to using `didChangeDependencies()` if the move to use `initState()` was not a conscious decision.

mobile/apps/photos/lib/ui/account/login_pwd_verification_page.dart
```diff
@@ -116,16 +110,23 @@ class _LoginPasswordVerificationPageState
...
             isPasswordInput: true,
             isRequired: true,
-            autoCorrect: false,
-            autoFocus: true,
-            onChange: (value) {
+            autocorrect: false,
+            autofocus: true,
+            shouldUnfocusOnClearOrSubmit: true,
+            onSubmit: _hasPassword
+                ? (_) async {
```
ashilkn: We can extract the code that runs on `onSubmit` here and on `onTap` at line 85 to a single function and call that.

mobile/apps/photos/lib/ui/account/password_entry_page.dart
```diff
@@ -251,18 +247,31 @@ class _PasswordEntryPageState extends State<PasswordEntryPage> {
...
-                textEditingController: _passwordController2,
+                controller: _passwordController2,
                 isPasswordInput: true,
                 isRequired: true,
-                autoCorrect: false,
+                autocorrect: false,
                 autofillHints: const [],
                 finishAutofillContextOnEditingComplete: true,
+                shouldUnfocusOnClearOrSubmit: true,
+                onSubmit: _passwordsMatch && _isPasswordValid
```
ashilkn, line 261: Code that runs on `onSubmit` should be extracted to a function and used here and at line 121.

ashilkn:
@codex As you can see I've added some comments to follow the DRY principle since there is duplication of code on the `onTap`s of buttons that the `onSubmit`s of the `TextField`s attached to those buttons. Could you check if there are any more of these I haven't flagged?

ashilkn:
@codex Are there anymore places on the onboarding or login flow (the area touched by this PR) where we can add support for submission via buttons in the keyboard ('done' on the iOS keyboard is an example and maybe the only one). `onSubmit` of `TextInputComponent` does the keyboard-button submission afaik?

mobile/apps/photos/lib/ui/account/login_page.dart
AmanRajSinghMourya: Reason: _updateEmail() currently does not read context, theme, localization, or inherited widgets. It only trims the email, writes the controller, and validates. That is exactly the kind of one-time setup initState() is for.
The old didChangeDependencies() was needed because the older code used theme from context inside the email update path. That reason no longer exists. An unguarded didChangeDependencies() is actually riskier here because it can rerun and overwrite a user-edited email when keyboard/MediaQuery dependencies change.

mobile/apps/photos/lib/ui/account/login_pwd_verification_page.dart
```diff
@@ -82,18 +82,21 @@ class _LoginPasswordVerificationPageState
           key: const ValueKey("verifyPasswordButton"),
           label: AppLocalizations.of(context).logInLabel,
           isDisabled: !_hasPassword,
-          onTap: _hasPassword
-              ? () async {
-                  FocusScope.of(context).unfocus();
-                  await verifyPassword(context, _passwordController.text);
-                }
-              : null,
+          onTap: _hasPassword ? _verifyEnteredPassword : null,
```
ashilkn: Redundant `_hasPassword` check. `_verifyEnteredPassword` already checks this.

ashilkn (approved):

PR #11013, opened 2026-06-16
[mobile][photos] Update delete UX

aspizu:
Only doing it for single file delete flow, multiple file delete needs more thought

old warning-red.png was renamed, its only used in two places, maybe needs a better name?

our design system does not have a variant for button with transparent background and gray text color, link variant of button is similar but has accent colored text and underline. so had to roll a temporary component, we should replace this later on with the button component

another issue is that the button component does not let you place an icon on the right side, which is good because we should always have icons on the left side, this is a special case where the icon is on the right side.

labelled control component was never used before so ive modified it to suit my needs
Simulator Screenshot - iPhone 17e - 2026-06-18 at 14 33 26
Simulator Screenshot - iPhone 17e - 2026-06-18 at 14 33 43
Simulator Screenshot - iPhone 17e - 2026-06-18 at 14 33 20


mobile/apps/photos/lib/ui/components/ghost_button.dart
AmanRajSinghMourya, line 1: Lets write a TODO to remove this once we have a support for this button in the components and for now lets use a Row with Gesture enabled rather then coding a component here

mobile/apps/photos/lib/ui/components/delete_bottom_sheet.dart
AmanRajSinghMourya, line 1: We already have a bottom sheet components in the ente_components package so lets use that instead of creating this

mobile/apps/photos/lib/ui/actions/file/file_actions.dart
```diff
@@ -59,73 +67,168 @@ Future<void> showSingleFileDeleteSheet(
...
+                    );
+                  } else if (deletePreferenceForFile == .DeleteFromBoth) {
+                    await deleteFilesFromEverywhere(context, [file]);
+                    showShortToast(context, l10n.movedToTrash);
+                  }
+                },
+              ),
+            ),
+          if (isBothLocalAndRemote)
+            SizedBox(
```
AmanRajSinghMourya: Lets make sure that when the text size increase we should handle that case, and in general the components should be capable enough to handle the text scaling

Issue #5428, opened 2025-03-24
Scan QR code camera is all blury


SwissOS:
### Description

I have a Nothing Phone 2 (Android 15), and when I try to scan a QR code with the camera, the image is all blury, and I can never scan the QR code.
As a workaround, I need to enter the secret code manually, which is a real pain.

Can you please fix the camera setting to use a higher resolution and make the image clear?

### Version

4.3.2

### What product are you using?

Ente Auth

### What platform are you using?

Mobile - Android

JanB97:
same problem with nothing phone 1 and Android 15

50P15:
I dont see a blurry image, but I cannot scan QR codes, or its very difficult (depending on the QR)
I tried "2fas" to discard its a problem with my phone/cam, but this app has no problems scanning any QR

EDIT: Also FreeOTP+ and Stratum doesnt have any problems, but Aegis and FreeOTP (not plus) do

Using crDroid 11.6 (Android 15)

ua741:
@50P15 @SwissOS @JanB97 

Can you please test the camera scanning on the latest nightly build
https://github.com/ente-io/nightly/releases/tag/auth-v4.4.24-beta

50P15:
Working fine and fast! Tried on three different services the stable version can't scan.

One of them was https://2fas.com/auth/check-token/

PR #11015, opened 2026-06-16
[mobile][auth] Replace QR scanner dependency

neeraj-pilot:
## Summary
- Replace `qr_code_scanner_plus` with an Ente-owned live QR scanner.
- Use CameraX + zxing-cpp on Android and AVFoundation on iOS.
- Keep gallery QR import on the existing `ente_qr` still-image decoder.

Addresses #5428


ua741:
@codex review

ua741 (approved):

PR #11026, opened 2026-06-17
[mob][photos] Add illustrated empty states

AmanRajSinghMourya:
## Summary
- Added illustrated empty-state artwork for album search, links, and trash.
- Added a shared EmptyStateComponent for these illustrated empty states.
- Replaced the local SoftIconButton wrapper with ente_components IconButtonComponent in remaining callsites.


mobile/apps/photos/lib/l10n/intl_en.arb
```diff
@@ -1294,6 +1295,7 @@
   "searchByExamples": "• Album names (e.g. \"Camera\")\n• Types of files (e.g. \"Videos\", \".gif\")\n• Years and months (e.g. \"2022\", \"January\")\n• Holidays (e.g. \"Christmas\")\n• Photo descriptions (e.g. “#fun”)",
   "youCanTrySearchingForADifferentQuery": "You can try searching for a different query.",
   "noResultsFound": "No results found",
```
ua741, line 1297: should we remove this old string to be consistant?

AmanRajSinghMourya: `noResultsFound` is still used across other search/no-result surfaces, so we should keep it.

ua741 (approved):

PR #11028, opened 2026-06-17
[mob][photos] Add received album filters

AmanRajSinghMourya:
## Summary
- Split Albums shared collections into Shared and Received filters.
- Kept the right-pinned more button with the fade overlay for hidden filters.

### DIFF
--- a/mobile/apps/photos/lib/l10n/intl_en.arb
+++ b/mobile/apps/photos/lib/l10n/intl_en.arb
@@ -1545,6 +1545,10 @@
   "@searchResultShared": {
     "description": "Search result subtitle label for shared-with-you results"
   },
+  "received": "Received",
+  "@received": {
+    "description": "Albums tab filter label for albums received from other people"
+  },
   "searchResultUploadedBy": "Uploaded by",
   "@searchResultUploadedBy": {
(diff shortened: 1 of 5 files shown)


mobile/apps/photos/lib/l10n/intl_en.arb
```diff
@@ -1541,9 +1541,10 @@
   "@searchResultPerson": {
     "description": "Search result subtitle label for person results"
   },
-  "searchResultShared": "Shared",
-  "@searchResultShared": {
-    "description": "Search result subtitle label for shared-with-you results"
+  "shared": "Shared",
```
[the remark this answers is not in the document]
AmanRajSinghMourya: Yes but this will get translated in the new translation and have removed redundant `share` keys

neeraj-pilot, line 1544: We should also probably add description for this. May be call it sharedAlbumsLabel and receivedAlbumsLabel?

AmanRajSinghMourya: Done added description and change key to `sharedAlbumsLabel` and `receivedAlbumsLabel`

ashilkn (approved):

PR #11055, opened 2026-06-18
[mob][photos] Fix failure to upload a photo to multiple albums

nuit-dhiver:
Remove GROUP BY on localID in files query so multiple pending entries for the same local file (one per collection) are returned — the uploader dedupes by localID in-memory and needs all pending collection rows. In collection actions, switch to an indexed loop and clear generatedID for collections after the first so each album insertion creates a distinct pending row (reusing generatedID would replace prior entries). Added explanatory comments for both changes.


mnvr (approved):
Thank you!

I've merged main into your branch and done a few tweaks, and also added a changes entry.

PR #11126, opened 2026-06-24
[mobile][photos] New cast UX

aspizu:
this adds a new column to the casting table to store the device's name, which is provided by the cast target client, it can be any string, older clients will continue to use the IP address instead of the name. newer clients can provide a name like "Desktop" or "Mobile"

casting via chromecast api is broken, the UX changes need to be tested and polished for this path, i have not done this because i couldn't get it to work.

casting via code is tested, multiple casts at the same time works.

there is one issue, the cast sheet is displayed after fetching active cast sessions, so there is no spinner displayed, its a delay before the sheet is displayed
Simulator Screenshot - iPhone 17e - 2026-06-25 at 13 31 25
Simulator Screenshot - iPhone 17e - 2026-06-25 at 13 31 34
Simulator Screenshot - iPhone 17e - 2026-06-25 at 13 31 22
Simulator Screenshot - iPhone 17e - 2026-06-25 at 13 31 07
Simulator Screenshot - iPhone 17e - 2026-06-25 at 13 31 03
Simulator Screenshot - iPhone 17e - 2026-06-25 at 13 30 49
Simulator Screenshot - iPhone 17e - 2026-06-25 at 13 30 33


neeraj-pilot:
@codex review

mobile/apps/photos/lib/ui/settings/cast/cast_settings_page.dart
```diff
@@ -1,118 +1,198 @@
...
+    return SettingsPageScaffold(
+      title: l10n.castSessions,
+      children: [
+        CastSessionsList(
+          showTitle: false,
+          fallback: Column(
+            mainAxisAlignment: MainAxisAlignment.center,
+            crossAxisAlignment: CrossAxisAlignment.center,
+            children: [
+              SizedBox(height: MediaQuery.sizeOf(context).height * 0.5 - 200),
```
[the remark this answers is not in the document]
aspizu: we are doing the same thing in a few other places, should be fixed across the codebase if really an issue

mobile/apps/photos/lib/ui/cast/cast.dart
```diff
@@ -0,0 +1,69 @@
...
+import "package:photos/theme/ente_theme.dart";
+import "package:photos/ui/cast/pair_with_auto.dart";
+import "package:photos/ui/cast/pair_with_code.dart";
+import "package:photos/ui/settings/cast/cast_settings_page.dart";
+import "package:photos/utils/dialog_util.dart";
+
+Future<void> showCastSheet(BuildContext context, Collection collection) async {
+  final l10n = AppLocalizations.of(context);
+  final textStyle = getEnteTextTheme(context);
+  await castService.closeActiveCasts();
```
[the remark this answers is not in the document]
aspizu: we need to remove the feature flag altogether right?

server/migrations/125_add_cast_device_name.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE casting ADD COLUMN device_name VARCHAR(512) DEFAULT NULL;
```
neeraj-pilot: Let's just use TEXT DEFAULT NULL; here. Currently we are not using VARCHAR anywhere yet, not strong reason to add it here

server/pkg/utils/ua/ua.go
```diff
@@ -0,0 +1,25 @@
...
+	if dderr != nil {
+		return "", dderr
+	}
+	info := dd.Parse(ua)
+	if info == nil {
+		return "", errors.New("failed to parse user agent")
+	}
+	if info.Type == "" {
+		return "", nil
+	}
```
neeraj-pilot, line 39: Do we want to show empty deviceName in this case of IP?

server/pkg/controller/cast/controller.go
```diff
@@ -29,7 +30,12 @@ func NewController(castRepo *castRepo.Repository,
 }
 
 func (c *Controller) RegisterDevice(ctx *gin.Context, request *cast.RegisterDeviceRequest) (string, error) {
-	return c.CastRepo.AddCode(ctx, request.PublicKey, network.GetClientIP(ctx))
+	ipAddress := network.GetClientIP(ctx)
+	deviceName, err := ua.GetDeviceType(ctx.GetHeader("User-Agent"))
+	if err != nil {
+		deviceName = ipAddress
```
neeraj-pilot, line 42: Let's log the error here as warning?

server/pkg/controller/cast/controller.go
```diff
@@ -29,7 +30,21 @@ func NewController(castRepo *castRepo.Repository,
 }
 
 func (c *Controller) RegisterDevice(ctx *gin.Context, request *cast.RegisterDeviceRequest) (string, error) {
-	return c.CastRepo.AddCode(ctx, request.PublicKey, network.GetClientIP(ctx))
+	ipAddress := network.GetClientIP(ctx)
+	userAgent := ctx.GetHeader("User-Agent")
+	if len(userAgent) > 5000 {
+		userAgent = userAgent[:5000]
+	}
```
ua741: As discussed, let's remove this?

ua741:
@codex review

ua741:
@codex review

mobile/apps/photos/lib/ui/cast/cast.dart
```diff
@@ -0,0 +1,70 @@
...
```
aspizu: [P1] This closes any live local cast as soon as the user opens the cast sheet. In single-cast mode the old flow asked before stopping, so now tapping Cast and then dismissing the sheet still drops the current cast/control path. Please defer closing active casts until the user confirms stopping or actually starts a replacement pair flow.

mobile/apps/photos/lib/ui/cast/cast.dart
```diff
@@ -0,0 +1,70 @@
...
+  final gw = CastGateway(NetworkClient.instance.enteDio);
+  final logger = Logger("showCastSheet");
+  late final List<CastInfo> sessions;
+  try {
+    sessions = await gw.getAllCastSessions();
```
aspizu: [P2] This fetch blocks the whole cast sheet even when `enableMultiCast` is false, but the only consumer of `sessions` is the hidden `CastSessionsList` below. If `/cast/device-info` fails or is not meant to be used in single-cast mode, users cannot reach pair-by-code or auto-pair. Please only fetch sessions when multi-cast is enabled, or let the pair actions render even if the sessions fetch fails.

aspizu (commented):
Found two mobile cast flow issues in the new sheet behavior.

ua741 (approved):

PR #11189, opened 2026-06-29
[mobile][photos] improved onboarding design

aspizu:
simulator_screenshot_59800C83-75D7-4E9E-B9D7-CF946265AAD5
simulator_screenshot_2AE0F697-4F56-4313-99A6-807E5CBD20D6
Simulator Screenshot - iPhone 17e - 2026-07-02 at 11 36 35
Simulator Screenshot - iPad (A16) - 2026-07-02 at 13 32 30


mobile/apps/photos/lib/ui/home/landing_page_widget.dart
```diff
@@ -228,76 +204,6 @@ class _LandingPageWidgetState extends State<LandingPageWidget> {
     );
   }
 
-  Widget _buildFeatureCarousel() {
-    final l10n = AppLocalizations.of(context);
-    final features = [
-      (
-        "assets/onboarding_1.png",
```
AmanRajSinghMourya, line 236: Lets delete these assets as well if they are not used anymore

aspizu:
@codex review

mobile/apps/photos/lib/ui/home/grant_permissions_widget.dart
```diff
@@ -317,23 +318,26 @@ class _GrantPermissionsWidgetState extends State<GrantPermissionsWidget> {
...
+                            context,
+                          ).grantGalleryPermissionTitle,
+                          textAlign: .center,
                           style: TextStyle(
-                            fontFamily: "Nunito",
-                            fontWeight: FontWeight.w900,
-                            fontSize: 32,
-                            letterSpacing: -1.4,
+                            fontWeight: .w800,
+                            fontFamily: TextStyles.outfitFontFamily,
```
AmanRajSinghMourya, line 369: Pass the font family package here

`const String fontPackage = 'ente_components';`

aspizu:
@codex review

aspizu:
@codex review

aspizu:
the centered layout for onboarding screen is gone now, its again like the old one, otherwise it is fine

ashil-pilot:
>Review complete. I reviewed [PR #11189](https://github.com/ente-io/ente/pull/11189) ("[mobile][photos] improved onboarding design") at extra-high effort: 10 finder angles produced ~60 candidates, deduplicated to 17, each checked by an independent verifier against the PR-head checkout, the Flutter framework source, and the rive package source, plus a final gap sweep. 15 findings survived (14 confirmed, 1 plausible); 2 were refuted and dropped.
>
>The ones I'd fix before merge:
>
>Landing page can break entirely in release builds. Two Flexible(child: SizedBox(...)) spacers are direct children of a ListView. In debug this only spams "Incorrect use of ParentDataWidget" errors, but in release the validity check is compiled out and the cast to FlexParentData throws a TypeError during layout on every frame — on the app's first-launch screen. Plain SizedBox fixes it.
The offline grant-permissions screen overflows on landscape phones. The title font is width * 0.125 with no clamp inside a non-scrollable Column; on an 812×375 phone the content needs ~419px in ~354px, clipping the Grant permission button with no way to scroll. Nothing locks orientation at PR head (an earlier commit in this branch added a portrait lock, a later one removed it).
Dark-mode users get a near-invisible description. The subtitle uses colorScheme.greenLight, which is #212121 in the dark scheme, painted on the theme-independent bright green background (main hard-coded pale green here).
Release self-hosters lose the custom-endpoint indicator — DeveloperSettingsWidget is newly gated behind kDebugMode, but it already self-gates on non-production endpoints and this was the only logged-out surface showing it.
The rest: unclamped width * 0.09 title on tablets (up to ~123px), the landing page's Rive FileLoader never disposed (the old dispose() was deleted and not replaced — the sibling widget disposes correctly), w900 silently rendering as 700 because ente_components bundles no Outfit-Black, textScaler.noScaling disabling accessibility scaling on the titles, a layout jump plus silently-swallowed failure state in both Rive builders, one pre-existing async-context hazard, and five cleanups (7th copy of the Rive boilerplate, bespoke duplicated text styles, dead dots_indicator dependency, ~160KB of orphaned PNGs still shipping, six dead carousel l10n keys).
>
>Notable refutations, so you don't chase them: the short-viewport button overflow doesn't happen on real geometries (fixed stack is only ~176dp and the old layout was worse), the FileLoader dispose race can't actually reach a use-after-free or setState-after-dispose, orphaned welcome keys in other locales self-heal via Crowdin sync, and the Dart dot-shorthand syntax is fine (SDK constraint is ≥3.10). The review worktree and temporary branch have been cleaned up.


@aspizu  Could you check if these comments are issues that needs to be fixed? This is a review from Claude.

aspizu (commented):
Review notes from the multi-agent pass. I found two small issues worth fixing before merge.

mobile/apps/photos/lib/ui/home/landing_page_widget.dart
```diff
@@ -25,179 +23,178 @@ import 'package:photos/ui/payment/subscription.dart';
...
 class _LandingPageWidgetState extends State<LandingPageWidget> {
-  static const _featureCount = 3;
-  static const _autoScrollInterval = Duration(seconds: 4);
-
-  int _currentPage = 0;
-  int _activeDotIndex = 0;
-  bool _autoScrollDisabled = false;
-  Timer? _autoScrollTimer;
-  late final PageController _pageController;
+  late final rive.FileLoader _onboardingAnimationLoader;
```
aspizu, line 36: **[P2] Dispose the Rive loader**

This state creates `_onboardingAnimationLoader` and passes it to `RiveWidgetBuilder`, but it never calls `dispose()`. `rive.FileLoader.dispose()` frees the loaded file, and the sibling permission screen already does this for its loader. Add a `dispose()` that calls `_onboardingAnimationLoader.dispose()` before `super.dispose()`.

mobile/apps/photos/lib/ui/home/grant_permissions_widget.dart
```diff
@@ -317,47 +351,54 @@ class _GrantPermissionsWidgetState extends State<GrantPermissionsWidget> {
...
-                        ),
-                        const SizedBox(height: 22),
+                        const SizedBox(height: 32),
+                        _buildPermissionsAnimation(context),
+                        const Flexible(child: SizedBox(height: 22)),
                         Text(
-                          AppLocalizations.of(context).welcome,
+                          AppLocalizations.of(
+                            context,
+                          ).grantGalleryPermissionTitle,
```
aspizu, line 364: **[P3] Add translations for this new title key**

This screen moved from the old translated `welcome` key to `grantGalleryPermissionTitle`, but the new key only exists in `intl_en.arb`. Other locales that had `welcome` translated now fall back to English on this first-run screen. Please add this key to the same locale files, or keep using an already translated key until translations land.

aspizu:
@codex review

ashil-pilot:
> The offline grant-permissions screen still clips at large accessibility text sizes. The font cap fixed width-driven blowup, but the layout is still a non-scrollable Column while the description, button label, and terms all honor system text scaling — at 1.5–2× scale it overflows by up to ~175px with no scroll fallback. The online variant in the same file already has the right structure (CustomScrollView + SliverFillRemaining).

@aspizu

ashilkn (dismissed):

ashilkn (approved):

PR #11235, opened 2026-07-02
[mobile][auth] Add opt-in menubar mode for macOS

ivnnv:
## Description

Adds a "Menubar mode" toggle under Settings > General (macOS only, off by default so theres no behaviour change unless you enable it). When enabled the app starts hidden without a dock icon and lives in the menubar: left click on the tray icon toggles a small popover (380x600) anchored under the icon, clicking anywhere outside hides it. Right click keeps the context menu, where Show Window now opens a regular centered window in case you want the full sized app one-off, hiding it goes back to popover behaviour.

Some notes:
- the only pubspec addition is screen_retriever, which already ships as a window_manager dependency, used to open the popover on the display where the icon was clicked (macOS mirrors the status item on every menubar but the reported tray bounds are static, so the target display comes from the cursor position)
- the popover only hides on blur when the app itself loses focus, so file pickers and dialogs opened from the popover dont dismiss it
- in menubar mode the UI renders at a slightly larger logical size scaled down, so screens designed for the full window fit the popover without per page changes
- the popover size is never persisted, normal windowed mode keeps restoring the previous window size
- toggling the setting needs an app restart to take effect

## Tests

Tested manually on macOS 15 (Apple Silicon). Popover anchors under the icon and clamps at screen edges, toggles on icon click and hides on outside click. File import via the picker works from the popover. The one-off window opens centered and resizable, stays on the current Space only, and reverts to popover behaviour on hide. Codes are fresh after re-show, and with the setting off the app behaves exactly as before. Tested on a 3 display setup: the popover opens under the menubar of the display where you click the icon.

## Screenshot

Screenshot 2026-08-01 at 09 38 02


mobile/apps/auth/lib/services/window_listener_service.dart
```diff
@@ -71,20 +91,68 @@ class WindowListenerService with WindowListener, TrayListener {
...
+      final displays = await screenRetriever.getAllDisplays();
+      Rect? fallback;
+      for (final display in displays) {
+        final position = display.visiblePosition;
+        final size = display.visibleSize ?? display.size;
+        if (position == null) continue;
+        final bounds =
+            Rect.fromLTWH(position.dx, position.dy, size.width, size.height);
+        fallback ??= bounds;
+        if (bounds.contains(point)) return bounds;
```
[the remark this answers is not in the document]
ivnnv: yeah valid on that revision, this got reworked in dc23cd1: the display match is now done on the x axis only (cursor x against each display x range), so the visible-frame y exclusion doesnt apply anymore. tested on a 3 display setup, the popover opens under the icon you actually clicked

mobile/apps/auth/lib/services/window_listener_service.dart
```diff
@@ -31,7 +37,19 @@ class WindowListenerService with WindowListener, TrayListener {
     _isListening = true;
   }
 
+  bool isMenubarMode() {
+    return Platform.isMacOS &&
+        (_preferences.getBool(PreferenceService.kMenubarMode) ?? false);
```
neeraj-pilot: **P2: Keep the selected window mode stable until restart**

The toggle persists this key immediately, so this method changes its answer in the current process even though the settings UI says a restart is required. Enabling from a normal window immediately reroutes tray, blur, and close handling into popover mode without running the startup-only title bar, size, always-on-top, and workspace setup; disabling from a popover leaves those native styles in place while routing events as a regular window. Cache the launch-time mode, or otherwise defer the runtime switch, and use that stable value for window behavior.

mobile/apps/auth/lib/services/window_listener_service.dart
```diff
@@ -71,20 +91,83 @@ class WindowListenerService with WindowListener, TrayListener {
...
+      final displays = await screenRetriever.getAllDisplays();
+      Rect? fallback;
+      for (final display in displays) {
+        final position = display.visiblePosition;
+        final size = display.visibleSize ?? display.size;
+        if (position == null) continue;
+        final bounds =
+            Rect.fromLTWH(position.dx, position.dy, size.width, size.height);
+        fallback ??= bounds;
+        if (x >= bounds.left && x < bounds.right) return bounds;
```
neeraj-pilot: **P2: Use a two-dimensional display hit test**

Matching only x cannot distinguish displays arranged above and below one another, since their horizontal spans overlap. A click on either menu bar resolves to whichever overlapping display comes first, so the popover can open on the wrong display. The match should use the cursor point and full display bounds in both axes; visible or work-area bounds also need care because a side-positioned Dock removes an x strip.

mobile/apps/auth/lib/services/window_listener_service.dart
ivnnv: good catch, fixed: the mode is now snapshotted at launch and the toggle only persists the pref, nothing reroutes until restart

mobile/apps/auth/lib/services/window_listener_service.dart
ivnnv: yeah, reworked to a 2d nearest-rect test on full cursor point. two notes: screen_retriever converts cursor y against the lowest display top instead of the primary one, so the y gets corrected from the display list first. and nearest-rect (instead of contains) absorbs the menubar/dock strips missing from the visible bounds without knowing their exact sizes

neeraj-pilot:
@codex review

neeraj-pilot:
@codex review this PR for correctness and UX regressions. Trace the changed behavior end-to-end through related existing code, platform lifecycle, and realistic setting combinations. Report only concrete PR-introduced issues with a clear repro.

ua741 (dismissed):

neeraj-pilot:
Could you add a changelog entry under `mobile/apps/auth/changes/`? Something like `macos-menu-bar-mode.md`:

- Added an optional menu bar mode on macOS for quick access to your 2FA codes (@ivnnv).

ivnnv:
done, added

ua741 (approved):

PR #11245, opened 2026-07-03
[mobile][photos] always show more options if the user has not set a preference

aspizu:
Signed-off-by: Priyanshu Dangare <priyanshu@ente.com>


ashilkn (approved):

aspizu (commented):
Reviewed the current head (`f0ab3b8`) with a multi-agent pass. I found no live bugs to comment on.

The earlier invalid-action concern appears fixed: the expanded three-button UI is now gated by `widget.isLocal && widget.isRemote`, while local-only and remote-only sheets force the single safe delete target regardless of any saved preference.

Verification note: `git diff --check` passed. I also tried the focused widget test, but a fresh PR worktree could not compile before test load because generated l10n / strings / Flutter Rust Bridge Dart files were missing, so I did not count that as a test pass.

PR #11337, opened 2026-07-08
[mobile][photos] Lossless rotation/flip of JPEG images

aspizu:
## Dependency bump

- Bumps the Dart `image` package from `4.5.4` to `4.7.0` in Photos and the shared mobile workspace pins.
- Bumps `pdf` from `3.11.3` to `3.12.0` so the shared workspace can use `image 4.7.0`.


aspizu:
@codex review

mobile/apps/photos/lib/utils/lossless_edits.dart
```diff
@@ -0,0 +1,158 @@
...
+  File? f;
+  try {
+    f = await getFile(src, isOrigin: true);
+    if (f == null) {
+      throw Exception("Failed to get file");
+    }
+    final bytes = await f.readAsBytes();
+    final exif = _isJpegFile(bytes)
+        ? img.decodeJpgExif(bytes) ?? img.ExifData()
+        : null;
```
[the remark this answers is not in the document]
aspizu: no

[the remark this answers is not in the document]
ashilkn: @aspizu Isn't this a valid issue?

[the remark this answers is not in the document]
aspizu: https://github.com/brendan-duncan/image/pull/762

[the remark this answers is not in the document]
ashilkn: @aspizu  Ente currently pins image: 4.5.4, which does not contain the fix.

mobile/apps/photos/lib/utils/lossless_edits.dart
```diff
@@ -0,0 +1,158 @@
...
+  }
+}
+
+void _applyTransformToExif(img.ExifData exif, LosslessEditTransform transform) {
+  var orientation = exif.imageIfd.orientation ?? 1;
+  if (orientation < 1 || orientation > 8) {
+    throw Exception("Unsupported EXIF orientation");
+  }
+
+  if (transform.flipX) orientation = _flipXOrientation[orientation];
```
aspizu, line 75: **MEDIUM** — These tables put each new edit on the wrong side of the old EXIF orientation. For an orientation-1 JPEG, `flipX` plus one clockwise turn reaches orientation 5 here, while ProImageEditor draws orientation 7; an orientation-6 JPEG with `flipX` also maps to 7 instead of 5. The saved photo is therefore mirrored on the wrong axis. Build the tag in display order (`edit × old orientation`) and test all eight starting orientations with each flip/turn mix.

```dart
if (transform.flipX) orientation = _flipXOrientation[orientation];
if (transform.flipY) orientation = _flipYOrientation[orientation];
for (...) {
  orientation = _rotateCwOrientation[orientation];
}
// └──────────────────────┬──────────────────────┘
//                        └─→ writes old × flip × turn, not turn × flip × old
```

aspizu (commented):

mobile/apps/photos/lib/utils/lossless_edits.dart
```diff
@@ -0,0 +1,158 @@
...
+    }
+    final bytes = await f.readAsBytes();
+    final exif = _isJpegFile(bytes)
+        ? img.decodeJpgExif(bytes) ?? img.ExifData()
+        : null;
+    if (exif == null) {
+      throw Exception("Failed to decode JPEG EXIF");
+    }
+    _applyTransformToExif(exif, transform);
+    final out = img.injectJpgExif(bytes, exif);
```
aspizu, line 57: **MEDIUM** — With the pinned `image` 4.7.0, this call drops every JPEG marker before an existing EXIF block. For a valid `SOI → APP0 → ICC APP2 → EXIF` file, it still returns bytes but removes APP0 and the ICC colour profile, so a rotate/flip-only save can show different colours. Use an injector or package version that keeps every non-EXIF marker, and add a fixture that checks those marker bytes stay unchanged.

```dart
final out = img.injectJpgExif(bytes, exif);
//          └──────────────┬──────────────┘
//                         └─→ 4.7.0 can drop APP0/ICC before EXIF
```

mobile/apps/photos/lib/utils/lossless_edits.dart
```diff
@@ -0,0 +1,158 @@
...
+      t.originalSize.width.isFinite &&
+      t.originalSize.height.isFinite &&
+      t.cropRect.left.isFinite &&
+      t.cropRect.top.isFinite &&
+      t.cropRect.right.isFinite &&
+      t.cropRect.bottom.isFinite &&
+      (t.angle != 0 || t.flipX || t.flipY) &&
+      t.isRectangularCropper &&
+      _isSameRect(t.cropRect, fullImageRect) &&
+      _isSameDouble(t.scaleUser, 1) &&
```
aspizu, line 101: **MEDIUM** — The shared `0.01` gap treats a real centred pinch such as `scaleUser == 1.005` as no zoom. If the user also turns or flips the photo, every other check can pass and the EXIF-only path drops the zoom; a 4000-pixel image that should be cropped to about 3980 pixels is saved at the full width. Use an exact or much tighter scale-only check, and test that a turn plus `scaleUser: 1.005` uses the rendered save path.

```dart
_isSameDouble(t.scaleUser, 1) &&
// └─────────────┬────────────┘
//               └─→ 1.005 is treated as 1
```

mobile/apps/photos/lib/utils/lossless_edits.dart
aspizu: No

mobile/apps/photos/lib/utils/lossless_edits.dart
aspizu: Ok?

ashilkn (approved):

ashilkn (approved):

PR #11340, opened 2026-07-08
Enable Vulkan GPU offload for Ensu desktop on Linux

bjorn:
## What

Enables the Vulkan GPU backend for the Ensu desktop app on Linux and defaults to offloading all model layers to the GPU.

- `rust/crates/ensu/Cargo.toml`: compile `llama-cpp-2` with the `vulkan` feature on `target_os = "linux"`. macOS keeps Metal, Android/others stay on CPU; Windows Vulkan is intentionally left out because it requires a `VULKAN_SDK`.
- `rust/apps/ensu/src-tauri/src/commands/llm.rs`: default `n_gpu_layers` to full offload when the web layer sends no value, tunable at runtime via the `ENSU_N_GPU_LAYERS` env var (`0` forces CPU-only). This mirrors the existing `default_threads()` pattern used for `n_threads`.
- CI: adds the Vulkan loader and `glslc` to every Linux job that compiles the ensu crate for the host: the desktop bundle build (via the LunarG repo, since ubuntu-22.04 has no `glslc` package; same approach as llama.cpp's own CI), `rust-lint`, `warm-caches`, and the Android jobs (their `cargo codegen native` builds the uniffi library for the Linux host).

## Why

Benchmarking on a Radeon RX 480 (Gemma 4 E4B, Q4_K_M, ctx 12032) showed roughly **3x generation throughput vs CPU** (~31 tok/s vs ~11 tok/s).

Full offload also turned out to be the *right* default, not just the fast one:
- Partial offload pays a per-token CPU↔GPU activation-transfer cost and was up to ~1.6x slower than full offload in the sweep.
- Certain partial CPU/GPU split configurations abort with a ggml scheduler assert (`GGML_SCHED_MAX_SPLIT_INPUTS`); keeping the whole graph on the GPU avoids that.
- The driver spills any VRAM overflow to system memory (GTT) rather than failing, so full offload is safe even when the model exceeds VRAM.

## Notes

- The shipped Linux binary gains a runtime dependency on `libvulkan` (the loader). Machines without a Vulkan driver still run: llama.cpp falls back to CPU when no device is found.
- Local Linux builds now need the Vulkan loader dev package and `glslc` (Fedora: `vulkan-loader-devel glslc`; Debian/Ubuntu 24.04+: `libvulkan-dev glslc`).

## Testing

Built and ran the Ensu desktop app on Fedora (Radeon RX 480, RADV/Mesa) via `npm run dev`. The load log confirms `offloaded 43/43 layers to GPU` on `Vulkan0 (AMD Radeon RX 480 Graphics (RADV POLARIS10))`, and generation runs at the throughput above. With `ENSU_N_GPU_LAYERS=0` it falls back to CPU as before.

### DIFF
--- a/.github/workflows/ensu-build.yml
+++ b/.github/workflows/ensu-build.yml
@@ -141,6 +141,14 @@ jobs:
               run: |
                   sudo apt-get update
                   sudo apt-get install -y libwebkit2gtk-4.1-dev libsoup-3.0-dev libssl-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev
+                  # The llama.cpp Vulkan backend needs the Vulkan loader and the
+                  # glslc shader compiler at build time. Ubuntu 22.04 has no
+                  # glslc package, so use the LunarG SDK repository (the same
+                  # approach as llama.cpp's own CI).
+                  wget -qO- https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo tee /etc/apt/trusted.gpg.d/lunarg.asc >/dev/null
+                  sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-jammy.list https://packages.lunarg.com/vulkan/lunarg-vulkan-jammy.list
+                  sudo apt-get update
+                  sudo apt-get install -y vulkan-sdk
 
             - name: Install macOS Rust targets
               if: matrix.platform.rust-target == 'universal-apple-darwin'
@@ -322,6 +330,13 @@ jobs:
               run: |
                   echo "y" | sdkmanager --install "ndk;${{ env.NDK_VERSION }}"
                   echo "ANDROID_NDK_HOME=$ANDROID_HOME/ndk/${{ env.NDK_VERSION }}" >> $GITHUB_ENV
+
+            # cargo codegen native builds the uniffi library for the host, which
+            # on Linux enables the llama.cpp Vulkan backend.
+            - name: Install Vulkan build dependencies
+              run: |
+                  sudo apt-get update
+                  sudo apt-get install -y libvulkan-dev glslc
 
             - run: cargo codegen native
               working-directory: rust
--- a/.github/workflows/warm-caches.yml
+++ b/.github/workflows/warm-caches.yml
@@ -148,8 +148,6 @@ jobs:
                   workspaces: rust
                   shared-key: ensu-android
 
-            # cargo codegen native builds the uniffi library for the host, which
-            # on Linux enables the llama.cpp Vulkan backend.
             - name: Install Vulkan build dependencies
               run: |
                   sudo apt-get update
--- a/.github/workflows/warm-caches.yml
+++ b/.github/workflows/warm-caches.yml
@@ -37,7 +37,7 @@ jobs:
             - name: Install Linux dependencies
               run: |
                   sudo apt-get update
-                  sudo apt-get install -y libwebkit2gtk-4.1-dev libssl-dev libayatana-appindicator3-dev librsvg2-dev
+                  sudo apt-get install -y libwebkit2gtk-4.1-dev libssl-dev libayatana-appindicator3-dev librsvg2-dev libvulkan-dev glslc
 
             - name: Setup Rust cache
               uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
--- a/.github/workflows/warm-caches.yml
+++ b/.github/workflows/warm-caches.yml
@@ -148,6 +148,13 @@ jobs:
                   workspaces: rust
                   shared-key: ensu-android
 
+            # cargo codegen native builds the uniffi library for the host, which
+            # on Linux enables the llama.cpp Vulkan backend.
+            - name: Install Vulkan build dependencies
+              run: |
+                  sudo apt-get update
+                  sudo apt-get install -y libvulkan-dev glslc
+
             - run: cargo codegen native
               working-directory: rust
 
--- a/rust/apps/ensu/changes/vulkan-gpu-linux.md
+++ b/rust/apps/ensu/changes/vulkan-gpu-linux.md
@@ -1,0 +1,1 @@
+- Chat responses are now GPU-accelerated on Linux via Vulkan, roughly 3x faster. Thanks @bjorn!
--- a/rust/apps/ensu/src-tauri/src/commands/llm.rs
+++ b/rust/apps/ensu/src-tauri/src/commands/llm.rs
@@ -400,6 +400,31 @@ pub async fn llm_load_model(
         "LLM",
         format!("load model requested model_path={}", params.model_path),
     );
+    // GPU offload selection, in order of precedence: an explicit value from the
+    // web layer > the ENSU_N_GPU_LAYERS env override > the default. Benchmarking
+    // on the Vulkan backend (RX 480, Gemma 4 E4B) showed full offload is both
+    // fastest (no per-token CPU<->GPU activation transfers; partial offload was
+    // up to ~1.6x slower) and most robust (it sidesteps a ggml scheduler assert
+    // that can abort partial CPU/GPU splits). The driver spills any VRAM overflow
+    // to system memory instead of failing, so "offload everything" is a safe
+    // default. Set ENSU_N_GPU_LAYERS to cap it on a small GPU, or to 0 for CPU.
+    const DEFAULT_GPU_LAYERS: i32 = 999; // llama.cpp clamps to the model's layer count
+    let mut params = params;
+    if params.n_gpu_layers.is_none() {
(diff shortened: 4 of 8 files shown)


rust/crates/ensu/Cargo.toml
```diff
@@ -39,7 +39,18 @@ zeroize = { version = "1.8", features = ["derive"] }
 [target.'cfg(target_os = "macos")'.dependencies]
 llama-cpp-2 = { version = "=0.1.144", default-features = false, features = ["mtmd"] }
 
-[target.'cfg(not(target_os = "macos"))'.dependencies]
+# Desktop Linux enables the Vulkan GPU backend, which runs on AMD, Intel and
+# NVIDIA GPUs through the system Vulkan loader (no vendor SDK required). The
+# build needs a Vulkan loader plus "glslc" to compile the compute shaders. GPU
+# offload is opt-in at runtime via the ENSU_N_GPU_LAYERS env variable; with it
+# unset the app still runs entirely on the CPU.
+[target.'cfg(target_os = "linux")'.dependencies]
```
mnvr, line 45: > "GPU
> offload is opt-in at runtime via the ENSU_N_GPU_LAYERS env variable; with it
> unset the app still runs entirely on the CPU." 

This seems stale. Also, in light of my other comment, maybe we can just delete this sentence entirely.

rust/apps/ensu/src-tauri/src/commands/llm.rs
```diff
@@ -400,6 +400,31 @@ pub async fn llm_load_model(
...
+    // web layer > the ENSU_N_GPU_LAYERS env override > the default. Benchmarking
+    // on the Vulkan backend (RX 480, Gemma 4 E4B) showed full offload is both
+    // fastest (no per-token CPU<->GPU activation transfers; partial offload was
+    // up to ~1.6x slower) and most robust (it sidesteps a ggml scheduler assert
+    // that can abort partial CPU/GPU splits). The driver spills any VRAM overflow
+    // to system memory instead of failing, so "offload everything" is a safe
+    // default. Set ENSU_N_GPU_LAYERS to cap it on a small GPU, or to 0 for CPU.
+    const DEFAULT_GPU_LAYERS: i32 = 999; // llama.cpp clamps to the model's layer count
+    let mut params = params;
+    if params.n_gpu_layers.is_none() {
```
mnvr, line 415: Let us entirely drop the "ENSU_N_GPU_LAYERS" env knob. for the rare cases, llama/Vulkan already provides knobs for this I think (GGML_VK_VISIBLE_DEVICES, VK_LOADER_DRIVERS_DISABLE=*)

.github/workflows/warm-caches.yml
```diff
@@ -37,7 +37,7 @@ jobs:
             - name: Install Linux dependencies
               run: |
                   sudo apt-get update
-                  sudo apt-get install -y libwebkit2gtk-4.1-dev libssl-dev libayatana-appindicator3-dev librsvg2-dev
+                  sudo apt-get install -y libwebkit2gtk-4.1-dev libssl-dev libayatana-appindicator3-dev librsvg2-dev libvulkan-dev glslc
```
mnvr, line 40: We also need to prime the ensu-android job in this workflow.

.github/workflows/ensu-build.yml
```diff
@@ -141,6 +141,14 @@ jobs:
...
                   sudo apt-get update
                   sudo apt-get install -y libwebkit2gtk-4.1-dev libsoup-3.0-dev libssl-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev
+                  # The llama.cpp Vulkan backend needs the Vulkan loader and the
+                  # glslc shader compiler at build time. Ubuntu 22.04 has no
+                  # glslc package, so use the LunarG SDK repository (the same
+                  # approach as llama.cpp's own CI).
+                  wget -qO- https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo tee /etc/apt/trusted.gpg.d/lunarg.asc >/dev/null
+                  sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-jammy.list https://packages.lunarg.com/vulkan/lunarg-vulkan-jammy.list
+                  sudo apt-get update
+                  sudo apt-get install -y vulkan-sdk
```
mnvr, line 151: The shipped Linux packages don't account for the new libvulkan.so.1 link dependency and would fail to load on distros without Vulkan. I think tauri.conf.json will need changes to make sure the final rpm/deb/appimage artifacts declare a dep on Vulkan? (I'm not sure of all this, but I do think we should test/investigate)

mnvr (changes requested):
Thank you for the solid PR!

mnvr:
Also, a small thing, please also add a `changes/` entry @ mentioning yourself :)  (`rust/apps/ensu/changes`)

rust/crates/ensu/Cargo.toml
bjorn: Removed the stale sentence in 03ce0d6 (the env knob it referred to is gone entirely).

rust/apps/ensu/src-tauri/src/commands/llm.rs
bjorn: Dropped in 03ce0d6. Full offload is now the unconditional default whenever the caller does not specify a count, and the comment points to the existing Vulkan/llama.cpp escape hatches (GGML_VK_VISIBLE_DEVICES, VK_LOADER_DRIVERS_DISABLE) for the rare opt-out cases.

.github/workflows/warm-caches.yml
bjorn: Good catch, added the same Vulkan install step to the ensu-android cache job in 03ce0d6.

.github/workflows/ensu-build.yml
bjorn: Investigated, and it splits by artifact type (addressed in 03ce0d6):

- **deb/rpm**: these do rely on the system loader, so the bundle config now declares the dependency: `libvulkan1` (deb) and `vulkan-loader` (rpm) under `bundle.linux` in tauri.conf.json.
- **AppImage**: `libvulkan.so.1` is *not* on linuxdeploy's excludelist (unlike libGL/libEGL), so the loader gets bundled into the AppImage and it stays self-contained even on distros without Vulkan installed. At runtime, llama.cpp falls back to CPU when no Vulkan device/ICD is present, so Vulkan-less machines keep working.

Worth confirming on the first nightly artifacts: `dpkg-deb -I` for the declared dependency, and for the AppImage extract + `ldd` to check the bundled loader. I was not able to produce release bundles locally to fully verify the AppImage path, so treat that leg as needing one nightly-cycle confirmation.

bjorn:
Added the release notes entry (rust/apps/ensu/changes/vulkan-gpu-linux.md) as part of 03ce0d6. All review points should be addressed now.

bjorn:
@mnvr Thanks for the thorough feedback!

Full disclosure: I was excited to try Ensu on my system, but it was way too slow for me to be usable and I noticed it was CPU-only. That's why I asked Claude whether it would be possible to use the GPU, and it did the full patch and testing from there. The x3 speed bump even on my old RX 480 is a very welcome one, and I bet people will more recent GPUs will benefit a lot more.

Let me know if you have any policy on AI, for example attributing the commits with Co-authored-by.

rust/apps/ensu/changes/vulkan-gpu-linux.md
```diff
@@ -0,0 +1 @@
+- Chat responses are now GPU-accelerated on Linux via Vulkan, roughly 3x faster. Thanks @bjorn!
```
bjorn, line 1:
```suggestion
- Chat responses are now GPU-accelerated on Linux via Vulkan. Thanks @bjorn!
```
I'm not sure if that thanks is appropriate here and the x3 speedup will of course very much depend on the CPU and GPU that are being compared, so I'd suggest keeping it simple.

@mnvr Edit: Ok, keeping the thanks because I see this was suggested by you, hah!

mnvr (approved):

rust/apps/ensu/changes/vulkan-gpu-linux.md
bjorn: Change was merged without applying this suggestion. You may want to change it in a follow-up commit?

mnvr: I didn't make this suggestion? Or at least, not intentionally :) Either ways, I think the one that got committed is fine as it is, so all good.

bjorn: No, I made the suggestion, because the x3 speedup was just calculated on my system, and does not apply to any other combination of CPU and GPU. One could make it more specific, like "3x faster on Ryzen 9 3950X + RX 480", or remove it. But if you're fine with the text as-is we can also leave it of course.

mnvr: Done! https://github.com/ente/ente/pull/11358

PR #11357, opened 2026-07-09
[mobile][photos] Add string response breadcrumbs

neeraj-pilot:
- Add metadata-only breadcrumbs when API responses decode as non-empty strings.
- This helps identify whether malformed responses are coming from redirects, proxies, or edge/self-hosted behavior.


neeraj-pilot:
@codex review

ua741 (approved):

PR #11368, opened 2026-07-09
[mobile] add swipe page transitions to Android

aspizu:


aspizu:
@codex review

mobile/apps/photos/lib/ui/settings/inherited_settings_state.dart
```diff
@@ -24,13 +26,47 @@ class _SettingsState extends State<SettingsStateContainer> {
...
   @override
   Widget build(BuildContext context) {
     return InheritedSettingsState(
       expandedSectionCount: _expandedSectionCount,
       increment: increment,
       decrement: decrement,
-      child: widget.child,
+      pushPage: pushPage,
+      child: AnimatedSlide(
+        offset: _isSubpageOpen ? const Offset(-1.0 / 3.0, 0) : Offset.zero,
```
aspizu, line 65: **MEDIUM** — When a top-level settings route is popped, `await routeToPage` does not finish until the pop animation ends, so `_isSubpageOpen` stays true and this backdrop remains at -1/3 throughout a system-back animation or edge swipe. The revealed settings page cannot track the gesture and then runs a separate 300 ms slide after the route is gone, which makes the return visibly happen in two steps and can expose a blank strip near the end of the swipe. Drive this offset from the route's secondary animation (for example by exposing the Cupertino delegated transition) and add a widget test that pauses an edge swipe halfway and checks the backdrop position.

aspizu (commented):

mobile/apps/photos/lib/ui/settings/inherited_settings_state.dart
aspizu: dont care

mobile/apps/photos/lib/services/magic_cache_service.dart
```diff
@@ -219,6 +219,7 @@ GenericSearchResult? toGenericSearchResult(
             matchedUploadedIDs: filesToUploadedFileIDs(enteFilesInMagicCache),
           ),
         ),
+        useFadeTransition: true,
```
ashilkn: What is the reason we're using fade transition here? It has been added in other galleries as well.

ashilkn:
Add an entry to change log so that internal testers are aware of this change.

ashil-pilot:
#### Code review by claude (fable)
>- Unintended blast radius. The new Android edge-swipe gesture steals left-edge drags from MapScreen and the panorama viewer (pops the screen mid-pan); 13 call sites newly forced to fade lose the swipe-back gesture iOS users already had; and the shared-package change flips Android navigation in locker/auth/lock_screen/legacy, which got no compensating theme.
>- Mixed-stack transitions don't coordinate. The canTransitionTo allowlist plus SwipeableRouteBuilder's missing delegatedTransition mean any Material↔swipe stack shows a frozen page instead of parallax, in both directions.
>- The settings parallax is a convention, not a mechanism. It freezes during interactive back-swipes, desyncs on release, and is already bypassed by the storage card, Family plans, and settings-search flows.
 
Once these are fixed, we should also make sure there are no regressions on other apps (Auth, Locker) too.

PR #11378, opened 2026-07-10
[desktop][ensu] Reuse the llama.cpp KV cache across chat turns

bjorn:
Multi-turn chat previously cleared the llama.cpp KV cache and re-decoded the entire conversation on every send, so prompt-processing cost (and the GPU stalls it causes) grew with conversation length on each turn.

This tracks exactly which tokens are in the KV cache, during prefill and as generated tokens are accepted, and on each generation decodes only the divergent tail after the longest common token prefix with the cache (evicting just the diverged suffix via `llama_memory_seq_rm`). Time-to-first-token on multi-turn conversations goes from scaling with history length to roughly constant, since only the new turn's tokens are decoded.

- An empty or diverged cache degrades gracefully to a full decode.
- Multimodal requests keep the clear-and-full-decode path; no prefix matching is attempted across image embeddings.
- For sliding-window-attention models the rollback is limited to what SWA can safely restore.
- With greedy sampling the output is byte-identical whether served from a cold cache or a reused prefix, covered by an ignored integration test (`kv_reuse_equivalence`) that runs the real model.

A second commit exposes `n_ubatch` in `ContextParams` so callers can shrink the micro-batch, which shortens the individual GPU dispatches issued during prompt processing and keeps the desktop responsive on GPUs without compute preemption. Opt-in: when the field is absent or null the llama.cpp default stays in effect.

Only the `ente-ensu` crate's `llm` module is touched (`context.rs`, `event.rs`, `generate.rs`).

### DIFF
--- a/rust/crates/ensu/src/llm/context.rs
+++ b/rust/crates/ensu/src/llm/context.rs
@@ -2,7 +2,8 @@ use llama_cpp_2::context::LlamaContext;
 use llama_cpp_2::context::params::LlamaContextParams;
 use llama_cpp_2::model::LlamaModel;
 use llama_cpp_2::mtmd::{MtmdContext, MtmdContextParams, mtmd_default_marker};
-use parking_lot::Mutex;
+use llama_cpp_2::token::LlamaToken;
+use parking_lot::{Mutex, MutexGuard};
 use self_cell::self_cell;
 use serde::{Deserialize, Serialize};
 use std::ffi::CString;
@@ -46,6 +47,15 @@ struct CachedMtmdContext {
 pub struct Context {
     cell: Mutex<ContextCell>,
     mtmd_context: Mutex<Option<CachedMtmdContext>>,
+    /// Tokens currently held in the KV cache, in order. Kept in sync with
+    /// what has actually been decoded so the next generation can reuse the
+    /// longest common prefix instead of re-decoding the whole prompt.
+    cached_tokens: Mutex<Vec<LlamaToken>>,
+    /// Whether the model uses sliding-window attention for any layer.
+    /// SWA KV cells older than the window are overwritten as decoding
+    /// advances, so prefix reuse must not roll back more than one position
+    /// (see `swa_safe_prefix_len` in generate.rs).
+    swa: bool,
 }
 
 pub type ContextRef = Arc<Context>;
@@ -56,12 +66,28 @@ unsafe impl Sync for Context {}
 impl Context {
     fn try_new(
         owner: ModelRef,
+        swa: bool,
         builder: impl for<'a> FnOnce(&'a ModelRef) -> Result<LlamaContext<'a>, Error>,
     ) -> Result<Self, Error> {
         ContextCell::try_new(owner, builder).map(|cell| Context {
             cell: Mutex::new(cell),
             mtmd_context: Mutex::new(None),
+            cached_tokens: Mutex::new(Vec::new()),
+            swa,
         })
+    }
+
+    /// Bookkeeping for the tokens currently decoded into the KV cache.
+    ///
+    /// Lock ordering: when both are needed, `cell` (via
+    /// [`Self::with_context_mut`]) is locked before `cached_tokens`.
+    pub(super) fn cached_tokens(&self) -> MutexGuard<'_, Vec<LlamaToken>> {
+        self.cached_tokens.lock()
+    }
+
+    /// Whether the model behind this context uses sliding-window attention.
+    pub(super) fn uses_swa(&self) -> bool {
+        self.swa
     }
 
     pub(super) fn with_context_mut<R>(
@@ -169,7 +195,8 @@ impl Context {
             context_params = context_params.with_n_batch(n_batch);
         }
 
-        let context = Context::try_new(Arc::clone(model), |model| {
+        let swa = model_uses_swa(model.model());
+        let context = Context::try_new(Arc::clone(model), swa, |model| {
             let backend = backend()?;
             model
                 .model()
@@ -193,5 +220,22 @@ impl Context {
             self.cached_mtmd_context(ctx.model, &mmproj_path, &marker)
                 .map(|_| ())
         })
+    }
+}
+
+/// Whether the model uses sliding-window attention (SWA) for any layer.
+///
+/// Detected from GGUF metadata (`<arch>.attention.sliding_window`), because
+/// llama-cpp-2 exposes neither `llama_model_n_swa` nor
+/// `llama_memory_seq_pos_min`. A missing key means no SWA; unreadable or
+/// unparseable metadata is treated as SWA, which only costs prefix reuse on
+/// rollbacks, never correctness.
+fn model_uses_swa(model: &LlamaModel) -> bool {
+    let Ok(arch) = model.meta_val_str("general.architecture") else {
+        return true;
+    };
+    match model.meta_val_str(&format!("{arch}.attention.sliding_window")) {
+        Ok(value) => value.trim().parse::<i64>().map(|n| n > 0).unwrap_or(true),
+        Err(_) => false,
     }
 }
--- a/rust/crates/ensu/src/llm/generate.rs
+++ b/rust/crates/ensu/src/llm/generate.rs
@@ -316,6 +316,123 @@ impl StreamDecoder {
     }
 }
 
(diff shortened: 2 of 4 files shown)


bjorn:
I noticed that as the chat grew in length, there was an ever increasing pause after each of my messages, so I asked Claude what was causing this and he pointed out the cache was being cleared every time using `ctx.clear_kv_cache()` and that because of this, the entire chat history needed to be processed again before a response could be generated.

So this patch mostly tried to avoid that call, and instead replaces it with a call to `ctx.clear_kv_cache_seq` which tries to retain as much as possible of the cache in memory. Often, the entire cache can be kept which means responses can start generating immediately.

I'm not sure if there might be a better way to do this or a more focused patch possible. It also needs further testing, I might have opened this PR too early.

rust/crates/ensu/src/llm/context.rs
```diff
@@ -18,6 +19,7 @@ pub struct ContextParams {
     pub context_size: Option<i32>,
     pub n_threads: Option<i32>,
     pub n_batch: Option<i32>,
+    pub n_ubatch: Option<i32>,
```
bjorn: I'll take out the addition of `n_ubatch` since it's independent of the cache reuse.

rust/crates/ensu/src/llm/generate.rs
```diff
@@ -467,6 +596,115 @@ impl Context {
...
+    /// cached prompt is a no-op. Decodes serialize on the context's internal
+    /// lock, so this never interleaves with a generation; `cancel_flag` is
+    /// checked between chunks so a pending generation waits at most one
+    /// chunk after cancelling. When something else owns the context right
+    /// now (a generation is running), the warm-up is skipped instead of
+    /// queued: whatever it would have decoded is stale by the time the
+    /// owner finishes. Returns the number of prompt tokens actually
+    /// decoded; `Error::Cancelled` when the flag was set or the context was
+    /// busy.
+    pub fn prefill_chat(
```
bjorn: This function is currently unused by the Ensu React app, I'll take it out of the PR.

bjorn:
> It also needs further testing, I might have opened this PR too early.

Ok this whole cache reuse approach broke down due to the Ensu React app putting in a fresh timestamp in the prefix every time. I think we could solve this by leaving the timestamp at the time the chat was started.

mnvr:
I think you're using Claude here, and Claude has a tendency to add slop comments and unnecessary abstractions. Also, now this is a more general change with unknown performance impact for edge cases (there must be reasons why llama is not doing by default what we're asking it to do). 

So we should scope our changes, and make them in a more surgical manner so that they can be reviewed in a manner that their impact is statically interpretable.

(I guess since Claude is handling your PR, Claude will be reading this anyway :) )

bjorn:
> I think you're using Claude here, and Claude has a tendency to add slop comments and unnecessary abstractions.

Certainly, I often ask it to remove its slop comments. There are a bunch of verbose comments here too, but I didn't ask to remove them because I thought they might be helpful in explaining what Claude is trying to do here.

> (there must be reasons why llama is not doing by default what we're asking it to do).

I did test the patch, btw, and it does have the desired result of answers coming in immediately. However, thinking about it, I now wonder whether maybe the only thing we actually needed was to change that date+time string that is prepended to the entire conversion. Because considering what you said, I think it's actually likely that llama.cpp already does exactly this type of optimization. But prepending a different string to the conversation every time we send another message would break that optimization down entirely.

By changing the start of the conversation essentially every second, llama.cpp always ends up needing to process the entire conversation for each message.

So I might have asked Claude to implement an optimization that's already there. I will test this.

> (I guess since Claude is handling your PR, Claude will be reading this anyway :) )

Fortunately I'm not pure robot. :-)

bjorn:
Hmm, unfortunately Claude believes the change can't be any more surgical, because both parts of this change are necessary for the optimization to work. It's verbose but it seems to me good explanation:

> **1. The old Ensu engine explicitly wiped the cache on every single request.** In upstream main (pre-PR) generate.rs, there are two unconditional ctx.clear_kv_cache() calls (lines 597 and 712), at the start of each generation, and zero prefix-reuse logic (0 matches for common_prefix_len / apply_prefix_reuse / reusable_prefix). So the old flow was, every turn: clear the whole KV cache → re-decode the entire history from scratch. When you throw the cache away each time, a stable prefix has nothing to reuse — the date being constant or changing is irrelevant, because there's no surviving cache to match against. The date fix on top of the old engine would have changed nothing about the pause.
> 
> **2. The reuse is application logic, not a library freebie, and Ensu uses the library, not the server.** This is the crux of your hypothesis, and it hinges on a distinction:
> - llama-server is a program. It implements prompt/prefix reuse in its own server code, that's what the --cache-reuse flag, the "prompt cache," and the slots in your startup log are. Those are all server-application features.
> - libllama (which Ensu links via the llama-cpp-2 crate) is the library. It gives you primitives: the KV cache persists across llama_decode calls, and you get clear_kv_cache, llama_memory_seq_rm, and decode. It does not automatically detect a common prefix and skip it. If it did, the old code's manual clear_kv_cache() would be the only thing standing between you and free reuse, and simply deleting that line would suffice.
>
> But deleting the clear alone would not work, it would produce wrong output, not free speedup: if you stop clearing but still feed llama_decode the full prompt at positions 0..N over a cache that already holds 0..K, you write duplicate/conflicting KV entries. To reuse correctly you must compute the common prefix, evict the divergent tail with seq_rm, and decode only the new tokens at the right positions. That's precisely the ~120 lines our change adds. llama-server contains its own copy of that logic; Ensu, embedding the library directly, did not, until we added it.
>
> **So the two fixes are complementary, not substitutes:**
> - Rust change → enables reuse (stop clearing, implement prefix reuse). Without it: full re-decode every turn regardless of the date.
> - Date fix → makes that reuse actually apply. Without it: reuse is implemented but the prefix never matches, so it degrades to a near-full decode.
> - Neither alone gives the speedup.
>
> **The legitimate kernel in your hypothesis** (and I want to be honest about it): if Ensu had been architected to run llama-server as a subprocess and talk to it over HTTP, instead of embedding libllama, then yes, the server's built-in reuse would have done this for free and the date fix would have been the only thing needed. Your instinct is correct in that counterfactual. It's just that Ensu made the opposite architectural choice (embed the library), so it had to bring its own reuse, which is what we added.

Now of course it might be that the way this change was implemented is not the architecture you're after, and I don't expect either that you'd just merge this. I do hope that it will make you aware of the performance issue (which on my RX 480 is impossible to miss, it's another one of those issues that make this app entirely unusable like #11340).

In the meantime, I have simply switched to running `llama-server` with the Gemma 4 model and using its web interface. It performs a lot better (it uses Vulkan, reusing prefix, its web interface is fast, it can fork chats). Another issue in Ensu is that it uses React in a way that makes it almost impossible to type in the input field - I might still open a PR for that, but of course Claude wrote it...

mnvr:
Just so that we're clear, the issue is not with using Claude. The last PR was done by Claude too, and I liked it! The issue is that just by looking at this PR, I cannot judge easily if it is safe to merge in a way that doesn't regress for someone (macOS, Android, iOS, Windows, non-Vulkan etc). This doesn't imply the change is wrong, and with more time investment I can figure out the safety, but if I had more time... 

I think you're an open source maintainer too, I hope you understand what I'm driving at.

Anyways, give me some time, I will try to understand and dig deeper what the PR is trying to achieve.

bjorn:
> I think you're an open source maintainer too, I hope you understand what I'm driving at.

@mnvr Yes, I fully understand. There is no rush and I would even understand if this PR mostly just communicates the idea rather than the implementation, because sometimes it's easier to implement something yourself than to verify an implementation by somebody else. And indeed, with this implementation being by Claude, I also can't explain the trade-offs it made.

Being a maintainer is also how I got access to Claude (through https://claude.com/contact-sales/claude-for-oss). While those 6 months last, I don't have to worry much about whether a prompt will be worth its token cost, which is a rather odd luxury that also leads to contributions like this.

mnvr:
I've done https://github.com/ente/ente/pull/11577 - the first commit there is a narrower subset that address most of these same points. There are other improvements but I think all of them start to go into much more complex areas (the mmproj cache, the title/chat split, the sliding window).

I've also done one more change there - which uses a quantized history inclusion to avoid thrashing of the cache when the chat becomes very long.

Thanks for this PR!

PR #11386, opened 2026-07-10
[mobile][photos] fix loading state black screen flash in video player on Android

aspizu:


aspizu:
@codex review

aspizu:
(invalid, not my changes)

**[Medium] Keep the loading mask tied to the final source load**

`_setFilePathForNativePlayer(..., update: true)` calls `setVideoSource()` here, but both callers start it once more: `loadPreview` always does `await setVideoSource()`, and `loadOriginal` does the same when `update` is true. The first call is also not awaited. This gives one stream switch two untracked `loadVideo` calls. A ready event from the first call can set `_isPlaybackReady` to true and drop the new loading mask while the second call is still resetting the native view, so the black flash this change aims to fix can still show. Please make one layer own the reload and await it once (or tag each load and only accept the ready event for the newest one).

```dart
_isPlaybackReady.value = false;  // ┌─ show the mask
if (update) {
  setVideoSource();              // ├─ load #1, not awaited
}

await setVideoSource();          // └─ load #2 in the caller
```

aspizu:
@codex review

ashilkn:
Let's add a change log entry for this. Also mention this issue only existed for video stream playbacks.

ashilkn (approved):

PR #11398, opened 2026-07-10
[web][photos] Fix/viewer transparent background

tanmaywankar:
## Description

This PR implements a checkerboard background pattern so that the transparent images do not just look black.
- Added background for both dark and light mode image viewing.
- note that the images uploaded before this update would not automatically get the new background in thumbnails.

## Comparision
Before:
Screenshot 2026-07-09 180434

After:

https://github.com/user-attachments/assets/d1bc061d-05ec-42f2-8148-fadc16339811

## Tests

Verified that frontend changes are clean and stable:
- Ran `npm run lint` to ensure code formatting aligns perfectly with repository rules.
- Executed the test suite for the local frontend workspace:
  `npm run test --workspace=photos`


AswinAsok:
Hi @tanmaywankar,

I reviewed your code changes. Though they look functionally correct, I have concerns about the overhead. For desktop PNG/WebP uploads, thumbnail generation moves from the native Electron path to the renderer, where the full image is loaded and decoded before resizing. This could cause significant memory pressure for large or concurrent uploads.

The PhotoSwipe checkerboard is less concerning, although it is applied behind every image without transparency detection.

For thumbnails, the ideal solution would be to use a transparency-preserving format, but that is a product decision which we need to me. Therefore, I won’t be able to merge this PR at the moment.

Thank you for the effort you put into this.

PR #11427, opened 2026-07-11
[mobile][photos] Add persistent video mute control

r4khul:
## Description

Adds a persistent mute button to the video player. Mute state is saved across app restarts and syncs instantly to all preloaded videos via the event bus. Works on both the native and MediaKit players.

Volume is now set before playback begins in-order to fix the bug where the native player would briefly play audio on startup even when muted.

**Before:**

image

**After:**

image

## Tests

Tested and verified on devices Realme C67 5G and Realme Pad 2.

### DIFF
--- a/mobile/apps/photos/lib/events/video_mute_changed_event.dart
+++ b/mobile/apps/photos/lib/events/video_mute_changed_event.dart
@@ -1,0 +1,7 @@
+import "package:photos/events/event.dart";
+
+class VideoMuteChangedEvent extends Event {
+  final bool isMuted;
+
+  VideoMuteChangedEvent(this.isMuted);
+}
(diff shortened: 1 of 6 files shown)


ua741:
@r4khul Thank you for the PR. Can you please rebase and share update screenshot? Sorry for the delay in getting back to this.

r4khul:
@ua741 done!

Here is the updated scrn shot:

image

**Functionality works as intented.**

r4khul:
the lint check should pass now.

ashilkn:
@r4khul One issue here would be once videos are muted, there would be no way to unmute videos that appear on memories (memories appear on the top of the homepage) since the video players in memories don't show video controls.  We could use the existing `isFromMemories` flag passed to the video players to avoid applying/subscribing to the mute state on memories.

ashilkn:
Adding an unmute button to the memories video player would require some thought about where to place it, so best quick solution is to avoid respecting the mute state on memories.

r4khul:
@ashilkn thanks for the catch.
yeh using the flag sounds good to me too.
will update the code & ping in a while.

r4khul:
PTAL @ashilkn

ashilkn:
@codex Review

ashilkn (approved):

ashilkn:
Thanks for the contribution @r4khul :)

PR #11430, opened 2026-07-12
[mobile] Centralize upload queue transitions

neeraj-pilot:
## Summary

- centralize upload queue, session counters, and backup status transitions
- prevent stale completion and background-poll results from mutating a newer request for the same file
- publish immutable backup snapshots, batch cancellation updates, and keep one background poll loop

User impact: upload progress stays consistent during retries and foreground/background handoffs, with less backup-status UI churn and no completed-item buildup in the uploader.


ua741:
@codex review

ua741 (approved):

PR #11454, opened 2026-07-13
[mobile][photos] Set search debounce to 314ms

aspizu:
## What changed

- Set the Photos global search debounce from 200 ms to 314 ms.

## Why

This gives the user a slightly longer typing gap before a search starts.

## Impact

Global search now starts 314 ms after the last input change.

## Checks

- `git diff --check`


aspizu:
@codex review

ashilkn (approved):

PR #11481, opened 2026-07-15
[web][photos] Add pre-upload report screen - p1

AswinAsok:
These changes introduce a new dialog that appears just before the upload process starts, showing a minimal summary of the files and album about to be uploaded. If it's a Takeout import, the dialog also includes the relevant options for that.

This implementation works, but the current data flow feels a bit rough, especially where it handles the submission/closure of the SingleInputDialog.

Update

The older implementation has a promised based resolve rejection for the dialog which made it feel not part of the upload flow, that has been updated to a more native one. Now it's perfect.


AswinAsok:
@codex review

anandbaburajan (approved):

PR #11498, opened 2026-07-16
[mobile][photos] add trash api to media store and refactor a bit

aspizu:


mobile/apps/photos/android/app/src/main/kotlin/io/ente/photos/MediaStoreChannel.kt
```diff
@@ -0,0 +1,112 @@
...
+        val intent = Intent(Settings.ACTION_REQUEST_MANAGE_MEDIA).apply {
+            data = Uri.parse("package:${activity.packageName}")
+            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+        }
+        activity.startActivity(intent)
+    }
+
+    private fun restoreTrashedFiles(call: MethodCall, result: MethodChannel.Result) {
+        startMediaRequest(call, result) { uris ->
+            MediaStore.createTrashRequest(activity.contentResolver, uris, false)
```
aspizu, line 79: **MEDIUM** — These new helpers can be called on Android 8–10, which the app still supports, but `createTrashRequest` and `createDeleteRequest` only exist from Android 11. On those devices the call can end in `NoSuchMethodError` on the main thread. Guard both paths with `SDK_INT >= R`, return a clear unsupported error, and add an API 29 test.

mobile/apps/photos/android/app/src/main/kotlin/io/ente/photos/MediaStoreChannel.kt
```diff
@@ -0,0 +1,112 @@
...
+        result: MethodChannel.Result,
+        createRequest: (List<Uri>) -> android.app.PendingIntent
+    ) {
+        if (pendingResult != null) {
+            result.error("request_in_progress", "Another media request is in progress", null)
+            return
+        }
+
+        val uris = getUris(call)
+        val request = createRequest(uris)
```
aspizu, line 100: **MEDIUM** — The app targets SDK 36, where these MediaStore APIs reject a request with more than 2,000 URIs. Passing a large trash set as one list throws before the system prompt opens, so restore or permanent delete fails. Split the work into safe batches and await each prompt, then add a test with 2,001 URIs.

aspizu (commented):

PR #11515, opened 2026-07-16
[mobile] Enforce safe BuildContext use across async flows

AmanRajSinghMourya:
Enables `use_build_context_synchronously` as an error across mobile and updates affected account, legacy, lock-screen, and shared-package async flows. UI work is guarded by the relevant mounted context while non-UI operations continue when possible.

Validation: `flutter analyze --no-pub`


AmanRajSinghMourya:
@codex review

ua741 (approved):

PR #11534, opened 2026-07-17
[mobile] Fix text input tap grouping

aspizu:
## What changed

- Give the inner `TextField` the same tap-region group ID as its wrapper.
- Refresh the group ID when the disabled state changes.
- Make the stored group ID non-null after setup.

## Why

The wrapper used a custom group ID, but the inner field kept Flutter's default ID. This split one text input across two tap groups and broke the intended focus behavior around disabled fields and field affordances.

## Impact

Each text input now acts as one tap region, while enabled and disabled states keep separate IDs as intended.

## Checks

- `fvm flutter test test/components/text_input_component_test.dart` (19 tests passed)

### DIFF
--- a/mobile/packages/ente_components/lib/components/text_input_component.dart
+++ b/mobile/packages/ente_components/lib/components/text_input_component.dart
@@ -166,6 +166,9 @@ class _TextInputComponentState extends State<TextInputComponent> {
   @override
   void didUpdateWidget(covariant TextInputComponent oldWidget) {
     super.didUpdateWidget(oldWidget);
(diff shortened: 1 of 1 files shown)


ashilkn:
@aspizu Is this the fix for the issue that we noticed on the reset password screen? 
@AmanRajSinghMourya Please review this PR.

aspizu:
The regression is for example tapping on the eye icon in password field dismisses the keyboard.

AmanRajSinghMourya:
@ashilkn LGTM

AmanRajSinghMourya (approved):

PR #11537, opened 2026-07-17
[mob][locker] Remove default text input limit 

AmanRajSinghMourya:


mobile/apps/locker/lib/ui/components/text_input_sheet.dart
```diff
@@ -10,13 +10,10 @@ Future<dynamic> showTextInputSheet(
   required FutureVoidCallbackParamStr onSubmit,
   String? initialValue,
   TextCapitalization textCapitalization = TextCapitalization.words,
-  int? maxLength = 200,
```
ua741, line 13: Do we need to add this limit in any place where we are using this field?

AmanRajSinghMourya: The public-link password field has its own explicit 256-character limit. That is the only place with max char limit

ua741 (approved):

PR #11543, opened 2026-07-17
[mobile][web] Distinguish owned uncategorized collections

neeraj-pilot:
Treat only the current user's uncategorized collection as the special fallback, while preserving incoming collections owned by others. Sharing controls remain disabled.


neeraj-pilot:
@codex review

ua741 (approved):

ua741:
Got an offline ack from @AswinAsok

PR #11585, opened 2026-07-21
fix(desktop): avoid renderer OOM during folder-watch upload of large libraries

DxTa:
## Problem

Large desktop folder-watch uploads could exhaust renderer memory and leave gallery state inconsistent. 

## Fix

- Shard local persistence into validated manifests/chunks with zero-cursor recovery.
- Bound upload, duplicate lookup, watcher, and gallery memory.
- Use one bounded background gallery hydrator.
- Preserve upload retries, progress, deduplication, and watcher synchronization.
- Refresh gallery once after watcher uploads complete.

### DIFF
--- a/desktop/changes/fix-watch-upload-large-library-oom.md
+++ b/desktop/changes/fix-watch-upload-large-library-oom.md
@@ -1,0 +1,1 @@
+- Fix the desktop app crashing during large folder-watch uploads by bounding renderer memory and refreshing gallery state after watcher completion.
--- a/desktop/src/main/services/ffmpeg.ts
+++ b/desktop/src/main/services/ffmpeg.ts
@@ -1,4 +1,9 @@
-import { wrap } from "comlink";
+/**
+ * @file A bridge to the ffmpeg utility process. This code runs in the main
+ * process.
+ */
+
+import { wrap, type Remote } from "comlink";
 import fs from "node:fs/promises";
 import type { FFmpegCommand, ZipItem } from "../../types/ipc";
 import {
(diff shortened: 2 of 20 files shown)


AswinAsok:
Hi @DxTa,

I was trying to reproduce the issue you mentioned. I had a library of about 30k test files totalling around 700MB, and the watch folder uploads seemed to work fine there.

Could you let me know the total file size for your 14k files, so I can put together a similar test dataset?

Also, it seems the `--single-generation` flag changes the GC behaviour for the entire renderer lifetime, which I'd prefer to avoid. Since this issue is specific to the watch folder, I'd rather go with a more scoped change than the current broader one.

Lastly, could you reproduce the issue and share the latest logs as well?

Thanks!

DxTa:
Recently, I got crash with the following uploads. So, I agree with your direction, I will try to reproduce and maybe also try a few directions to check. Now, my hunch is towards Ubuntu Wayland :-)

```
- 1,233 files, 1.03 GB total
- min 13 KB, max 64.1 MB, avg 877 KB
- types: 993 jpg, 210 jpeg, 22 png, 8 gif
- size dist: 171 <100KB · 421 100-500KB · 366 500KB-1MB · 230 1-2MB · 31 2-5MB · 14 >5MB
```

DxTa:
@AswinAsok

Updating direction based on your feedback and local testing.

**Dropping the V8-flag approach.** The `--max-old-space-size=8192` appendSwitch turned out to be actively harmful: it clamps the renderer V8 heap to ~4 GB (verified via `--enable-logging=stderr` OOM: `limit=4096.0MB`), which is *lower* than Chromium's default renderer cap. Without the flag, the renderer gets a higher default limit and survives. `--single-generation` also changes GC behaviour for the whole renderer lifetime, which I agree is too broad for a watch-folder-specific issue. So I'm removing the entire `appendSwitch`.

**New approach — purely app-level (the scoped fix you preferred):**
- `upload-manager.ts`: sub-batch the folder-watch upload (200-file chunks) and flush `itemResults` to GC at each chunk boundary instead of accumulating the full pending set.
- `gallery/reducer.ts` + `gallery/utils/file.ts`: the `uploadFile` reducer case was re-sorting the entire ever-growing `lastSyncedCollectionFiles` and allocating a new sort-key `Map` on *every single file upload* — O(N²) allocation churn that was the real OOM driver (the codebase even had a `// TODO: Consider batching this` on it). Replaced with a single-position sorted insert (no full re-sort, no `Map`).

**Verified locally** on the reproducing library (15k files / 46 GB, avg 3.15 MB/file): with no V8 flags and only the app-level changes, the renderer survives well past the previous ~6 min crash point with a stable ~3.4 GB heap (it previously OOM'd at ~6 min). Your 30k/700 MB set (23 KB/file avg) never hit this because per-file memory pressure scales with payload, not count.

I'll monitor a bit more locally before cleaning this up for review. Branch updated on my fork if you want a look.

*(Drafted with Pi.)*

AswinAsok:
Hey @DxTa,

Do ping me once the changes are ready for review. Or better yet, you could convert the PR to a draft until things are finalized on your end.

AswinAsok:
Hey @DxTa,

I'm planning to close this PR this evening. Just wanted to check first, are you still working on it? Happy to keep it open if so.

DxTa:
I seems still got crashes, although it's less frequent. I also was busy a bit recently, but if you can keep the PR for about a week, it would be great. I can try to rebase and do final checks soon.

AswinAsok:
Hi @DxTa,

Just a quick heads-up: I won't be able to merge this PR as the diff is around 3,000 lines of code, which is too large for me to review, and it touches too many core files. I'm closing it for the time being.

If you're able to redo the fix as a separate, isolated PR with fewer lines of change, I'd really appreciate the same.

DxTa:
Yeah, I know, although it seems to fix my issue, it just cannot merge. Just thinking if it might share some ideas for you or other later. Otherwise, I surrender Ente on linux.

AswinAsok:
I can try to do the fix from my end. It would be great if you could describe the issue you faced and the fix you made in a brief, well-documented way.

The thing is, I haven't heard about this issue from anyone else, so merging this many lines of code into the codebase for it isn't very practical. Hope you understand my concern.

PR #11600, opened 2026-07-22
fix(gallery): read caption as fallback when description missing in Google Takeout metadata

### LINKED ISSUE #11594: Google Takeout import ignores video description
### Description

Hello,

I am using the Google Takeout import feature of ente, and it works mostly fine ! (nice software btw, I love it)

However I have a slight problem : the description metadata is not imported for videos. Strangely, it works fine for pictures however.

### Version

v1.7.24

Ente Desktop

NixOS 26.05

okxint:
Fixes #11594.

When importing from Google Takeout, video descriptions are silently dropped if the `.json` sidecar uses `"caption"` instead of `"description"`.

**Root cause:** `parseMetadataJSONText` reads `metadataJSON.description` for the description field. Google Photos consistently uses `"description"`, but some Takeout exports (especially for videos and for content imported from external sources) store the value under `"caption"` instead. There is no fallback, so those descriptions are lost on import.

**Fix:** Fall back to `metadataJSON.caption` when `metadataJSON.description` is absent. This is a one-character change (`?? metadataJSON.caption`) and is safe — `parseGTNonEmptyString` rejects empty/null values, so a `null` caption won't overwrite a valid description.

### DIFF
--- a/web/packages/gallery/services/upload/metadata-json.ts
+++ b/web/packages/gallery/services/upload/metadata-json.ts
@@ -258,7 +258,7 @@ const parseMetadataJSONText = (text: string) => {
         parseGTLocation(metadataJSON.geoDataExif);
 
     parsedMetadataJSON.description = parseGTNonEmptyString(
-        metadataJSON.description,
+        metadataJSON.description ?? metadataJSON.caption,
     );
 
     if (metadataJSON.favorited === true) {


D4Delta:
This does not fixes my issue unfortunately, my json file does not have a `caption` field (actually, none of my 1000+ json files contains such a field, so I'm wondering if it's actually used ?)

ua741:
> This does not fixes my issue unfortunately, my json file does not have a `caption` field (actually, none of my 1000+ json files contains such a field, so I'm wondering if it's actually used ?)

Thanks for checking this. 

@okxint Please avoid creating PR unless you have explicitly tested it.

PR #11604, opened 2026-07-22
[mobile][photos] Preserve device trash local IDs

aspizu:
When a file that was uploaded by device A is deleted by device B (together with a local downloaded copy), the local ID stored in Trash DB is from device A, which is useless on device B.

With move to trash on android, we need to keep track of the local ID of the trashed file, so we can later on restore/delete it.

This makes device B store the local ID of the local copy that was trashed, device A's Trash DB is unaffected and it's local ID column will still have the local ID from device A's local copy of the file.

The code also sets local ID to null if any image fails to load, for trash files specifically we do not do this, and fix the rest of the code paths that depend on this behavior to fallback to loading the image from network if its a trash file specifically.

### DIFF
--- a/mobile/apps/photos/lib/db/trash_db.dart
+++ b/mobile/apps/photos/lib/db/trash_db.dart
@@ -182,13 +182,13 @@ class TrashDB {
     );
   }
 
-  Future<void> markTrashedOnDevice(Map<int, String> localIDs) async {
+  Future<void> markTrashedOnDevice(List<(int, String)> localIDs) async {
     if (localIDs.isEmpty) return;
     final db = await instance.database;
     await db.transaction((transaction) async {
-      for (final entries in localIDs.entries.toList().chunks(400)) {
+      for (final entries in localIDs.chunks(400)) {
         final batch = transaction.batch();
-        for (final entry in entries) {
+        for (final (uploadedFileID, localID) in entries) {
           batch.rawInsert(
             '''
           INSERT INTO $tableName (
@@ -204,7 +204,7 @@ class TrashDB {
             $columnLocalID = excluded.$columnLocalID,
             $columnIsTrashedOnDevice = 1
         ''',
-            [entry.key, entry.value],
+            [uploadedFileID, localID],
           );
         }
         await batch.commit(noResult: true);
--- a/mobile/apps/photos/lib/db/trash_db.dart
+++ b/mobile/apps/photos/lib/db/trash_db.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:ente_pure_utils/ente_pure_utils.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart';
 import 'package:path_provider/path_provider.dart';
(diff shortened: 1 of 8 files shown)


mobile/apps/photos/lib/utils/delete_file_util.dart
```diff
@@ -128,7 +128,16 @@ Future<void> deleteFilesFromEverywhere(
...
           .toList();
+      final trashedLocalIDsByUploadedID = <int, String>{
+        for (final file in deletedFiles)
+          if (file.uploadedFileID != null &&
+              file.localID != null &&
+              result.trashedIDs.contains(file.localID))
+            file.uploadedFileID!: file.localID!,
+      };
       await trashSyncService.trashFilesOnServer(uploadedFilesToBeTrashed);
+      await trashSyncService.syncTrash();
```
aspizu: **MEDIUM** — The OS move and server trash call have already succeeded before this added sync runs. If the diff fetch, decrypt, or DB write fails, this catch rethrows before `FilesDB.deleteMultipleUploadedFiles` and the update events, so the app reports failure and keeps the file in its main DB even though it is already trashed on the server. The device-local ID is also only in memory; on retry the asset is already missing, so that ID cannot be rebuilt. Save the pending uploaded-ID/local-ID map before the network calls, keep local cleanup from being gated on this follow-up sync, and test a successful trash POST followed by a failed diff fetch/restart.

mobile/apps/photos/lib/db/trash_db.dart
```diff
@@ -108,11 +119,29 @@ class TrashDB {
   }
 
   Future<void> insertMultiple(List<TrashFile> trashFiles) async {
+    if (trashFiles.isEmpty) return;
     final startTime = DateTime.now();
     final db = await instance.database;
+    final uploadedFileIDs = trashFiles
+        .map((trash) => trash.uploadedFileID!)
+        .join(',');
+    final rowsTrashedOnDevice = await db.query(
```
aspizu: **MEDIUM** — This lookup and the later `REPLACE` batch are separate DB operations, while `syncTrash()` can run both here and in an overlapping `RemoteSyncService.sync()`. One sync can read the old false flag, `markTrashedOnDevice` can then save the device ID, and the first sync can finally replace that row with the server ID and false flag. That loses the device ID this change is meant to keep. Make the read-and-write atomic (for example, use one transaction or an upsert that preserves the current flag/local ID at write time) and add an interleaving test.

aspizu (commented):

mobile/apps/photos/lib/services/sync/trash_sync_service.dart
```diff
@@ -138,10 +141,11 @@ class TrashSyncService {
         }
       }
     }
-    final batchedItems = uniqueItems.chunks(batchSize);
-    for (final batch in batchedItems) {
-      final items = batch.map((item) => item.toJson()).toList();
-      await _trashFiles(items);
+    for (final batch in uniqueItems.chunks(batchSize)) {
+      await _trashDB.markTrashedOnDevice({
```
aspizu: **MEDIUM** — This marker can still be lost when a trash sync is already handling an older restore event for the same file. That sync can delete the row after this write, then the trash request can succeed; the next diff will have no marker and will save the upload-time local ID again. Keep pending device markers apart from rows removed by restore/delete diffs, or serialize these flows, and add a test for this order: save marker, apply restore diff, finish trash request, then apply trash diff.

aspizu (commented):

mobile/apps/photos/lib/services/sync/trash_sync_service.dart
aspizu: nothing bad happens if markers are lost, they will just not be deduped in the trash view so it will appear as two files one local and one remote (but in this case ur saying that remote one was restored before sync happened) so this is fine

aspizu:
@codex review

mobile/apps/photos/lib/db/trash_db.dart
```diff
@@ -108,24 +120,36 @@ class TrashDB {
...
+              'IN (${trashFiles.map((trash) => trash.uploadedFileID!).join(',')})',
+        ))
+          row[columnUploadedFileID] as int: row[columnLocalID] as String?,
+      };
+      for (final trashFileBatch in trashFiles.chunks(400)) {
+        final batch = transaction.batch();
+        for (final trash in trashFileBatch) {
+          if (localIDs.containsKey(trash.uploadedFileID)) {
+            trash.localID = localIDs[trash.uploadedFileID];
+            trash.isTrashedOnDevice = true;
```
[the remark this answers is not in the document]
aspizu: fixed

neeraj-pilot:
@codex Review for correctness, any simplification.

mobile/apps/photos/lib/db/trash_db.dart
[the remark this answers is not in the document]
aspizu: this is mostly fine, we can add this later if needed, its okay to keep them unlinked for now

aspizu:
@codex review

aspizu:
@codex review

mobile/apps/photos/lib/ui/viewer/file/thumbnail_widget.dart
```diff
@@ -437,6 +440,11 @@ class _ThumbnailWidgetState extends State<ThumbnailWidget> {
       return null;
     }
 
+    // File is a non-system-only trash file, load thumbnail from remote
+    if (widget.file is TrashFile && widget.file.uploadedFileID != null) {
+      return null;
```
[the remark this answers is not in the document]
aspizu: don't care

mobile/apps/photos/lib/module/download/file.dart
```diff
@@ -35,8 +36,9 @@ Future<File?> getFile(
   bool isOrigin = false,
   bool forGalleryDownload = false, // only relevant for live photos
 }) async {
+  final isRemoteTrashFile = (file is TrashFile && file.uploadedFileID != null);
   try {
-    if (file.isRemoteOnlyFile) {
+    if (file.isRemoteOnlyFile || isRemoteTrashFile) {
```
aspizu, line 41: **MEDIUM** — `getFile()` now downloads every uploaded `TrashFile`, but a retained `localID` still makes `isRemoteOnlyFile` false. On iOS, `getExif()` then treats the result as a local origin and deletes it after reading it; here that result is the cache manager’s remote copy. Opening image details can therefore remove the cached trash image, so a later offline open fails and an online open downloads it again. Return a disposable copy for remote `isOrigin` reads, track the real source when cleaning up, or narrow this remote branch, and add an iOS details-then-reopen test.

mobile/apps/photos/lib/ui/viewer/file/thumbnail_widget.dart
```diff
@@ -337,10 +337,13 @@ class _ThumbnailWidgetState extends State<ThumbnailWidget> {
           if (thumbData == null) {
             if (widget.file.isUploaded) {
               _logger.info("Removing localID reference for " + widget.file.tag);
-              widget.file.localID = null;
-              if (widget.file.isTrash) {
-                unawaited(TrashDB.instance.update(widget.file as TrashFile));
+              if (widget.file is TrashFile) {
+                if (!(widget.file as TrashFile).isTrashedOnDevice) {
+                  widget.file.localID = null;
+                  unawaited(TrashDB.instance.update(widget.file as TrashFile));
```
aspizu: **MEDIUM** — This writes the whole `TrashFile` after an async thumbnail read based only on the object’s old flag. If `markTrashedOnDevice()` saves the device local ID and `true` while this read is in flight, the stale object reaches here with `false`, clears the ID, and writes the DB flag back to `false`. That loses the marker and can show separate local and remote trash items. Clear the ID with a narrow DB update that only runs while the stored marker is still `false`, and add an interleaving test for this order.

aspizu (commented):

PR #11668, opened 2026-07-24
added dark mode to ente.space

snowwysillycat:
## Description
added dark mode and changed all color from hard coded to get from var

## Tests
Screenshot 2026-07-24 170033
Screenshot 2026-07-24 170416
Screenshot 2026-07-24 170425
Screenshot 2026-07-24 170433


anandbaburajan:
Thanks @snowwysillycat! I tried it locally and noticed that the colors of a lot of existing light mode components have changed. There are also a lot of components which look odd in dark mode. Also, let's just auto detect the mode from the system. We can think about giving a toggle later.

snowwysillycat:
> Thanks @snowwysillycat! I tried it locally and noticed that the colors of a lot of existing light mode components have changed. There are also a lot of components which look odd in dark mode. Also, let's just auto detect the mode from the system. We can think about giving a toggle later.

yw can you give give me wt colors has broken in dark mode and ig i broke sm white things while copying i will use the old code white colors later im outside for the day

anandbaburajan:
@snowwysillycat a few specific issues I found:

- `/invite` can get white text/logo on the light `#FAFAFA` background.
- Changing `space.svg` to black breaks the onboarding logo.
- The profile viewer and cover backgrounds should stay black/dark.
- Removing `.green-bg` changes light mode.
- Login, signup and verification are still light while the rest can be dark.

Please restore the original light colors, `space.svg`, `.green-bg` and the fixed viewer colors from `main`.

For the theme, remove the toggle, `localStorage`, `isDarkMode` and `SpaceThemeManager`. Set MUI to `defaultMode="system"` with `storageManager={null}`, and keep `CssBaseline enableColorScheme`.

You can keep the CSS variables, but use the exact old colors in `:root` and put the dark colors under `@media (prefers-color-scheme: dark)`. Then check the remaining screens, dialogs and toasts. Also, `theme-color` needs real colors instead of `var(...)`.

Light mode should look the same as `main`. Push after that and I’ll check it again.

PR #11682, opened 2026-07-26
[mob][auth] Add multi-account support with profile switching

benf2004:
## Description

Lets users stay signed in to several Ente Auth vaults at once and switch between them from **Settings → Account**, up to a cap of 5. Both online accounts and offline vaults can be added, and offline vaults can be given a name (they have no email to tell them apart).

### How the isolation works

Each account owns a **scope** — a prefix applied to its preference keys, its secure storage keys, and its database filenames. `BaseConfiguration.scopedKey()` is the single place that applies it.

The account that exists before this change keeps the **empty** scope, so its preferences, keychain entries and `ente.authenticator.db` stay exactly where they are. There is no migration step, and an install that never adds a second account behaves as it does today.

`ProfileService` owns the registry (which vaults exist, which is active) and the switch itself: it suspends syncing, re-points the databases and configuration, rebuilds the per-account caches, then resumes. It also reconciles on startup, dropping records whose account data was cleared by a logout path that cannot know about profiles (the lock screen's too-many-attempts logout, the revoked-session logout).

### What stays app-wide

The lock screen, theme, locale and general preferences are deliberately shared — they guard the app, not any one vault. In particular, logging out of one account no longer removes the PIN protecting the others.

### Shared package changes

`base_configuration`, `user_service`, `network` and `lock_screen` are touched, but the changes are **no-ops for single-account apps**: the scope defaults to empty, `logoutPreservedKeyPrefixes` defaults to empty (preserving the existing wipe-everything logout), and the new `navigate:` and `shouldClearAppLockOnSignOut:` parameters default to current behavior. Photos and Locker are unaffected.

One behavioral note in `network.dart`: `init()` is now safe to call more than once. The `Dio` instances are created once and mutated in place rather than replaced, because services capture them in `final` fields at construction — replacing them would leave those services pointed at the previous account's endpoint.

## Tests

Adds coverage for key scoping and database naming, profile seeding/reconciliation, what survives an unscoped logout, the settings subtitle source of truth, and the switcher's entry point. `flutter analyze` is clean and all 103 tests in `mobile/apps/auth` pass.

## Strings

Adds 8 keys to `app_en.arb` (`accounts`, `addAccount`, `maxAccountsReached`, `offlineVault`, `renameVault`, `vaultNameHint`, `switchAccount`, `alreadySignedInToAccount`). No other locale files were touched.

---

Supersedes #11681, which was closed to drop an unrelated local build script from the branch history.

### DIFF
--- a/mobile/apps/auth/lib/core/app_wide_preferences.dart
+++ b/mobile/apps/auth/lib/core/app_wide_preferences.dart
@@ -1,0 +1,47 @@
+import 'package:ente_auth/services/auth_theme_preferences.dart';
+import 'package:ente_auth/services/preference_service.dart';
+
+// Preference keys that belong to the app rather than to any one account.
+//
+// They are stored unprefixed, so a logout of the pre-existing (unscoped)
+// profile would otherwise delete them along with that account's keys, silently
+// resetting the surviving profile's theme, language and backup configuration.
+// Configuration.logoutPreservedKeyPrefixes spares everything listed here.
+//
+// Add a key here whenever a new setting is app wide. Account owned settings
+// must NOT be listed: they would then survive a logout and leak into the next
+// session.
+const kAppWidePreferenceKeys = <String>[
+  // Language.
+  localePreferenceKey,
+  // Appearance. ("theme_mode" is a field inside the adaptive blob below, not a
+  // preference key of its own, so it needs no entry here.)
+  AuthThemePreferences.authThemeModeKey,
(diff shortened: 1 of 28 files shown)


benf2004:
See https://github.com/ente/ente/discussions/2510 for a discussion on this feature

benf2004:
I am going to close this PR because I didn't add a feature to remove an offline-local online vault. And to do some more thorough testing.

benf2004:
See new PR: https://github.com/ente/ente/pull/11686

Issue #11518, opened 2026-07-16
[Auth] System tray icon mouse over is displaying unexpected text


p-prakash:
### Description

When mouse hover the system tray icon of Ente Auth app it's displaying unexpected text. As shown in the image below.

<img width="365" height="152" alt="Image" src="https://github.com/user-attachments/assets/f95271fc-fbfc-44f2-bd20-6647b6b972c1" />

### Version

4.4.24

### What product are you using?

Ente Auth

### What platform are you using?

Desktop - Windows

Bitsnbytes14:
Looked into this — Ente Auth's desktop app (Flutter) sets up the tray icon in mobile/apps/auth/lib/main.dart, inside initSystemTray(). It sets the icon and context menu via tray_manager, but never calls trayManager.setToolTip(...). Since no tooltip is explicitly set, Windows falls back to showing the executable/process name on hover, which explains the unexpected text.

Fix should just be adding a setToolTip call alongside the existing setIcon/setContextMenu calls in that function, e.g. await trayManager.setToolTip('Ente Auth');.

I'll take this one and open a PR.

AmanRajSinghMourya:
Closing as resolved in Auth v4.4.25. PR #11683 sets an explicit “Ente Auth” tooltip for the Windows tray icon, replacing the unexpected hover text reported here.

Please re-open this if the issue still exits in the latest version

Bitsnbytes14:
Cool!

PR #11683, opened 2026-07-26
[auth] Set tray icon tooltip on Windows

Bitsnbytes14:
## Description
Ref #11518

Ente Auth's system tray icon on desktop was showing garbled/unexpected text on hover. `tray_manager` copies an uninitialized native tooltip buffer on Windows when no tooltip is explicitly set via `setToolTip()`, which is why the hover text was garbage rather than a clean fallback.
initSystemTray()` sets up the icon and context menu but never calls setToolTip()`. Setting an explicit tooltip is the correct app-level fix.

### DIFF
--- a/mobile/apps/auth/lib/main.dart
+++ b/mobile/apps/auth/lib/main.dart
@@ -54,7 +54,9 @@ Future<void> initSystemTray() async {
       ? 'assets/icons/auth-icon-monochrome-padded.png'
       : _linuxTrayIconPath();
   await trayManager.setIcon(path, isTemplate: true);
-  await trayManager.setToolTip('Ente Auth');
+  if (Platform.isWindows) {
+    await trayManager.setToolTip("Ente Auth");
+  }
   Menu menu = Menu(
     items: [
       MenuItem(key: 'hide_window', label: 'Hide Window'),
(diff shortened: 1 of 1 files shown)


ua741:
@Bitsnbytes14 why do we think this is the actual issue? Also, let's change it to Ente Auth

Bitsnbytes14:
pushed a fix for the capitalization.

On why I think this is the root cause: initSystemTray() sets the icon and context menu via tray_manager, but never actually calls
setToolTip(...). So when there's nothing explicitly set, Windows just falls back to showing the process name on hover  which lines up withwhat got reported as "unexpected text" in #11518. Adding the missing setToolTip call should fix that.
Haven't been able to test it live on a device yet (ran into some local Windows build issues), so this is based on tracing through theray_manager setup rather than seeing it confirmed on screen happy tofollow up once I get that sorted if it'd help.

neeraj-pilot:
Thanks. Please make this Windows-only:

```dart
if (Platform.isWindows) {
  await trayManager.setToolTip("Ente Auth");
}
```

`initSystemTray()` also runs on Linux, where `tray_manager` 0.5.2 does not implement `setToolTip`. That produces a `MissingPluginException`, stopping execution before `setContextMenu()`; the outer `.ignore()` then hides the error.

Also, the Windows text is not a process-name fallback. `tray_manager` is copying an uninitialized native tooltip buffer. Setting an explicit tooltip is a valid app-level workaround, but please update the PR explanation accordingly.

Bitsnbytes14:
Updated the PR description

ua741 (approved):

PR #11686, opened 2026-07-26
[mob][auth] Multi-account support with profile switching

benf2004:
## Description

Lets Auth users stay signed in to several vaults at once and switch between them from Settings > Account, up to a cap of 5.

Each account owns a *scope*: a prefix applied to its preference keys, secure storage keys and database filenames. The account that exists before this change keeps the empty scope, so its data stays exactly where it is and nothing needs migrating.

App wide state — the lock screen, theme, locale and general preferences — stays shared, so logging out of one account cannot remove the PIN protecting the others.

The shared package changes (`base_configuration`, `user_service`, `network`, `lock_screen`) are no-ops for single-account apps: the scope defaults to empty, `logoutPreservedKeyPrefixes` defaults to empty, and both new parameters default to the existing behavior.

Offline vaults can be removed: they have no session to log out of and no account to delete, so Settings > Account gains a destructive "Remove vault" action where Delete account sits for online accounts.

## Tests

- Unit tests for key scoping, database naming, profile seeding from legacy state, reconciliation of profiles whose account data is gone, the profile cap, and add/commit idempotency.
- Widget tests for the switcher row, the settings subtitle, the disabled add row at the cap, and the offline-only remove action.
- `flutter analyze` clean; manually exercised on iOS: add a second account, switch, log out, add and remove an offline vault.


mnvr:
Hey, this is a rather big PR both in terms of code and in terms of product changes. I would like the feature myself, but we can't do it ad-hoc on only on just one product on just one platform. So I'll have to close the PR. Please do open a feature request in the discussions so that people can upvote this, helps us prioritize!

Thank you for the effort

benf2004:
@mnvr ok, thanks for the feedback. There is already a discussion open about it since July 2024:

https://github.com/ente/ente/discussions/2510

What would be required for a PR like this to be approved? I will probably sideload and keep iterating bug fixes until this feature can be added.

mnvr:
The blockers are not conceptual. We will have to spend non-trivial effort and engineering bandwidth to review the code and ensure that it does multi-accounts the way we'd want to add it (we are also adding multi-account support to the new WIP CLI). These 2k lines are all touching core parts of the code, and any bug here can be critical. 

So it is a question of prioritization. We have limited engineering bandwidth.

I'd love to add this sooner than later, and I totally understand the separate work account / personal account use case for auth specifically.

PR #11707, opened 2026-07-27
[mobile][photos] display system trash entries on trash page

aspizu:
- **[mobile][photos] simplify getTrashedFiles function**
- **[mobile][photos] list trash from system trash on android in trash**

### DIFF
--- a/mobile/apps/photos/lib/db/trash_db.dart
+++ b/mobile/apps/photos/lib/db/trash_db.dart
@@ -5,6 +5,7 @@ import 'package:ente_pure_utils/ente_pure_utils.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart';
 import 'package:path_provider/path_provider.dart';
+import "package:photos/models/file/file.dart";
 import 'package:photos/models/file/trash_file.dart';
 import 'package:photos/models/file_load_result.dart';
 import 'package:sqflite/sqflite.dart';
@@ -217,14 +218,14 @@ class TrashDB {
     final results = await db.query(
       tableName,
       where: '$columnTrashDeleteBy != -1',
-      orderBy: '$columnTrashDeleteBy ASC',
+      orderBy: '$columnTrashDeleteBy DESC',
     );
     final files = _convertToFiles(results);
     return FileLoadResult(files, false);
   }
 
-  List<TrashFile> _convertToFiles(List<Map<String, dynamic>> results) {
-    final List<TrashFile> trashedFiles = [];
+  List<EnteFile> _convertToFiles(List<Map<String, dynamic>> results) {
+    final List<EnteFile> trashedFiles = [];
     for (final result in results) {
       trashedFiles.add(_getTrashFromRow(result));
     }
--- a/mobile/apps/photos/lib/db/trash_db.dart
+++ b/mobile/apps/photos/lib/db/trash_db.dart
@@ -212,25 +212,15 @@ class TrashDB {
     });
   }
 
-  Future<FileLoadResult> getTrashedFiles(
-    int startTime,
-    int endTime, {
-    int? limit,
-    bool? asc,
-  }) async {
+  Future<FileLoadResult> getTrashedFiles() async {
     final db = await instance.database;
-    final order = (asc ?? false ? 'ASC' : 'DESC');
     final results = await db.query(
       tableName,
-      where:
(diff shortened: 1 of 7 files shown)


mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -40,13 +39,38 @@ class TrashPage extends StatelessWidget {
...
+          // Insert local only trash entries
+          final localTrashAssets = (await Future.wait(
+            localTrash.map((item) => AssetEntity.fromId(item.localID)),
+          )).whereType<AssetEntity>().toList();
+          for (var i = 0; i < localTrash.length; i++) {
+            final file = TrashFile.fromEnteFile(
+              fileFromAsset('Unknown Folder', localTrashAssets[i]),
+              createdAt: 0,
+              updateAt: 0,
+              deleteBy: localTrash[i].deleteBy,
```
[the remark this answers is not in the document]
aspizu: we will handle this later when adding separate actions for this case.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -40,13 +39,41 @@ class TrashPage extends StatelessWidget {
...
-        return TrashDB.instance.getTrashedFiles(
-          creationStartTime,
-          creationEndTime,
-          limit: limit,
-          asc: asc,
-        );
+      asyncLoader: (creationStartTime, creationEndTime, {limit, asc}) async {
+        final remoteTrash = await TrashDB.instance.getTrashedFiles();
+        if (flagService.internalUser && Platform.isAndroid) {
+          final localTrash = await NativeService.getTrash();
```
aspizu: **MEDIUM** — If the MediaStore query or method channel fails—for example after media access is revoked or while the provider volume is unavailable—this await throws after `remoteTrash` has already loaded. The gallery’s first load then never installs those valid remote rows, so the trash page can stay loading instead of showing Ente trash. Catch and log failures from the native lookup, merge an empty local list, and add a test where the channel throws while remote trash is non-empty.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -40,13 +39,41 @@ class TrashPage extends StatelessWidget {
...
+              localTrash.removeWhere((item) => item.localID == file.localID);
+            }
+          }
+          // Insert local only trash entries
+          final localTrashAssets = (await Future.wait(
+            localTrash.map((item) => AssetEntity.fromId(item.localID)),
+          )).toList();
+          for (var i = 0; i < localTrash.length; i++) {
+            if (localTrashAssets[i] == null) continue;
+            final file = TrashFile.fromEnteFile(
```
aspizu, line 65: **MEDIUM** — This local-only file gets a `localID`, but no `generatedID` or `uploadedFileID`. `SelectedFiles._isMatch` checks only the two missing IDs, so the first long-press selects the row but a second tap cannot find and remove it; a reload can also leave a stale selection that the rebuilt tile does not show. Match on a non-null `localID` as a fallback (or give these rows another stable selection ID), and test select, tap again to deselect, and reload while selected.

mobile/apps/photos/lib/db/trash_db.dart
```diff
@@ -212,29 +213,19 @@ class TrashDB {
...
     final results = await db.query(
       tableName,
-      where:
-          '$columnTrashDeleteBy != -1 AND '
-          '$columnCreationTime >= ? AND $columnCreationTime <= ?',
-      whereArgs: [startTime, endTime],
-      orderBy: '$columnCreationTime ' + order,
-      limit: limit,
+      where: '$columnTrashDeleteBy != -1',
+      orderBy: '$columnTrashDeleteBy DESC',
```
aspizu, line 221: **MEDIUM** — Ordering by `deleteBy` breaks the gallery’s capture-time grouping contract. `GalleryGroups` splits adjacent rows and searches group ranges by `creationTime`; expiry order and capture order are unrelated, so the same day or month can appear under several out-of-order headers and time-based scroll lookup can miss its group. Keep the merged list ordered by `creationTime`, or change/disable capture-time grouping for this page, and test rows whose expiry order is the reverse of their capture order.

aspizu (commented):

PR #11712, opened 2026-07-27
[mob][photos] Refresh referral screens

aman-pilot:
Refreshes the Photos referral screens with the current Ente app bar, theme, menu, input, and button components.

Before / After:
Referral overview before Referral overview after

Tested with Flutter analyze and on the iOS simulator.


mobile/apps/photos/changes/referral-components-ui.md
ua741, line 1: Can we modify existing change log, to keep the noise little less.

AmanRajSinghMourya: Removed, agent might have added it.

ua741 (approved):

PR #11734, opened 2026-07-28
Avoid stale files in Memory Lane links

laurenspriem:
## Summary

- Filter Memory Lane link entries to files in active, non-hidden collections with usable key data.
- Await collection file cleanup before continuing synchronization.
- Preserve deleted file rows when emitting local deletion events.

## Why

Cached Memory Lane timelines can retain file IDs after files or collections are removed. Stale local rows can then be submitted when creating a link, causing the server to reject the complete request. These changes reduce stale candidates and make local deletion synchronization consistent.

## Validation

- `dart format` on the three changed files
- `flutter analyze` on the three changed files

### DIFF
--- a/mobile/apps/photos/lib/db/files_db.dart
+++ b/mobile/apps/photos/lib/db/files_db.dart
@@ -1552,11 +1552,9 @@ class FilesDB with SqlDbBase {
 
   Future<void> deleteCollection(int collectionID) async {
     final db = await instance.sqliteAsyncDB;
-    unawaited(
-      db.execute('DELETE FROM $filesTable WHERE $columnCollectionID = ?', [
-        collectionID,
-      ]),
-    );
+    await db.execute('DELETE FROM $filesTable WHERE $columnCollectionID = ?', [
+      collectionID,
+    ]);
   }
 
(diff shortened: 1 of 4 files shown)


mobile/apps/photos/lib/db/files_db.dart
```diff
@@ -1552,11 +1552,9 @@ class FilesDB with SqlDbBase {
...
   Future<void> deleteCollection(int collectionID) async {
     final db = await instance.sqliteAsyncDB;
-    unawaited(
-      db.execute('DELETE FROM $filesTable WHERE $columnCollectionID = ?', [
-        collectionID,
-      ]),
-    );
+    await db.execute('DELETE FROM $filesTable WHERE $columnCollectionID = ?', [
+      collectionID,
+    ]);
```
neeraj-pilot, line 1557: This might increase the local sync time for fresh. I think during remote collection sync,  we are not batch deleting collections, so if a user has lot of collections that are deleted, we will await on db call, which may be fast enough. no action needed, just fyi @ashilkn

ua741 (approved):

PR #11748, opened 2026-07-29
[mobile][photos] make entire mems banner tappable

aspizu:

### DIFF
--- a/mobile/apps/photos/lib/ui/home/memories/craft_memories.dart
+++ b/mobile/apps/photos/lib/ui/home/memories/craft_memories.dart
@@ -45,72 +45,70 @@ class _CraftMemoriesState extends State<CraftMemories> {
     final l10n = AppLocalizations.of(context);
     return Padding(
       padding: const EdgeInsets.all(MemoryCoverWidget.gap / 2.0),
-      child: SizedBox(
-        width: widget.width,
-        height: widget.height,
-        child: ClipRRect(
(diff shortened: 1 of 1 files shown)


ashilkn (dismissed):

mobile/apps/photos/lib/ui/home/memories/craft_memories.dart
```diff
@@ -50,94 +49,87 @@ class _CraftMemoriesState extends State<CraftMemories> {
...
-                  builder: (BuildContext context, rive.RiveState state) {
-                    if (state is rive.RiveLoaded) {
-                      return rive.RiveWidget(
-                        controller: state.controller,
-                        fit: rive.Fit.cover,
-                      );
-                    }
-                    return const SizedBox.shrink();
-                  },
+          child: GestureDetector(
```
aspizu, line 52: **MEDIUM** — This detector keeps the default `deferToChild` hit-test behavior, so it only receives a tap where a child wins hit testing. While the Rive asset is loading or if it fails, it builds a `SizedBox.shrink()`, and taps on the blank parts of the card are ignored; only the painted text/button area remains reliable. Set `behavior: HitTestBehavior.opaque` (or add another full-size hit-test target) and add a widget test that taps outside the text while Rive is not loaded.

aspizu (commented):

ashilkn (approved):

PR #11761, opened 2026-07-29
[mobile][photos] add native channel to get system trash

aspizu:


mobile/apps/photos/android/app/src/main/kotlin/io/ente/photos/NativeChannel.kt
```diff
@@ -0,0 +1,79 @@
...
+            }
+            else -> result.notImplemented()
+        }
+    }
+
+    private fun getTrash(contentResolver: ContentResolver): List<Map<String, Any>> {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) return emptyList()
+
+        val files = mutableListOf<Map<String, Any>>()
+        val uri = MediaStore.Files.getContentUri(MediaStore.VOLUME_EXTERNAL)
```
aspizu: **MEDIUM** — `MediaStore.Files` also contains audio and other non-photo rows. On Android 11–12, the broad storage permission lets this trash-only query return those rows, but the Dart result has no media type with which to reject them, so callers can treat unrelated system-trash entries as photos. Add a SQL query arg that limits `MEDIA_TYPE` to `MEDIA_TYPE_IMAGE` and `MEDIA_TYPE_VIDEO`, and cover a trashed audio row in a regression test.

aspizu (commented):

PR #11762, opened 2026-07-29
[mobile][photos] support system trash file tracking

aspizu:


mobile/apps/photos/lib/models/file/file.dart
```diff
@@ -79,10 +80,47 @@ class EnteFile {
...
+    pubMagicMetadata = file.pubMagicMetadata;
+    debugCaption = file.debugCaption;
+  }
+
   Future<AssetEntity?> get getAsset {
     if (localID == null) {
       return Future.value(null);
     }
+    if (this is TrashFile && (this as TrashFile).systemTrashID != null) {
+      return AssetEntity.fromId((this as TrashFile).systemTrashID!.toString());
```
aspizu: **MEDIUM** — This redirects normal asset reads to `systemTrashID`, but the live-photo motion path returns earlier from `_getLocalDiskFile` and still calls `Motionphoto.getLivePhotoFile(file.localID!)`. When a remote trash row has a stale or different `localID` from its matched system-trash row, the still image can load locally but the motion clip fails offline even though it is on the device. Use the system-trash ID in the `liveVideo` branch too (or add one helper for the effective asset ID), and add a test for a live photo whose two IDs differ.

aspizu (commented):

mobile/apps/photos/lib/models/file/trash_file.dart
```diff
@@ -11,4 +22,10 @@ class TrashFile extends EnteFile {
   // time after which will will be deleted from trash & user's storage usage
   // will go down
   late int deleteBy;
+
+  // This file is not present on Ente's trash because it was never uploaded, systemTrashID will not not null.
```
ashilkn: Correction: "systemTrashID will not be null."

mobile/apps/photos/lib/ui/viewer/file/thumbnail_widget.dart
```diff
@@ -333,10 +332,13 @@ class _ThumbnailWidgetState extends State<ThumbnailWidget> {
           if (thumbData == null) {
             if (widget.file.isUploaded) {
               _logger.info("Removing localID reference for " + widget.file.tag);
-              widget.file.localID = null;
-              if (widget.file.isTrash) {
-                unawaited(TrashDB.instance.update(widget.file as TrashFile));
+              if (widget.file.asTrashFile != null) {
+                if (!widget.file.asTrashFile!.isSystemOnly) {
+                  widget.file.localID = null;
```
ashilkn, line 337: Why is localID not set to null if the trash file is system only? Will the app not crash (like how we observed for trash) this File is opened outside of trash? It will try to get the asset from localID when the local file is in system trash?

Also, I don't think _getThumbnailFromDisk() would return null thumbData for system trash files since you mentioned fetching of system trash thumbnails do not fail.

mobile/apps/photos/lib/models/file/extensions/file_props.dart
```diff
@@ -42,7 +42,7 @@ extension FilePropsExtn on EnteFile {
 
   bool get canEditMetaInfo => isUploaded && isOwner;
 
-  bool get isTrash => this is TrashFile;
+  TrashFile? get asTrashFile => (this is TrashFile) ? this as TrashFile : null;
```
aspizu, line 48: add isTrashFile ext

mobile/apps/photos/lib/models/file/trash_file.dart
```diff
@@ -1,6 +1,17 @@
...
 class TrashFile extends EnteFile {
+  TrashFile();
+
+  TrashFile.fromEnteFile(
+    super.file, {
+    required this.createdAt,
+    required this.updateAt,
+    required this.deleteBy,
+    required this.isSystemOnly,
+    required this.systemTrashID,
```
aspizu: remove systemTrashID, use localID instead

mobile/apps/photos/lib/models/file/trash_file.dart
```diff
@@ -11,4 +21,7 @@ class TrashFile extends EnteFile {
   // time after which will will be deleted from trash & user's storage usage
   // will go down
   late int deleteBy;
+
+  // This file is not present in Ente's trash because it was never uploaded.
```
ashilkn: Let's remove this comment? It's inaccurate also, an uploaded file can be deleted from device only.

mobile/apps/photos/lib/models/file/trash_file.dart
```diff
@@ -11,4 +21,7 @@ class TrashFile extends EnteFile {
...
+  bool isSystemOnly = false;
```
ashilkn: I think 'isDeviceOnly' would be a better name since we already use 'device only' term for photos that are on device only. It would be easier to understand that it means this trashed file is available on device only.

mobile/apps/photos/lib/module/download/file.dart
```diff
@@ -79,15 +79,16 @@ Future<File?> _getLocalDiskFile(
...
     return localFile.exists().then((exist) {
       return exist ? localFile : null;
     });
   } else if (file.fileType == FileType.livePhoto && liveVideo) {
     return Motionphoto.getLivePhotoFile(file.localID!);
   } else {
     return file.getAsset.then((asset) async {
-      if (asset == null || !(await asset.exists)) {
+      if (asset == null ||
+          !(await asset.exists || file.isSystemOnlyTrashFile)) {
```
ashilkn: If file is `systemOnlyTrashFile`, let's return null since the asset.file would be useless? If there is an uploaded file who's local file is in system trash, opening the uploaded file from outside of trash would try to use `asset.file` which would fail.

ashilkn: We should add a short comment specifying why we return null if it's a system only trash file so that it's clear.

mobile/apps/photos/lib/ui/viewer/file/file_bottom_bar.dart
ashilkn, line 1: The delete from trash option in file viewer would be missing for a device trash file since we're only including it for Ente trash file here

mobile/apps/photos/lib/models/file/trash_file.dart
```diff
@@ -1,14 +1,23 @@
 import 'package:photos/models/file/file.dart';
 
-class TrashFile extends EnteFile {
+sealed class TrashFile extends EnteFile {
   TrashFile();
 
-  TrashFile.fromEnteFile(
+  TrashFile.from(super.file, {required this.deleteBy}) : super.from();
+
+  // time after which will will be deleted from trash & user's storage usage
```
ashilkn, line 8: Can remove this comment, TMI I feel.

mobile/apps/photos/lib/ui/viewer/file/file_bottom_bar.dart
aspizu: this is not part of this PR, will be added in later PRs

mobile/apps/photos/lib/utils/delete_file_util.dart
```diff
@@ -292,7 +293,11 @@ Future<List<EnteFile>> deleteFilesOnDeviceOnly(
   return deletedFiles;
 }
 
-Future<bool> deleteFromTrash(BuildContext context, List<EnteFile> files) async {
+Future<bool> deleteFromTrash(
```
ashilkn, line 296: TODO: rename this to deleteFromEnteTrash

ashilkn (approved):

PR #11785, opened 2026-07-30
[mobile][photos] add ente and device trash pages

aspizu:


mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -1,70 +1,111 @@
...
+                    setState(() {
+                      _isOnEnteTrash = true;
+                    });
+                  },
+                ),
+                TagChipComponent(
+                  label: l10n.onDevice,
+                  state: !_isOnEnteTrash ? .selected : .unselected,
+                  onTap: () {
+                    setState(() {
```
aspizu: **HIGH** — Switching trash sources keeps the same selection even though the selected thumbnails disappear. If an Ente-trash item is selected, then the user taps On device and confirms Permanently delete, the action still deletes the hidden Ente item. Clear the selection before changing sources, or keep separate selection state per source, and add a select-switch-delete test.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -82,4 +123,41 @@ class TrashPage extends StatelessWidget {
...
+      return await TrashDB.instance.getTrashedFiles(
+        creationStartTime,
+        creationEndTime,
+        limit: limit,
+        asc: asc,
+      );
+    }
+    final deviceTrash = await NativeService.getTrash();
+    final deviceTrashAssets = await Future.wait(
+      deviceTrash.map((t) => AssetEntity.fromId(t.localID.toString())),
```
aspizu, line 164: **MEDIUM** — This lookup uses photo_manager’s normal MediaStore query, which does not opt into trashed rows. `NativeService.getTrash()` can find them because its query sets `QUERY_ARG_MATCH_TRASHED`, but `AssetEntity.fromId()` returns null here and line 149 drops the row, so the On device page can stay empty while system trash has files. Return the needed asset fields from the trash-aware native query, or add a trash-aware asset lookup, and test with a trashed image on Android 11+.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -82,4 +123,41 @@ class TrashPage extends StatelessWidget {
...
+      final trash = deviceTrash[i];
+      final asset = deviceTrashAssets[i];
+      if (asset == null) continue;
+      files.add(
+        TrashFile.fromEnteFile(
+          fileFromAsset(trash.deviceFolder, asset),
+          createdAt: trash.deleteBy - 30 * 86400,
+          updateAt: trash.deleteBy - 30 * 86400,
+          deleteBy: trash.deleteBy,
+          isSystemOnly: true,
```
aspizu: **MEDIUM** — These system-only rows are still shown with the Ente trash action bar. They have no uploaded file ID, but both restore and permanent delete later force `uploadedFileID!`, so either action throws after the user confirms it and the device trash is unchanged. Route system-only files to native trash actions, or hide these actions until they are supported, and test both buttons with a device-only file.

mobile/apps/photos/lib/ui/viewer/gallery/gallery_app_bar_widget.dart
```diff
@@ -559,6 +572,10 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
       return actions;
     }
 
+    if (galleryType == GalleryType.trash) {
```
aspizu: **MEDIUM** — This removes the overflow menu for every trash page, and this PR also removes its `emptyTrash` option. The Ente trash view therefore loses its only one-step Empty trash action on Android and on platforms that never show the new device tab. Keep the Ente action and suppress or replace it only while device trash is active; add a widget test for both sources.

aspizu (commented):

aspizu (commented):

mobile/apps/photos/lib/ui/viewer/actions/file_selection_actions_widget.dart
```diff
@@ -1065,15 +1070,35 @@ class _FileSelectionActionsWidgetState
...
   }
 
-  Future<void> _permanentlyDelete() async {
-    if (await deleteFromTrash(context, widget.selectedFiles.files.toList())) {
+  Future<void> _permanentlyDeleteFromTrash() async {
+    final isSystemOnly = widget.selectedFiles.files.every(
+      (f) => f.asTrashFile!.isSystemOnly,
+    );
+    if (isSystemOnly) {
+      await _deleteFromSystemTrash(widget.selectedFiles);
```
aspizu: **MEDIUM** — If MediaStore rejects this delete because the trash row expired or media access changed after the page loaded, this error escapes the tap callback, so the user gets no error message and cannot tell whether anything was deleted. Catch the failure and show the same error dialog used by the system-restore path, add a failing-platform-call test, and guard the single-file system delete path in `file_bottom_bar.dart` the same way.

mobile/apps/photos/lib/ui/viewer/file/file_bottom_bar.dart
```diff
@@ -239,6 +242,13 @@ class FileBottomBarState extends State<FileBottomBar> {
           child: IconButton(
             icon: const Icon(CupertinoIcons.gobackward, color: Colors.white),
             onPressed: () {
+              if (widget.file.isDeviceTrash) {
```
aspizu, line 244: **MEDIUM** — This device-trash branch cannot run because `_addTrashOptions` is called only for `isEnteTrash`, while the normal action block also excludes every trash file. Opening an On-device-trash item therefore shows no Restore or Delete buttons. Call `_addTrashOptions` for all trash files (or build a device-trash bar) and add a detail-view test for both actions.

mobile/apps/photos/lib/models/file/trash_file.dart
```diff
@@ -30,4 +36,19 @@ class EnteTrashFile extends TrashFile {
 
 class DeviceTrashFile extends TrashFile {
   DeviceTrashFile();
+
+  DeviceTrashFile.from(super.file, {required super.deleteBy}) : super.from();
```
aspizu, line 34: **MEDIUM** — This copy leaves both `generatedID` and `uploadedFileID` null. `SelectedFiles._isMatch` checks only those IDs, so tapping a device-trash tile adds it to the selection but tapping it again cannot find and remove it; selection state and actions can then disagree. Give device-trash files a stable ID derived from `localID`, or match on `localID`, and test select then deselect.

aspizu (commented):

mobile/apps/photos/lib/models/file/trash_file.dart
aspizu: will be fixed in upstack

ashilkn:
@codex review

mobile/apps/photos/lib/models/file/trash_file.dart
```diff
@@ -18,7 +20,11 @@ class EnteTrashFile extends TrashFile {
     required super.deleteBy,
     required this.createdAt,
     required this.updateAt,
-  }) : super.from();
+  }) : super.from() {
+    if (localID != null) {
```
ashilkn: Since this ID derivation is only needed for device-trash files, could we move it to DeviceTrashFile.from? Ente trash files loaded from TrashDB already receive -uploadedFileID as their synthetic ID. Keeping the device-specific behavior on DeviceTrashFile makes that distinction explicit.

mobile/apps/photos/lib/models/file/trash_file.dart
```diff
@@ -30,4 +36,19 @@ class EnteTrashFile extends TrashFile {
 
 class DeviceTrashFile extends TrashFile {
   DeviceTrashFile();
+
+  DeviceTrashFile.from(super.file, {required super.deleteBy}) : super.from();
+}
+
+AssetEntity trashFileToAssetEntity(EnteFile file) {
```
ashilkn: Let's use `DeviceTrashFile` as accepted parameter instead of `EnteFile`? If we are not planning to use this function for anything except `DeviceTrashFile`, let's scope it to only work for `DeviceTrashFile`. Explicitly casting `EnteFile` to `DeviceTrashFile` at callers of this function should make it work.

mobile/apps/photos/lib/ui/viewer/actions/file_selection_actions_widget.dart
```diff
@@ -1259,3 +1284,24 @@ class _FileSelectionActionsWidgetState
     }
   }
 }
+
+Future<void> _restoreFilesFromDeviceTrash(SelectedFiles selectedFiles) async {
+  final files = selectedFiles.files
```
ashilkn: Let's log on failure using Logger.warning()

ashilkn: @aspizu Logger.severe gets sent to sentry. Logger.warning doesn't. I don't think this issue is serious enough to be sent to sentry. 

Also, this class already has a Logger instance `_logger`, let's use that. This is the general convention in the codebase.

mobile/apps/photos/lib/ui/viewer/file/file_bottom_bar.dart
```diff
@@ -263,11 +272,20 @@ class FileBottomBarState extends State<FileBottomBar> {
               color: Colors.white,
             ),
             onPressed: () async {
-              final trashedFile = <EnteTrashFile>[];
-              trashedFile.add(widget.file.asEnteTrashFile!);
-              if (await deleteFromTrash(context, trashedFile) == true) {
-                if (!mounted) return;
-                Navigator.pop(context);
+              if (widget.file.isDeviceTrash) {
+                final selectedFiles = SelectedFiles()
```
ashilkn: Using `SelectedFiles` here doesn't feel right.

aspizu: `permanentlyDeleteFromDeviceTrash` (and thus `_deleteFromDeviceTrash`) is used for both deleting a selection, and a single item. `_deleteFromDeviceTrash` unselects the deleted items after a success and fires a bus event to trigger re-render of the trash gallery. creating a temporary `SelectedFiles` object is a pattern from the existing code, see: https://github.com/ente/ente/blob/f495c6416652909214e7820d2cf5d7b2cf5da221/mobile/apps/photos/lib/ui/viewer/file/file_bottom_bar.dart#L242

ashilkn: Let's not follow bad practices from other parts of the codebase. I guess an agent did this because it is being used elsewhere. We need to work towards agents not writing bad code.

mobile/apps/photos/lib/ui/viewer/gallery/gallery_app_bar_widget.dart
```diff
@@ -559,6 +575,10 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
...
```
ashilkn: Making _hasOverflowMenuActions return false for Trash would be the better way to do this.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -1,70 +1,130 @@
...
+import "package:photos/ui/viewer/actions/file_selection_overlay_bar.dart";
+import "package:photos/ui/viewer/gallery/gallery.dart";
+import "package:photos/ui/viewer/gallery/gallery_app_bar_widget.dart";
 import "package:photos/ui/viewer/gallery/state/gallery_boundaries_provider.dart";
 import "package:photos/ui/viewer/gallery/state/gallery_files_inherited_widget.dart";
 import "package:photos/ui/viewer/gallery/state/selection_state.dart";
+import "package:photos/utils/device_info.dart";
+
+Future<void> showTrashPage(BuildContext context) async {
+  final isDeviceTrashSupported =
```
ashilkn, line 33: Let's make it available to internal users only too.

ashilkn (dismissed):

ashilkn:
@codex Review the changes made since I last approved the changes

aspizu:
@codex review

mobile/apps/photos/lib/utils/delete_file_util.dart
```diff
@@ -1165,12 +1200,14 @@ class DeleteConfirmationSheetState extends State<DeleteConfirmationSheet> {
   Widget build(BuildContext context) {
     final l10n = context.strings;
     final title = l10n.deleteItemsQuestion(count: widget.count);
-    var body = l10n.selectedFilesSavedOnDeviceOnly;
+    var body = widget.isTrashAction
+        ? l10n.filesCanBeRestoredFromTrash(count: widget.count)
```
aspizu: **HIGH** — This promises recovery based only on the platform, but local-only `ente-shared-media://` files take `_tryDeleteSharedMediaFiles`, which permanently deletes the app-sandbox file. On iOS and internal Android 11+ builds, a user can therefore erase the only copy after being told it can be restored from Trash. Derive this copy from the actual deletion path (and disable it for sandbox-backed or mixed selections), then add a local-only shared-media regression test.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -82,4 +144,37 @@ class TrashPage extends StatelessWidget {
...
+    final deviceTrashAssets = await Future.wait(
+      deviceTrash.map((t) => AssetEntity.fromId(t.localID.toString())),
+    );
+    final List<EnteFile> files = [];
+    for (var i = 0; i < deviceTrash.length; i++) {
+      final trash = deviceTrash[i];
+      final asset = deviceTrashAssets[i];
+      if (asset == null) continue;
+      files.add(
+        DeviceTrashFile.from(
```
aspizu, line 172: **MEDIUM** — Device-trash videos created here enter the default Android native viewer, whose `loadOriginal()` calls `getFile(..., isOrigin: true)`. `_getLocalDiskFile` deliberately returns null for every device-trash file, and these rows have no uploaded fallback, so opening one never supplies a playable source. Use the asset media URL as the media-kit path does (or route this type to media-kit), and add an Android playback regression test.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -1,70 +1,132 @@
...
+        appBar: GalleryAppBarWidget.sliverConfig(
+          GalleryType.trash,
+          l10n.trash,
+          _selectedFiles,
+          subtitle:
+              l10n.itemsShowTheNumberOfDaysRemainingBeforePermanentDeletion,
+          bottom: appBarBottom,
+        ),
+        asyncLoader: _asyncLoader,
+        reloadEvent: Bus.instance.on<FilesUpdatedEvent>().where(
```
aspizu, line 113: **MEDIUM** — This Ente-only reload filter cannot refresh the new device-trash source. If an on-device item is restored, deleted, or expires while Ente is backgrounded, returning to the open tab leaves stale entries because no lifecycle reload is registered and device-only files have no uploaded ID. Reload the device source on resume or listen for relevant local-media changes, and test an external trash mutation followed by resume.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
aspizu: todo, will be fixed later

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
aspizu: todo, will be fixed later

ashilkn (approved):

PR #11848, opened 2026-08-02
[mobile][photos] adjust SVG icon sizing in video editor bottom bar

r4khul:
### Issue: Bottom bar icons were barely visible in video editor

The editor bottom bar icons looked way too tiny and faint on screen because the source SVG assets have ~50% internal margins (`24px` artwork sitting inside a `41px` canvas).

* **Fix**: Scaled `CircularIconButton` to `size * 0.68` (40.8px). This forces the inner SVG artwork to render at an actual `24px` visual weight, matching `HugeIcon` and system defaults.

**Visuals:**

image

## Tests
Tested & verified on physical devices - realme C67 5G & realme Pad 2.

### DIFF
--- a/mobile/apps/photos/changes/video-editor-bottom-bar-icons.md
+++ b/mobile/apps/photos/changes/video-editor-bottom-bar-icons.md
@@ -1,0 +1,1 @@
+- Adjusted SVG icon sizing in the video editor bottom bar for better visiblity. (@r4khul)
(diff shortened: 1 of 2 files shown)


r4khul:
PTAL @ashilkn 

i do think we could eventually switch to HugeIcons for the trim, crop, and rotate options to stay consistent with the image editor's bottom bar. However, this fix gets the job done for now since it directly addresses the crop ratio SVGs inside the video editor too.

the above fix is a simple math + intuition.
let me know if anything like a comment or variable for 0.68 is needed (since codex pinned it as magic num).

ashilkn:
Thanks @r4khul, looks good. 
Could you add a change entry like : https://github.com/r4khul/ente/pull/1 but with a (@r4khul) and the end of the string? We use this for preparing change log and to know what changes have been merged for testing.

r4khul:
@ashilkn added the change entry.
will remember to add this hereafter!

ashilkn (approved):

ashilkn:
Thanks for the contribution @r4khul!

PR #11944, opened 2026-08-05
[server] Harden sharing recipient identity lookup

neeraj-pilot:
- Restore public-key lookup to the authenticated email-only flow and require bulk shares to bind recipient email and user ID before any writes.
- Reject automatic bulk shares before processing when owner and recipient do not share the same non-null family admin, using a dedicated 403 error code.
- The request contract changes are safe to coordinate because these APIs are consumed only by Ente internal clients.
- Validation: cd server && ./scripts/lint.sh && ./scripts/test-with-postgres.sh host; cd rust && cargo test --locked --features museum,ente-photos/ml-assets.


neeraj-pilot:
@codex review

ua741 (approved):

PR #11974, opened 2026-08-05
[rust][photos] Reduce bulk vector search memory

laurens-pilot:
## Opportunity

Similar Images passes every eligible indexed photo ID to VectorDB.bulk_search_keys. The Rust implementation allocated a contiguous scratch buffer sized as potential keys times embedding dimensions before running the searches, even though only one query vector is needed at a time.

## Evidence of value

CLIP image embeddings are 512-dimensional, so the removed scratch buffer consumed 2,048 bytes per potential key: 195.3 MiB for a 100,000-photo library.

A temporary 5,000-key, 512-dimension probe isolated this buffer using the same test binary. Peak RSS fell from 43,728,896 bytes to 33,357,824 bytes, a 10,371,072-byte reduction that matches the removed allocation. Wall time was effectively unchanged (4.46 s before, 4.42 s after), so this PR makes no latency claim.

## Implementation

Reuse one embedding-sized query buffer, search each contained key immediately, and append its keys and distances to the aligned result vectors. A focused test covers missing-key filtering, contained-key order, self matches, and result alignment.

## Risk

Low. The public API and output shape are unchanged, and the method still performs the same contains, get, and search operations in input order. The change only shortens the lifetime and count of query-vector storage.

## Verification

- cargo fmt --all -- --check
- cargo test -p ente-photos --features usearch ml::vector_db::tests::bulk_search_keys_preserves_contained_key_order_and_result_alignment
- cargo clippy -p ente-photos --features usearch --tests -- -D warnings

### DIFF
--- a/rust/crates/photos/src/ml/vector_db.rs
+++ b/rust/crates/photos/src/ml/vector_db.rs
@@ -354,36 +354,22 @@ impl VectorDB {
         exact: bool,
     ) -> Result<BulkSearchByKeyMatch, String> {
         let dimensions = self.index.dimensions();
-        let mut embedding_data = vec![0.0f32; potential_keys.len() * dimensions];
-        let mut contained_keys = Vec::with_capacity(potential_keys.len());
-        let mut actual_query_count = 0;
+        let result_capacity = potential_keys.len().min(self.index.size());
+        let mut contained_keys = Vec::with_capacity(result_capacity);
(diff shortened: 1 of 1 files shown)


laurens-pilot:
@codex review

laurenspriem (approved):

mnvr (approved):

PR #12117, opened 2026-08-11
[mobile] Add floating action button component

aspizu:
## What changed

- add a reusable floating action button to `ente_components`
- support primary and secondary styles with icon-only, label-only, and icon-plus-label content
- handle async taps with guarded loading and success states
- add the component to the catalog with light and dark theme previews
- link the component to its Figma spec

## Why

This gives mobile apps one shared FAB that follows the Ente design system and the same async execution rules as the existing button components.

## Checks

- `fvm dart analyze lib/components/buttons/fab_component.dart example/lib/main.dart`

### DIFF
--- a/mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
+++ b/mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
@@ -143,18 +143,26 @@ class _TrashPageState extends State<_TrashPage> {
               ],
             ),
           ),
-          floatingActionButton: Padding(
-            padding: const EdgeInsets.only(bottom: 4),
-            child: FABComponent(
-              label: l10n.deleteAll,
-              icon: const HugeIcon(icon: HugeIcons.strokeRoundedDelete02),
-              shouldShowSuccessConfirmation: false,
-              shouldShowSuccessState: false,
-              shouldSurfaceExecutionStates: false,
-              onTap: () async {
-                await emptyTrash(context, _isOnEnteTrash);
-              },
-            ),
+          floatingActionButton: ListenableBuilder(
+            listenable: _selectedFiles,
+            builder: (context, _) {
+              if (_selectedFiles.files.isNotEmpty) {
+                return const SizedBox.shrink();
+              }
+              return Padding(
+                padding: const EdgeInsets.only(bottom: 4),
+                child: FABComponent(
+                  label: l10n.deleteAll,
+                  icon: const HugeIcon(icon: HugeIcons.strokeRoundedDelete02),
+                  shouldShowSuccessConfirmation: false,
+                  shouldShowSuccessState: false,
+                  shouldSurfaceExecutionStates: false,
+                  onTap: () async {
+                    await emptyTrash(context, _isOnEnteTrash);
+                  },
+                ),
+              );
+            },
           ),
           floatingActionButtonLocation:
               FloatingActionButtonLocation.centerFloat,
--- a/mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
+++ b/mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
@@ -153,9 +153,10 @@ class _TrashPageState extends State<_TrashPage> {
                 child: FABComponent(
                   label: l10n.deleteAll,
                   icon: const HugeIcon(icon: HugeIcons.strokeRoundedDelete02),
-                  onTap: () async {
-                    await emptyTrash(context, _isOnEnteTrash);
-                  },
+                  onTap: () => showConfirmDeleteAllTrashSheet(
+                    context,
+                    _isOnEnteTrash,
+                  ),
                 ),
               );
             },
--- a/mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
+++ b/mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
@@ -51,6 +51,7 @@ class _TrashPage extends StatefulWidget {
 
 class _TrashPageState extends State<_TrashPage> {
   bool _isOnEnteTrash = !isLocalGalleryMode;
+  bool _hasTrashFiles = false;
   final _selectedFiles = SelectedFiles();
 
   @override
@@ -71,6 +72,7 @@ class _TrashPageState extends State<_TrashPage> {
                       setState(() {
                         _selectedFiles.clearAll();
                         _isOnEnteTrash = true;
+                        _hasTrashFiles = false;
                       });
                     }
                   },
@@ -83,6 +85,7 @@ class _TrashPageState extends State<_TrashPage> {
                       setState(() {
                         _selectedFiles.clearAll();
                         _isOnEnteTrash = false;
+                        _hasTrashFiles = false;
                       });
                     }
                   },
@@ -145,7 +148,7 @@ class _TrashPageState extends State<_TrashPage> {
           floatingActionButton: ListenableBuilder(
             listenable: _selectedFiles,
             builder: (context, _) {
-              if (_selectedFiles.files.isNotEmpty) {
+              if (_selectedFiles.files.isNotEmpty || !_hasTrashFiles) {
                 return const SizedBox.shrink();
               }
               return Padding(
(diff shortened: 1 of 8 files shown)


mobile/apps/photos/lib/utils/delete_file_util.dart
```diff
@@ -1381,3 +1386,20 @@ Future<void> _deleteFromSystemTrash(SelectedFiles selectedFiles) async {
...
 }
+
+Future<void> _emptySystemTrash(BuildContext context) async {
+  final trash = (await NativeService.getTrash())
+      .map((f) => f.localID.toString())
+      .toList();
+  final deletedIDs = <String>{};
+  try {
+    for (final batch in trash.chunks(batchSize)) {
+      final result = await PhotoManager.editor.deleteWithIds(batch);
```
aspizu, line 1442: **MEDIUM** — With more than `batchSize` trashed items, cancelling Android’s delete prompt returns an empty result, but this loop continues and opens the prompt for the next batch. A later batch can then be deleted even though the user cancelled the operation. Stop after an empty or incomplete result, keep the reload for any items already removed, and add a test where one of several batches is cancelled.

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
```diff
@@ -0,0 +1,317 @@
...
+
+  final FutureOr<void> Function()? onTap;
+
+  final FABComponentVariant variant;
+
+  final bool isDisabled;
+
+  final Widget? icon;
+
+  const FABComponent({
```
aspizu, line 22: **MEDIUM** — The public API supports icon-only FABs, but it has no semantic label or tooltip input. `HugeIcon` also adds no label, so a screen reader announces an unnamed tap target and the user cannot tell what it does. Require an accessible label for the icon-only form, expose it through `Semantics` or `Tooltip`, and add a semantics test.

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
```diff
@@ -0,0 +1,317 @@
...
+      mainAxisSize: MainAxisSize.min,
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [
+        if (widget.icon != null)
+          IconTheme.merge(
+            data: IconThemeData(color: foreground, size: IconSizes.small),
+            child: widget.icon!,
+          ),
+        if (widget.label != null)
+          Text(
```
aspizu: **MEDIUM** — This `Text` is a non-flex child of a shrink-wrapped `Row`, so it is laid out without a finite horizontal limit and `TextOverflow.ellipsis` cannot take effect. A long translation or large text scale on a narrow screen causes a `RenderFlex` overflow and can draw the FAB past the viewport. Give the label a finite width, such as with a constrained FAB and `Flexible`, and add a narrow-screen, large-text widget test.

aspizu (commented):

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
AmanRajSinghMourya, line 1: We have a button called `DynamicFAB` in `mobile/packages/ui/lib/components/buttons/dynamic_fab.dart`.
Maybe you can use that button instead of creating a new one

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
```diff
@@ -0,0 +1,317 @@
...
+        !widget.isDisabled &&
+        widget.onTap != null &&
+        !_isExecuting &&
+        !_isSuccessful;
+    final colors = context.componentColors;
+    final backgroundColor = switch (widget.variant) {
+      FABComponentVariant.primary =>
+        _isPressed ? colors.primaryDark : colors.primary,
+      FABComponentVariant.secondary =>
+        _isPressed ? colors.fillDarker : colors.fillDark,
```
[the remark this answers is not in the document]
aspizu: we don't need this currently, could be added later on.

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
aspizu: we need to remove dynamic FAB in the future

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
aspizu: can be added later

mobile/apps/photos/lib/utils/delete_file_util.dart
```diff
@@ -1417,3 +1434,22 @@ Future<(Set<String>, Object?)> _deleteFromDeviceTrash(
...
+Future<bool> _emptyDeviceTrash() async {
+  final trash = (await DeviceTrashClient.instance.getFiles())
+      .map((f) => f.localID.toString())
+      .toList();
+  final deletedIDs = <String>{};
+  try {
+    for (final batch in trash.chunks(batchSize)) {
+      final result = await PhotoManager.editor.deleteWithIds(batch);
+      deletedIDs.addAll(result);
+      if (result.length != batch.length) break;
```
ashilkn, line 1444: Why break here? If we remove this, if it doesn't throw, it would just continue with the rest of the batches which would be better.

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
```diff
@@ -0,0 +1,149 @@
...
+import "package:ente_components/theme/icon_sizes.dart";
+import "package:ente_components/theme/motion.dart";
+import "package:ente_components/theme/text_styles.dart";
+import "package:ente_components/theme/theme.dart";
+import "package:flutter/material.dart";
+
+enum FABComponentVariant { primary, secondary }
+
+/// Figma: https://www.figma.com/design/BuBNPPytxlVnqfmCUW0mgz/Ente-Visual-Design?node-id=21843-6537&m=dev
+/// Section: Buttons / Floating action button
```
ashilkn: @AmanRajSinghMourya Only the figma link needs to be kept, right?

mobile/apps/photos/lib/utils/delete_file_util.dart
aspizu: if the user hits Deny on one prompt, we stop immediately instead of doing more prompts for the rest of the batches

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
AmanRajSinghMourya: Yes we can remove rest and  just keep Figma link

aspizu (commented):
Automated code review: actionable findings inline.

mobile/packages/ente_components/lib/components/buttons/fab_component.dart
```diff
@@ -0,0 +1,143 @@
...
+                (_, _) => const EdgeInsets.only(left: 20, right: 24),
+              },
+              child: Row(
+                spacing: 8,
+                mainAxisSize: MainAxisSize.min,
+                mainAxisAlignment: MainAxisAlignment.center,
+                children: [
+                  if (widget.icon != null)
+                    IconTheme.merge(
+                      data: IconThemeData(
```
aspizu: **LOW** — `IconTheme` cannot enforce this size for `HugeIcon`: version 1.1.7 defaults its own `size` to 24 and reads `IconTheme.size` only when that value is null. The trash FAB passes a default `HugeIcon`, so it renders at 24px instead of the intended 18px; the catalog masks this by passing 18px explicitly. Constrain the icon child to `IconSizes.small` and add a widget test using a default `HugeIcon`.

mobile/apps/photos/lib/ui/viewer/gallery/trash_page.dart
```diff
@@ -140,6 +145,27 @@ class _TrashPageState extends State<_TrashPage> {
...
           ),
+          floatingActionButton: ListenableBuilder(
+            listenable: _selectedFiles,
+            builder: (context, _) {
+              if (_selectedFiles.files.isNotEmpty || !_hasTrashFiles) {
+                return const SizedBox.shrink();
+              }
+              return Padding(
+                padding: const EdgeInsets.only(bottom: 4),
+                child: FABComponent(
```
aspizu, line 156: **LOW** — This adds a user-facing Delete all action, but the PR has no entry under `mobile/apps/photos/changes`. Those files are concatenated into Photos release notes, so this feature will be omitted. Add a short entry such as `trash-delete-all.md` describing the action.

ashilkn (dismissed):

ashilkn (approved):

PR #12526, opened 2026-09-01
[mobile][photos] generate memory lane schedule on startup

aspizu:
## Summary

- generate and persist a memories-strip schedule on startup for eligible person and cluster timelines
- expose the currently scheduled timeline with feature, visibility, hidden-file, and face-crop validation
- invalidate unusable schedule entries and retain a 30-day cooldown before rescheduling a timeline

## Scope

This PR only appends to the existing valid schedule when the future scheduling horizon needs another entry. If the current entry becomes invalid while a future entry still exists, repairing the resulting gap or rebuilding the schedule is intentionally out of scope.

### DIFF
--- a/mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
+++ b/mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
@@ -116,7 +116,11 @@ class MemoryLaneService {
           assigned.add(cluster.id);
         }
       }
-      _topNClusters = await _mlDataDB.getClustersForMemoryLane(assigned);
+      try {
+        _topNClusters = await _mlDataDB.getClustersForMemoryLane(assigned);
+      } catch (e, s) {
+        _logger.severe("getClustersForMemoryLane failed:", e, s);
+      }
     }
     if (flagService.internalUser) {
       try {
--- a/mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
+++ b/mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
@@ -82,13 +82,6 @@ class MemoryLaneService {
       Bus.instance.on<MLConsentChangedEvent>().listen(_handleMlConsentChange);
       _scheduleStartupBackfill();
       _initialized = true;
-      if (flagService.internalUser) {
-        try {
-          await _scheduleTimelinesForMemoriesStrip();
-        } catch (e, s) {
-          _logger.severe("_scheduleTimelinesForMemoriesStrip failed:", e, s);
-        }
-      }
       await _queueFullRecompute();
     } catch (e, s) {
       _logger.severe("Initialization failed", e, s);
@@ -115,13 +108,6 @@ class MemoryLaneService {
         return;
       }
       persons.addAll(await PersonService.instance.getPersons());
-      for (final person in persons) {
-        if (person.data.isIgnored) {
-          await _invalidateTimeline(person.remoteID);
-          continue;
-        }
-        schedulePersonRecompute(person.remoteID, force: force);
-      }
     }
     if (flagService.internalUser) {
       final assigned = <String>{};
@@ -131,6 +117,24 @@ class MemoryLaneService {
         }
       }
       _topNClusters = await _mlDataDB.getClustersForMemoryLane(assigned);
+    }
+    if (flagService.internalUser) {
+      try {
+        await _scheduleTimelinesForMemoriesStrip();
+      } catch (e, s) {
+        _logger.severe("_scheduleTimelinesForMemoriesStrip failed:", e, s);
+      }
+    }
+    if (!isLocalGalleryMode) {
+      for (final person in persons) {
+        if (person.data.isIgnored) {
+          await _invalidateTimeline(person.remoteID);
+          continue;
+        }
+        schedulePersonRecompute(person.remoteID, force: force);
+      }
+    }
+    if (flagService.internalUser) {
       final cache = await _cacheService.getCache();
       final List<Future<void>> tasks = [];
       for (final timeline in cache.allTimelines) {
--- a/mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
+++ b/mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
@@ -82,6 +82,9 @@ class MemoryLaneService {
       Bus.instance.on<MLConsentChangedEvent>().listen(_handleMlConsentChange);
       _scheduleStartupBackfill();
       _initialized = true;
+      if (flagService.internalUser) {
+        await _scheduleTimelinesForMemoriesStrip();
+      }
       await _queueFullRecompute();
     } catch (e, s) {
       _logger.severe("Initialization failed", e, s);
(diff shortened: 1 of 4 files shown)


mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -82,6 +82,9 @@ class MemoryLaneService {
       Bus.instance.on<MLConsentChangedEvent>().listen(_handleMlConsentChange);
       _scheduleStartupBackfill();
       _initialized = true;
+      if (flagService.internalUser) {
+        await _scheduleTimelinesForMemoriesStrip();
```
laurens-pilot: A cache-write failure here now aborts init and skips `_queueFullRecompute`: `_writeCacheUnsafe` rethrows write errors, `updateMemoriesStripSchedule` propagates them, and this `await` sits inside `init`'s try block before `await _queueFullRecompute()`. Previously the startup path rarely wrote to disk (first run or a log-version bump only); now every startup that mutates the schedule writes, so a failed write (disk full, etc.) would kill the whole recompute pipeline for the session. Consider wrapping this call in its own try/catch, and/or moving it after `await _queueFullRecompute()`.

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -898,6 +901,69 @@ class MemoryLaneService {
       files.map((file) => MapEntry(localIdToId[file.localID]!, file)),
     );
   }
+
+  Future<void> _scheduleTimelinesForMemoriesStrip() async {
```
laurens-pilot: This only checks `flagService.internalUser`, not `isFeatureEnabled`. If ML consent is revoked or the `facesTimeline` flag is off, cached timelines persist (nothing ever calls `MemoryLaneCacheService.clear()`), so the scheduler keeps pruning/appending schedule entries for a feature that can't display, and the trailing `ensureTimelineReachability` silently no-ops on its own `isFeatureEnabled` check. An early `if (!isFeatureEnabled) return;` would match the other entry points in this service.

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -898,6 +901,69 @@ class MemoryLaneService {
...
+    if (timeline == null) {
+      await _cacheService.updateMemoriesStripSchedule(invalid, null);
+      return;
+    }
+
+    final newBeginShowingAt = valid
+        .map((entry) => entry.value.beginShowingAt + scheduleWindowMs)
+        .followedBy([nowMicros])
+        .max;
+    await _cacheService.updateMemoriesStripSchedule(
```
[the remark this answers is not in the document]
aspizu: no

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -898,6 +901,69 @@ class MemoryLaneService {
...
+        invalid.add(entry.key);
+      }
+    }
+
+    final valid = cache.memoriesStripSchedule.entries
+        .where((entry) => !invalid.contains(entry.key))
+        .toList();
+    final alreadyScheduledAhead = valid.any(
+      (s) => s.value.beginShowingAt >= nowMicros + scheduleWindowMs,
+    );
```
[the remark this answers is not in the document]
aspizu: dont care

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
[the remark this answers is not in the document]
aspizu: dont care

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -108,6 +108,28 @@ class MemoryLaneService {
...
+      for (final person in persons) {
+        for (final cluster in person.data.assigned) {
+          assigned.add(cluster.id);
+        }
+      }
+      try {
+        _topNClusters = await _mlDataDB.getClustersForMemoryLane(assigned);
+      } catch (e, s) {
+        _logger.severe("getClustersForMemoryLane failed:", e, s);
+      }
```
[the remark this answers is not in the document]
aspizu: too bad

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -108,6 +108,28 @@ class MemoryLaneService {
...
+      final assigned = <String>{};
+      for (final person in persons) {
+        for (final cluster in person.data.assigned) {
+          assigned.add(cluster.id);
+        }
+      }
+      try {
+        _topNClusters = await _mlDataDB.getClustersForMemoryLane(assigned);
+      } catch (e, s) {
+        _logger.severe("getClustersForMemoryLane failed:", e, s);
```
laurens-pilot, line 122: Catching and continuing here means a transient DB failure runs everything downstream with `_topNClusters` left at its startup default of `{}`: the scheduler then prunes every cluster schedule entry (`entry.value.isCluster && !_topNClusters.contains(entry.key)`), and the cleanup block below removes every cluster timeline and compute-log entry from the cache, forcing full recomputation and crop regeneration. Before the catch was added, a throw aborted init without touching the cache. Consider skipping the cluster-specific logic (scheduler cluster pruning + the cluster cleanup/recompute block) when this fetch failed.

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -898,6 +949,72 @@ class MemoryLaneService {
...
+    final alreadyScheduledAhead = valid.any(
+      (s) => s.value.beginShowingAt >= nowMicros + scheduleWindowMicros,
+    );
+    if (alreadyScheduledAhead) {
+      await _cacheService.updateMemoriesStripSchedule(invalid, null);
+      return;
+    }
+
+    final timelines = cache.allTimelines.toList()..shuffle();
+    final timeline = timelines.firstWhereOrNull(
```
laurens-pilot, line 1000: This pick filter only looks at timeline state and never consults the person entity, but `getScheduledMemoriesStripTimeline` refuses to show `isIgnored` / `hideFromMemories` persons. So a hidden-from-memories person can be scheduled here, and for their entire 3-day window the consumer returns null with nothing backfilling the slot (scheduling is append-only and runs once per process) - a guaranteed blank strip window. Consider excluding `isIgnored`/`hideFromMemories` persons at pick time via a `PersonService` lookup for non-cluster candidates.

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -219,29 +234,65 @@ class MemoryLaneService {
...
+        .toSet();
+    if (timeline.entries.any((entry) => hiddenFileIds.contains(entry.fileId))) {
+      return null;
     }
 
-    _logger.info("Removing timeline with hidden files for $personId");
-    await _invalidateTimeline(personId);
-    schedulePersonRecompute(personId, isCluster: isCluster, force: true);
-    return null;
+    if (!await areFullFaceCropsCached({
```
laurens-pilot, line 294: When this check fails, the method returns null without queuing repair - unlike `getTimeline`, which calls `_queueTimelineCropReadiness` in the same situation. An evicted crop therefore means a blank strip that only heals via unrelated paths (the recompute sweep after its 2h cooldown, or the user opening the person page). Adding `_queueTimelineCropReadiness(schedule.personID, isCluster: schedule.isCluster)` before the null return would make it self-heal.

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
aspizu: ~~this is invalid because the scheduler runs at app startup, there is no chance of the user ignoring or hiding a person before this runs~~

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
aspizu: queue full recompute already handles this scenario, so there is no need to do this in two places

mobile/apps/photos/lib/services/memory_lane/memory_lane_service.dart
```diff
@@ -219,29 +243,65 @@ class MemoryLaneService {
...
+    }
+    final schedule = await _cacheService.getCurrentMemoriesStripSchedule();
+    if (schedule == null) {
+      return null;
+    }
+    final timeline = await _cacheService.getTimeline(schedule.personID);
+    if (timeline == null || timeline.entries.isEmpty) {
+      return null;
+    }
+    if (!schedule.isCluster) {
```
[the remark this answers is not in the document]
aspizu: this scenario is handled by the UI logic, but that would mean no memory lane is displayed until next 3 days.

laurenspriem (approved):

PR #12906, opened 2026-09-15
[server][space] Identify whose friend limit was reached

anandbaburajan:
Return a separate error when the other person has filled all 9 friend spots. Add tests for both users’ limits.


server/space/repo/friends.go
```diff
@@ -15,6 +15,7 @@ var (
 	ErrAlreadyFriends                 = errors.New("space users are already friends")
 	ErrSelfFriendship                 = errors.New("space users cannot friend themselves")
 	ErrSpaceFriendLimitReached        = errors.New("space friend limit reached")
+	ErrSpaceOtherFriendLimitReached   = errors.New("other space friend limit reached")
```
neeraj-pilot, line 18: Could we simplify this by returning the appropriate `ApiError` directly from the repo and dropping both sentinel errors? The space ID needn’t be exposed in `Message`.

PR #12911, opened 2026-09-15
vecdb: keep or renumber the graph through compaction instead of rebuilding it

laurens-pilot:
Compaction rebuilt the HNSW graph from scratch inside the caller's add or remove, and it fired even when nothing could be reclaimed: re-adding a key in place appends a record that counts as dead without freeing a slot, so a full re-index by re-adding every key ran eight compactions that each rebuilt the graph for nothing (5 s each at 50k i8 on an M4, 46 s at 100k on a Pixel 8, with the writer lock held and every search falling back to exact meanwhile).

Two commits:

1. A compaction with no dead arena slot rewrites the log and keeps the arena and graph as they are. The live slots are exactly 0..n-1, they are staged in ascending order, and replay numbers them identically, so the retained graph is what the compacted log would rebuild. A compaction with dead slots compacts the arena and rebuilds the graph exactly as before.
2. Re-adding a key in place detaches its old node and, when the new vector lands elsewhere, nothing refills the lists that pointed at it. Main repaired that only by accident through the rebuilds above, so the fast path also rebuilds once four times the number of nodes detached since the last rebuild reaches the live count. That count is the insert ordinal minus the live nodes, which the snapshot already carries: no new state, no format change.

An earlier revision of this PR renumbered the graph in place instead of rebuilding it when slots were dead. Review found that concentrated deletions (removing most of one similarity cluster) left the survivors unreachable, so that part is gone; both review scenarios are now regression tests, compared against a fresh rebuild of the same log.

Measured on an M4, i8, 512 dims, min-max over 3 interleaved runs under load (an unrelated benchmark was running):

| | main | this PR |
|---|---|---|
| compaction at 50k with 11 % of keys re-added in place | 3.7-4.1 s | about 50 ms |
| compaction at 100k, same | 10.5-10.6 s | about 70 ms |
| full re-index of 50k by re-add, batches of 200 | 8 rebuilds, 49-60 s | 2 rebuilds, 16-23 s |
| compaction after deleting 10 % at 50k | 4.4-5.9 s | 4.1-4.4 s, still a rebuild |

Independently verified: fast-path soundness by reopening from the log alone across seeds, the reviewer's two cluster reproductions passing, mutation checks on the gate and the fast path, no unsafe, no dependency, no on-disk change. 318 crate tests, the FRB crate tests and lint are green.


laurens-pilot:
Closing in favour of #12949. The two-batch finding was right, and the root cause is older than this PR: an in-place re-add detaches the old node and leaves holes in its neighbours' lists, on main as well (one batch of 270 leaves 25 of 30 survivors unfindable, and it survives flush and reopen). Main only repaired it by accident, through the full rebuilds it ran every 11 % of re-adds. Rather than gate around it, the replacement PR stops cutting nodes out of the graph: a re-add retires the old slot like a remove and appends a new node, so the graph only ever gains nodes between rebuilds and this class of damage cannot occur. It removes code instead of adding it.

PR #12938, opened 2026-09-16
[mobile][photos] improve referral and cast load/network error ui handling

r4khul:
## Description
- Show error bottom sheet when Referrals or Cast Sessions fail to load, instead of opening a broken screen. Initial data is fetched before navigation so the destination screen opens with valid data. 
- Change log entry points (like from home widget) show a loading dialog while fetching.

**Visuals:**

image

## Tests
  - Audited with physical device.


r4khul:
@codex review

r4khul:
PTAL @ashilkn

ashilkn (approved):

PR #12939, opened 2026-09-16
[mobile] Fix progress dialog lifecycle race

neeraj-pilot:
Progress dialogs kept their presentation state and dismissal context in library-level globals. When dismissal raced the dialog builder, or multiple dialogs overlapped, `hide()` could target the wrong route or hit a null context; Auth could then remain blocked behind “Please wait…” after successful authentication.

Make each dialog own its state and exact `DialogRoute`, wait until that route is removed before returning from `hide()`, and remove the fixed timing delay. Add focused regression coverage for immediate dismissal, overlapping dialogs, barrier dismissal, and navigation above an active dialog.


ua741 (approved):

neeraj-pilot:
@codex review

PR #12944, opened 2026-09-16
[web][photos][auth][locker] Consolidate authentication pages and remove unused UI fallbacks

AswinAsok:
## Description

Consolidate repeated auth-page shell wiring across Photos, Auth, and Locker. And removing the unwanted presentation switching fallback which was added to switch to the legacy UI


AswinAsok:
@codex review

PR #12946, opened 2026-09-16
[mobile][photos] show create album when an album search via copy/move has no results

r4khul:
## Description
Show Create album when an album search has no results. The searched name is prefilled in the create-album dialog.

**Visuals**
image

## Tests
  - Audited with physical device.


r4khul:
PTAL @ashilkn

ashilkn (approved):

PR #13047, opened 2026-09-20
[auth] Pin flutter_secure_storage_linux to 1.2.1

### LINKED ISSUE #5589: [Auth] Malformed JSON in keyring
### Description

After running ente-auth and signing in, the token is added to the login keyring file (~/.local/share/keyrings/login.keyring), but the keyring item fails to load after a reboot (doesn't appear in seahorse until I manually remove the secret). The cause is the JSON secret not being minified (or at least containing newlines I'm pretty sure):
non working secret:
```ini
[5]
item-type=0
display-name=io.ente.auth/FlutterSecureStorage
secret={
"auth_secret_key": "REDACTED",
"key": "REDACTED",
"secret_key": "REDACTED"
}
mtime=1744323457
ctime=1744323457
```
working secret:
```ini
[4]
item-type=0
display-name=io.ente.auth/FlutterSecureStorage
secret={"auth_secret_key":"REDACTED","key":"REDACTED","secret_key":"REDACTED"}
mtime=1744323118
ctime=1744323118

```

Steps to Reproduce:
1. Run ente-auth and complete the sign-in process.
2. Verify that the token is written to the keyring file with newlines inside the JSON content.

This issue doesn't happen in version 4.2.8 and prior.

### Version

4.3.2+432

Ente Auth

Desktop - Linux

Arzaroth:
## Description

Fixes #5589

`flutter_secure_storage_linux` 1.2.2 changed how the keyring payload is serialised:

```diff
-    const std::string output = value.dump();
+    const std::string output = value.dump(0);
```

In nlohmann/json, `dump(indent)` pretty-prints for any `indent >= 0`. Only the default `-1` is compact. So `dump(0)` asks for pretty-printing with zero-width indentation, which still emits newlines:

```
dump(0)  => [{
"auth_secret_key": "A",
"key": "B",
"secret_key": "C"
}]
dump()   => [{"auth_secret_key":"A","key":"B","secret_key":"C"}]
```

gnome-keyring's plaintext keyring file stores a secret on a single `secret=<value>` line, so those newlines corrupt the file. The item is dropped on the next unlock and Auth comes back logged out. An offline-only user loses their codes with it.

The JSON is well-formed, since newlines are legal JSON whitespace. The damage is to the keyring file format, so nothing in Dart can detect it or repair it.

Upstream restored `value.dump()` in `flutter_secure_storage_linux` 3.x ("Reverted json.dump with indentations due to problems", in the 10.0.0 notes). That line needs `flutter_secure_storage` 10+, but the repo is held at 9.0.0 (cf. #10104).

1.2.1 is the last 1.x release without the regression.

## Tests

- [x] Reproduced the corruption on 1.2.3 in an isolated keyring, matching the report byte for byte
- [x] Confirmed the item is gone on 1.2.3 after a keyring daemon restart
- [x] Confirmed 1.2.1 writes the secret on a single line
- [x] Confirmed the secret survives a daemon restart on 1.2.1
- [x] `flutter build linux --release` succeeds for `mobile/apps/auth` with the pin, plugin built from 1.2.1

**1. Root cause, isolated.** I compiled a small throwaway exec from each version's `Secret.hpp` linked against libsecret and stored the same three keys Auth stores. It ran against a private gnome-keyring with its own D-Bus session, its own `XDG_DATA_HOME`, and a `login` keyring with an empty password, so the file lands in the plaintext format from the issue report.

On 1.2.3 `login.keyring` reproduces the report exactly:

```
[3]
item-type=0
display-name=io.ente.authTEST/FlutterSecureStorage
secret={
"auth_secret_key": "AAAA",
"key": "BBBB",
"secret_key": "CCCC"
}
```

Restart the keyring daemon against that file, which is what a reboot does, and `secret-tool lookup` returns nothing. The item is gone.

On 1.2.1:

```
[3]
item-type=0
display-name=io.ente.authTEST/FlutterSecureStorage
secret={"auth_secret_key":"AAAA","key":"BBBB","secret_key":"CCCC"}
```

After the same restart, `secret-tool lookup` returns the payload intact.

**2. Build.** `flutter build linux --release` for `mobile/apps/auth` on Flutter 3.47.4 succeeds with the pin, and `libflutter_secure_storage_linux_plugin.so` is built from `flutter_secure_storage_linux-1.2.1`. `APPLICATION_ID` resolves, as the CMake ordering above predicts.

## AI disclosure

This PR (well, the analysis and code, not the PR body) was done with AI (Opus 5 if that's relevant). The changes are minimal and I can stand behind them.

This last bit is claude analysis:

**The build failure on new clang.** With clang 22, `flutter build linux` dies before the link step, on `-Werror` against the bundled `json.hpp`:

```
.../flutter_secure_storage_linux/linux/include/json.hpp:24392:35: error: identifier '_json' preceded by
whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator]
```

This is pre-existing, since the header is identical in 1.2.1 and 1.2.3, and it does not affect CI's toolchain. I left it out to keep the diff on the bug. It is the same breakage #10104 was chasing, and the compile flag suggested there clears it: `-Wno-error=deprecated-literal-operator`, next to the existing `-Wno-error=deprecated-declarations` line in `mobile/apps/auth/linux/CMakeLists.txt`.

**The reason given for the 9.0.0 pin.** Four pubspecs (`apps/auth`, `apps/photos`, `packages/configuration`, `packages/lock_screen`) carry this comment above `flutter_secure_storage: 9.0.0`:

```yaml
  # Version 9.2.4 leaves lock-screen keys behind after an app reinstall.
  # Upgrade only after this is fixed or with a migration.
  # https://github.com/juliansteenbakker/flutter_secure_storage/issues/870#issuecomment-2777447937
```

That issue closed as completed on 2025-12-11, and 11.2.0 shipped a further fix in the same area, `fix(darwin): find keychain items across accessibility levels` (#1269). The condition the comment names has been met upstream, even if the broader reasons from #10104 still hold.


ua741 (approved):

neeraj-pilot:
Could you move this override below `flutter_cache_manager` to fix `sort_pub_dependencies`, and shorten its comment to:

```yaml
  # 1.2.2 and 1.2.3 write secrets that GNOME Keyring cannot reload (#5589).
  flutter_secure_storage_linux: 1.2.1
```

